Issue #14, opened 2026-04-20
[feat] Add language skills: Java, Kotlin, Swift


lugassawan:
## Goal
Add three more language skills — Java, Kotlin, and Swift — cookie-cutter off the existing language-skill template.

## Motivation
Completes JVM + Apple platform coverage alongside the existing `language-go`, `language-rust`, and `language-typescript` skills. The language-skill template is already proven and well-documented in `docs/extending.md`.

## Sequencing
Sequenced after #8 (Python) by convenience, not blocked. The template is already validated by `language-go`, `language-rust`, and `language-typescript`. These three skills can ship independently of #8 if needed.

## Deliverables
- `skills/language-java/SKILL.md`
- `skills/language-kotlin/SKILL.md`
- `skills/language-swift/SKILL.md`
- All under 150 lines each
- Update the Languages table in `docs/catalog.md`

## Template / references
- Follow `docs/extending.md` § "Adding a language skill"
- Mirror `skills/language-go/SKILL.md`
- Trigger keywords per language:
  - Java: `.java`, `pom.xml`, `build.gradle`, JVM, Spring, records, sealed classes, virtual threads
  - Kotlin: `.kt`, `build.gradle.kts`, coroutines, `suspend`, `Result`, sealed interfaces
  - Swift: `.swift`, `Package.swift`, actors, `async`/`await`, Result builders, optionals, protocols

## Acceptance
- [ ] Three files, each at `skills/language-<name>/SKILL.md` with `name:` frontmatter matching the directory name exactly
- [ ] Each under 150 lines
- [ ] Each focuses on language-specific idioms, not general SWE advice (that lives in the `principle-*` skills)
- [ ] Java and Kotlin skills don't overlap — each explains its own idioms

---

## Updated 2026-04-29
Realigned with current codebase: corrected all skill paths from nested `skills/languages/<name>/` to flat `skills/language-<name>/` per the layout enforced by `CONTRIBUTING.md`; softened the blocker framing (not blocked — template already proven by `language-go`, `language-rust`, `language-typescript`); updated template reference to `skills/language-go/SKILL.md`; added catalog update to deliverables. Acceptance criteria are unchanged.

Issue #8, opened 2026-04-20
[feat] Add Python language skill


lugassawan:
## Goal
Add a Python language skill that auto-loads when Claude is editing `.py` files or the user mentions Python idioms.

## Motivation
Python is the largest missing language skill. `docs/extending.md` already uses it as the canonical "how to add a language" worked example and targets `skills/language-python/` as the exact destination — no new decisions needed, just follow the recipe.

## Deliverables
- `skills/language-python/SKILL.md` (under 150 lines)
- Update the Languages table in `docs/catalog.md`

## Template / references
- Mirror shape of `skills/language-go/SKILL.md`
- Follow the recipe in `docs/extending.md` § "Adding a language skill"
- Frontmatter trigger keywords should cover: `.py` files, `pyproject.toml`, `requirements.txt`, `setup.py`, and terms like Python, pytest, mypy, asyncio, dataclass, type hints.

## Suggested body coverage
- Error handling (exceptions, `raise ... from`, `contextlib.suppress`)
- Typing (PEP 484/604, `TypedDict`, `Protocol`, `Final`, `cast`)
- Dataclasses vs Pydantic vs `TypedDict` — when to reach for each
- `async`/`await` pitfalls and structured concurrency (`asyncio.TaskGroup`)
- Packaging (pyproject.toml, src/ layout, editable installs)
- Testing idioms (pytest fixtures, parametrize, `monkeypatch`)

## Acceptance
- [ ] `skills/language-python/SKILL.md` exists with valid frontmatter; `name:` field matches directory name exactly
- [ ] Under 150 lines
- [ ] Trigger keywords cause auto-load on `.py` files
- [ ] No overlap with what `superpowers:*` already covers

---

## Updated 2026-04-29
Realigned with current codebase: corrected skill path (`skills/languages/python/` → `skills/language-python/`) and template reference (`skills/languages/go/` → `skills/language-go/`) per the flat layout enforced by `CONTRIBUTING.md`; replaced "validates template before Java/Kotlin/Swift" motivation with the fact that `docs/extending.md` already targets this exact path; added catalog update to deliverables. Acceptance criteria are unchanged.

Ansi-A:
I can work on this. Please assign it to me.

PR #22, opened 2026-04-28
[feat] Add JAVA 

7vikfox:
## Summary

-

## Test Plan

- [ ]

Closes #


lugassawan:
## Code Review

`★ Insight ─────────────────────────────────────`
Skill files in this repo are parsed by a custom `validate.py` that expects exact structural conventions: `---`-delimited YAML frontmatter, `name` matching the parent directory, and triggers embedded in prose `description` rather than a YAML list. Deviating from any of these causes silent drops or hard CI failures — not syntax errors that editors catch.
`─────────────────────────────────────────────────`

---

### Critical

**`skills/SKILL.md` | Python skill at wrong directory depth — never loaded**

The validator globs `skills/*/SKILL.md` (depth 2). `skills/SKILL.md` is at depth 1 and is silently skipped — the Python skill is dead on arrival. Even if the loader did pick it up, the parent dir would be `skills`, which will never match a `name` field.

> Fix: move to `skills/language-python/SKILL.md`; set `name: language-python`.

---

**`skills/SKILL.md` | Missing closing `---` frontmatter delimiter**

The frontmatter opens with `---` but never closes. `parse_frontmatter()` returns `None`, emitting "missing or malformed frontmatter" and failing CI.

> Fix: add `---` after the `triggers` block, before the first `##` heading.

---

**`skills/languages_swift/SKILL.md` | `---` appears mid-frontmatter, terminating parsing early**

The closing `---` lands immediately after the last trigger entry, before the body. YAML parsers treat the first `---` they find as the end of frontmatter, so `name` and `description` are sliced off and the parsed dict is missing required fields.

> Fix: ensure the closing `---` is the first bare `---` line after all scalar frontmatter fields.

---

### High

**`skills/languages_java/SKILL.md`, `skills/languages_kotlin/SKILL.md` | No opening `---` delimiter**

`parse_frontmatter()` checks `if not text.startswith("---"): return None`. Both files begin with `name: ...` — they fail the check unconditionally.

> Fix: wrap frontmatter in `---` / `---` delimiters.

---

**All four files | `triggers:` YAML block is not supported by the infrastructure**

Existing skills (`language-go`, `language-rust`, `language-typescript`) have no `triggers:` key. The loader matches skills via the prose `description:` string. The indented `- value` lines in `triggers:` are not parsed by `parse_frontmatter()`'s line-by-line scalar parser — they are silently dropped, leaving the skill with zero triggers and no auto-activation.

> Fix: remove `triggers:` entirely from all four files. Embed trigger conditions in `description:` as a natural-language sentence, e.g.:
> `"Java idioms — JVM, Spring, concurrency. Auto-load when working with .java, pom.xml, build.gradle, or when the user mentions Java, records, virtual threads, or Spring."`

---

**`skills/languages_java/`, `skills/languages_kotlin/` | Directory naming breaks `name == dir` invariant**

Every existing skill uses `language-{lang}` (singular, kebab-case). These dirs use `languages_` (plural, underscore). `validate.py` enforces `frontmatter.name == skill_dir_name` — both skills will fail this check. A loader filtering by the `language-` prefix will also miss them entirely.

> Fix: rename directories to `language-java` and `language-kotlin`; update `name:` fields to match.

---

### Medium

**All four files | `name` field values do not match expected directory names**

Even after fixing delimiters, the current `name` values (`languages-java`, `languages-kotlin`, `languages-swift`, `SKILL.md`) will fail the name-matches-directory check. The Python skill's `name: SKILL.md` matches its filename, not a language identifier.

> Fix: after renaming directories, set `name:` to `language-java`, `language-kotlin`, `language-swift`, `language-python`.

---

**All four files | No code examples — thinner than existing skills**

Go, Rust, and TypeScript skills include concrete annotated snippets that make guidance copy-pasteable. These files are bullet lists only. The content is topically correct but will be less actionable in practice.

> Fix: add at least one code example per major section (error handling, concurrency, typing), following the pattern in `skills/language-go/SKILL.md`.

---

### Low

**PR title | Doesn't match required format**

`"Added JAVA "` fails `^\[(feat|fix|refactor|test|ci|docs|perf|chore|polish|breaking)\] .+` — this is the direct cause of the CI failure.

> Fix: rename to `[feat] Add Java, Kotlin, Swift, and Python language skills`

**PR body | `Closes #` left empty**

Per the template instructions, an empty `Closes #` will fail the issue-reference CI check. Either link an issue number or replace with a standalone `N/A — <reason>` line.

---

## Summary

| Category | Finding |
|---|---|
| Correctness | None of the 4 files survive `validate.py` as written — wrong depth, missing delimiters, or bad name fields |
| Design | Invented `triggers:` YAML block has no infrastructure support; underscore+plural naming breaks every loader convention |
| Test gaps | `bash scripts/validate.sh` locally would have caught all of these before CI |

**What's good:** the content topics are right — null safety, virtual threads, structured concurrency, coroutines. With the structural fixes above and a few code examples, these would be solid contributions.

7vikfox (commented):
Fixed review feedback:
- Corrected directory structure to match validator requirements
- Added proper frontmatter delimiters
- Renamed skill to match folder name
- Removed unsupported triggers block and embedded logic in description
- Added basic code example

lugassawan (changes requested):
### Critical — skills won't load

**1. Python: broken directory path**
The directory was committed as `skills /language_python/` (there is a literal space between `skills` and `/language_python/`). The skill loader walks `skills/` and will never find this file. Rename to `skills/language-python/`.

**2. All four files: missing `---` frontmatter fences**
Every existing skill wraps its metadata in `---` delimiters:
\`\`\`markdown
---
name: language-go
description: ...
---
\`\`\`
Without them, `name:` and `description:` are treated as raw Markdown body text — the skill has no machine-readable identity and won't load. All four files need both opening and closing `---`.

The Swift file has a closing `---` but no opening one — same result.

**3. Python: `name:` field is wrong**
`name:language_python` has two problems: missing space after the colon (malformed YAML), and snake_case instead of kebab-case. The issues (#8) specify the directory as `skills/language-python/` and the `name:` must match exactly:
\`\`\`yaml
name: language-python
\`\`\`

**4. `triggers:` is a non-standard key (Java, Kotlin, Swift)**
No existing skill uses a `triggers:` block — it's silently ignored. The canonical pattern embeds trigger hints directly in `description`:
\`\`\`yaml
description: Java best practices ... Auto-load when working with .java, pom.xml, build.gradle files, or when the user mentions JVM, Spring, records, sealed classes, or virtual threads.
\`\`\`
Move the trigger lists from all three files into their `description` fields.

---

### Medium — formatting

**5. Missing `##` section headers (Java, Kotlin, Swift)**
Sections like `Modern Java`, `Testing`, `Build Tools`, `Spring Ecosystem`, `Null Safety`, `Coroutines`, `Swift Fundamentals`, etc. are plain text lines, not Markdown headers. Every existing skill uses `##` for all sections. Add `## ` prefix to each.

**6. Python example code contradicts the file's content**
The `## Example` block shows:
\`\`\`python
def add(a, b):
    return a + b
\`\`\`
This has no type annotations and no docstring — which is exactly what the `## Typing` section below it warns against. Either remove the example block or replace it with something that demonstrates the file's own advice (e.g. a typed function using `TypedDict` or `Protocol`).

---

### Missing deliverables

**7. `docs/catalog.md` not updated**
Both #8 and #14 list updating the Languages table in `docs/catalog.md` as a required deliverable. This file is not in the diff.

---

### Low / housekeeping

- **PR title** — `[feat] Add JAVA` doesn't reflect that Python, Kotlin, and Swift are also added. Suggest: `[feat] Add Python, Java, Kotlin, and Swift language skills`
- **PR description** — Summary section is a bare `-`, `Closes #` is incomplete, test plan boxes are all unchecked. Please fill these in and manually verify each skill loads before marking ready.
- **Commit history** — 11 commits for 4 new files, including several "Update and rename SKILL.md to SKILL.md" and two "Merge branch 'main' into main". Please squash before merge.

---

The content of each skill is a reasonable starting point — this is fixable. Once the frontmatter is correct and the catalog is updated, the files should be loadable and we can do a content pass.

7vikfox (commented):
Addressed all review feedback:

- Fixed directory structure and naming (kebab-case, correct depth)
- Added proper frontmatter (--- delimiters) to all skill files
- Updated name fields to match folder names
- Removed unsupported `triggers:` blocks and moved activation hints into descriptions
- Converted section titles to proper Markdown headers (##)
- Improved Python example to align with typing guidelines


Please let me know if anything else needs adjustment.

PR #132, opened 2026-05-08
[fix] cleanup-merged: ExitWorktree before rimba removal; rimba-first in Phase 1 template

lugassawan:
## Summary
- **Step 3 `ExitWorktree action=keep` guard**: the rimba post-merge hook fires during `git pull` in Step 3 — with the session still inside the worktree being deleted, rimba's subprocess inherits a dead cwd, leaving the branch alive and stranding the session at `$HOME`. `ExitWorktree action=keep` is now the first action of Step 3, before `MAIN_REPO=` derivation and before `git pull`.
- **rimba partial-success recovery**: when rimba removes the worktree directory but fails at `git branch -D` (dead cwd from the hook fast path), the skill now treats this as partial success and falls through to Step 5 rather than aborting.
- **Phase 1 template rimba-first**: `plan-workflow-section.md` Phase 1 previously listed `superpowers:using-git-worktrees` as a peer-primary to `rimba add`. Agents then invoked it when rimba was available, triggering `EnterWorktree name=…` which mangles slash-containing branch names (e.g. `feature/101-foo` → `worktree-feature+101-foo`). Phase 1 now marks rimba as Primary and `using-git-worktrees` as Fallback-only with an explicit DO NOT warning.
- **Bonus fixes from review**: Step 4 `WORKTREE_FOUND` now uses `-C "$MAIN_REPO"` anchor; Step 4 `WORKTREE_GONE=1` wording clarified; rimba partial-success path clarifies `WORKTREE_GONE` semantics.

## Test Plan
- [x] `python3 scripts/validate.py` — clean
- [x] `pytest tests/ -q` — 223/223 pass (6 new in `tests/test_cleanup_merged_skill.py`)
- [x] `ExitWorktree` appears in Step 3 before `MAIN_REPO=` (L52) and before `git pull` (L70) — verified by test ordering assertions
- [x] rimba block has both "partial" and "fall through to Step 5" — verified by tightened AND assertion
- [x] Phase 1 template: rimba before `using-git-worktrees` as explicit fallback — verified by test

Issue: N/A — rimba/worktree workflow hardening surfaced during PR #131 cycle

### DIFF
--- a/skills/workflow-cleanup-merged/SKILL.md
+++ b/skills/workflow-cleanup-merged/SKILL.md
@@ -84,7 +84,11 @@ After Step 3 sync, run the verification gate to check whether the rimba post-mer
 ```bash
 WORKTREE_FOUND=$(git worktree list --porcelain \
   | awk '/^branch refs\/heads\/<headRefName>$/{print 1; exit}')
-BRANCH_FOUND=$(git -C "$MAIN_REPO" rev-parse --verify "refs/heads/<headRefName>" 2>/dev/null && echo 1 || true)
+if git -C "$MAIN_REPO" rev-parse --verify "refs/heads/<headRefName>" 2>/dev/null; then
+  BRANCH_FOUND=1
+else
+  BRANCH_FOUND=
+fi
 WORKTREE_GONE=0
 [ -z "$WORKTREE_FOUND" ] && [ -z "$BRANCH_FOUND" ] && WORKTREE_GONE=1
 ```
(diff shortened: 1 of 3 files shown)


skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -161,7 +170,8 @@ The hook silently swallows errors (`|| true`). If the verification gate yields `
 3. (Once per repo) recommend the user run `rimba hook install` to automate future post-merge cleanups via a git hook — this removes the need for manual `/swe-workbench:cleanup-merged` invocations.
 
 **Failure handling:**
-- On failure, report the rimba error verbatim and abort. Do not proceed to branch deletion.
+- If rimba reports that the worktree was removed but branch deletion failed (e.g. "worktree removed but failed to delete branch", or rimba exits non-zero after the worktree directory is confirmed gone) — treat as **partial success**. The worktree is already gone; `WORKTREE_GONE` remains `0` (Step 4 ran before rimba), so Step 5 executes normally. Fall through to Step 5 (`git branch -D`) from `$MAIN_REPO`. Do NOT abort.
+- On any other rimba failure (e.g. dirty worktree refused, cannot locate worktree), report the rimba error verbatim and abort. Do not proceed to branch deletion.
```
lugassawan, line 184: **Medium** — Partial-success detection relies on matching rimba stdout text.

The failure-handling bullet uses "e.g. \"worktree removed but failed to delete branch\"" as the signal, but rimba's exact message format is unspecified. If rimba emits a different message or only a non-zero exit with a stack trace, an agent cannot reliably distinguish partial success from a full failure.

**Suggested fix:** After `$RIMBA remove` exits non-zero, run a filesystem check as the canonical signal:
```bash
[ -d "<worktree-path>" ] || # worktree directory is gone → partial success → fall through to Step 5
```
This decouples the decision from rimba's output format.

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -221,7 +231,8 @@ git worktree remove "$WORKTREE"
 | Step 3 (sync main) fails | Non-zero exit from `git checkout` or `git pull` | Warn in report. Do not abort — sync is best-effort; cleanup proceeds. |
 | PR number not derivable from current branch | `gh pr view` fails | Ask the user for the PR number explicitly. |
 | Hook ran but did not clean | `WORKTREE_GONE=0` after sync despite hook active | Fall through to rimba-binary or shell strategy. No abort. |
-| cwd deleted mid-flow by hook | `fatal: not a git repository` on next command | Step 3 cwd-anchor (`cd "$MAIN_REPO"` before pull) prevents this. If observed, re-run from the main repo root. |
+| cwd deleted mid-flow by hook | `fatal: not a git repository` on next command | Re-run from the main repo root. (Step 3a `ExitWorktree action=keep` prevents this when followed.) |
```
lugassawan: **Low** — Action column leads with recovery before prevention.

Previously the row led with the preventive mechanism. The current text leads with recovery ("Re-run from the main repo root"), making prevention feel secondary. For an instructional table agents scan during live failures, order matters.

**Suggested swap:**
```
| Step 3a `ExitWorktree action=keep` prevents this when followed. If observed, re-run from the main repo root. |
```

lugassawan (commented):
Reviewed by `swe-workbench:reviewer`. Posted 2 inline comments, deduped 0.

No critical issues. The three-part Step 3 sub-structure (3a/3b/3c) is the right granularity. Phase 1 template change is well-scoped with the correct DO NOT warning. The `WORKTREE_FOUND` revert (after user feedback on `-C "$MAIN_REPO"` false confidence) is correct — the same-bash-invocation note now makes the scoping contract explicit.

PR #161, opened 2026-05-10
[fix] Prevent temp-file re-glob in capture / product-manager

### LINKED ISSUE #160: [feat] Reduce swe-workbench plugin's self-usage footprint (audit + cheaper subagent models + tighten ticket-context)
## Product framing

**Problem.** As the plugin author, I just observed that 71% of my own Claude usage in the last 24h came from the `swe-workbench` plugin — with `/swe-workbench:ticket-context` (30%) and subagents under it (19%) as the single largest contributors. Subagent-heavy sessions accounted for 97% of usage overall. Every routine workflow run is silently spending more tokens than it needs to because subagents inherit the default model and skill prompts haven't been tuned for cost.

**Value.** Every plugin user (starting with me) gets measurably lower token spend per session — especially on commands that auto-load `ticket-context` — without losing the structured context that makes those commands useful. Cheaper subagent models on simpler roles free budget for the agents that genuinely need top-tier reasoning.

**Acceptance criteria.**
- An audit document exists listing every agent + skill the plugin contributes, each annotated with: current model (or "inherits"), whether it spawns subagents, and a recommended cost tier (S/M/L).
- `ticket-context` skill and its subagents are reviewed: prompt is tightened (char/token delta recorded) and/or a cheaper model is set in frontmatter where the role doesn't need top-tier reasoning.
- Agent frontmatter for clearly-simple subagents (fetchers, formatters, single-purpose extractors) declares an explicit cheaper model instead of inheriting the default.
- A short note in the plugin README or `docs/` documents the cost-tier convention so future agents/skills follow it by default.

**Impact / Effort.**
- **Impact: M** — touches every session that uses the plugin; meaningful token reduction, no new user-facing capability.
- **Effort: M** — audit across `agents/` + `skills/` is mechanical; tightening `ticket-context` and validating cheaper models still produce good output requires before/after testing.

## Goal

Cut the `swe-workbench` plugin's own contribution to Claude usage by tightening prompts and configuring cheaper models on subagents that don't need top-tier reasoning, without regressing capability on the commands users rely on.

## Motivation

Last-24h usage breakdown surfaced by Claude Code:

> - 97% of your usage came from subagent-heavy sessions
> - 19% of your usage came from subagents under "swe-workbench:ticket-context"
> - 30% of your usage came from /swe-workbench:ticket-context
> - 71% of your usage came from plugin "swe-workbench"

`ticket-context` + its subagents = **49% of total Claude usage** in that window. It is the single dominant cost driver. The Claude Code UX explicitly suggests three levers — cheaper models for simple subagents, tighter prompts for heavy skills, and review of what the plugin contributes — and all three apply here.

## Surface

- [x] Skill — `workflow-*` (specifically `ticket-context` and any other heavy skills surfaced by the audit)
- [x] Subagent (frontmatter changes across simple subagents)
- [x] Documentation (cost-tier convention note)
- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Slash command
- [ ] Hook
- [ ] Build / CI / release

## Sketch

Three workstreams, sequenced:

1. **Audit pass.** Walk `agents/` and `skills/` and produce a table:
   `surface | path | current model | spawns subagents? | recommended tier (S/M/L) | notes`.
   Land as `docs/cost-audit.md` (or similar).

2. **`ticket-context` tightening.** Read the skill + any subagents it spawns. Cut redundant instructions, collapse examples, and (where the role permits) set a cheaper model in frontmatter. Capture before/after char count and a brief sanity-check transcript. Priority: this is the single biggest lever (49% of usage).

3. **Frontmatter sweep on simple subagents.** For agents identified as tier S in the audit (single-purpose fetchers/formatters/extractors), add an explicit cheaper model line to frontmatter. Leave reasoning-heavy agents (reviewer, debugger, designer, etc.) on the default.

Then add a short README/`docs/` section: "Cost tiers — when to set `model:` in agent/skill frontmatter."

## Acceptance

- [ ] `docs/cost-audit.md` (or equivalent) lists every plugin agent + skill with current model, subagent fan-out, and recommended tier.
- [ ] `ticket-context` prompt char/token count is reduced and/or its subagents declare a cheaper model; before/after numbers recorded in the PR description.
- [ ] At least the tier-S subagents identified in the audit have explicit `model:` frontmatter set to a cheaper option.
- [ ] Plugin docs explain the cost-tier convention so it's followed by default in future contributions.
- [ ] A follow-up usage check (after a representative day of use) shows the plugin's share trending down vs. the 71% baseline. Recorded as a comment on this issue or in a follow-up.

lugassawan:
## Summary

- `product-manager.md` step 7: obtain timestamp once via `date +%s`, reuse for both the body file and `.cmd` sidecar — never re-derive or re-glob.
- Both files: enforce `Write` tool over Bash heredoc so the file path is always deterministic.

## Root cause

A prior `/capture` session wrote the body to `/tmp/capture-swe-workbench-<timestamp>.md` via a Bash heredoc, then re-found it with `ls /tmp/capture-swe-workbench-*.md | tail -1`. A stale file from a different session (`capture-swe-workbench-rimba.md`) sorted last alphabetically, so the wrong body was filed as issue #160. The `.cmd` sidecar pattern (step 9 reads the pre-written command verbatim) was already the right defense; this closes the one remaining gap where two independently derived timestamps could produce mismatched filenames.

Closes #160

## Test plan

- [ ] Run `/capture` in a repo that has a stale `/tmp/capture-*` file from a prior session — confirm the preview and filed issue use the correct body.
- [ ] Confirm `product-manager` step 9 reads the `.cmd` sidecar and runs it verbatim without regenerating the path.

### DIFF
--- a/commands/capture.md
+++ b/commands/capture.md
@@ -23,7 +23,7 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
 
 6. **Draft.** With a template: fill its sections, prepend `## Product framing`. Without a template: use `## Problem` / `## Value` / `## Acceptance criteria` / `## Impact / Effort` / `## Additional context`.
 
-7. **Preview gate.** Obtain a Unix timestamp once (`date +%s`) and reuse it — never re-derive or re-glob. Write the body to `/tmp/capture-<repo-slug>-<timestamp>.md` using the `Write` tool (not a Bash heredoc), then print:
+7. **Preview gate.** Obtain a Unix timestamp once (`date +%s`) and reuse it — never re-derive or re-glob. Write the body to `/tmp/capture-<repo-slug>-<unix-timestamp>.md` using the `Write` tool (not a Bash heredoc). Also write the exact `gh issue create --title "..." --body-file <path>` command to `/tmp/capture-<repo-slug>-<unix-timestamp>.cmd` using the `Write` tool. Then print:
    ```
    Filing into: <owner>/<repo>
    Template: <chosen template> | none — default body
@@ -39,4 +39,4 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
    ```
    **Wait for the user to reply `confirm`. Do NOT run `gh issue create` on this turn.**
 
-8. **File on confirm.** Only when the user replies `confirm`, run the exact printed command and return the issue URL.
+8. **File on confirm.** Only when the user replies `confirm`, read the command from the `.cmd` sidecar written in step 7 and run it exactly as written — do not regenerate the title or path. Return the issue URL.
--- a/commands/capture.md
+++ b/commands/capture.md
@@ -23,7 +23,7 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
 
 6. **Draft.** With a template: fill its sections, prepend `## Product framing`. Without a template: use `## Problem` / `## Value` / `## Acceptance criteria` / `## Impact / Effort` / `## Additional context`.
 
-7. **Preview gate.** Write the body to `/tmp/capture-<repo-slug>-<timestamp>.md`, then print:
+7. **Preview gate.** Obtain a Unix timestamp once (`date +%s`) and reuse it — never re-derive or re-glob. Write the body to `/tmp/capture-<repo-slug>-<timestamp>.md` using the `Write` tool (not a Bash heredoc), then print:
    ```
    Filing into: <owner>/<repo>
    Template: <chosen template> | none — default body
--- a/agents/product-manager.md
+++ b/agents/product-manager.md
@@ -51,7 +51,7 @@ You apply lightweight PM lenses, not a heavy framework. No RICE math beyond Impa
      ## Additional context
      ```
 
-7. **Write temp file.** Derive `<repo-slug>` from the `nameWithOwner` value, replacing `/` with `-` and stripping any character outside `[a-zA-Z0-9_-]`. Write the drafted body to `/tmp/capture-<repo-slug>-<unix-timestamp>.md` using the `Write` tool. Also write a one-line command file to `/tmp/capture-<repo-slug>-<unix-timestamp>.cmd` containing the exact `gh issue create --title "..." --body-file <path>` command (title double-quoted, path absolute). Do NOT run `gh issue create` yet.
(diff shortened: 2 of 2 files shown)


commands/capture.md
```diff
@@ -23,7 +23,7 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
 
 6. **Draft.** With a template: fill its sections, prepend `## Product framing`. Without a template: use `## Problem` / `## Value` / `## Acceptance criteria` / `## Impact / Effort` / `## Additional context`.
 
-7. **Preview gate.** Write the body to `/tmp/capture-<repo-slug>-<timestamp>.md`, then print:
+7. **Preview gate.** Obtain a Unix timestamp once (`date +%s`) and reuse it — never re-derive or re-glob. Write the body to `/tmp/capture-<repo-slug>-<timestamp>.md` using the `Write` tool (not a Bash heredoc), then print:
    ```
```
lugassawan, line 27: **Placeholder inconsistency: `<timestamp>` vs `<unix-timestamp>`**

`agents/product-manager.md` step 7 uses `<unix-timestamp>` (explicit: epoch integer from `date +%s`). This file uses `<timestamp>` (ambiguous). An agent parsing this could plausibly use a human-readable form (e.g. `20260510-143022`), producing filenames that don't match when the subagent is involved.

**Suggested fix:** Change `<timestamp>` → `<unix-timestamp>` to match the agent file and the `date +%s` instruction in the same sentence.

lugassawan:
**Missing `.cmd` sidecar write — step 7/8 inconsistency**

`commands/capture.md` step 7 now correctly writes the body file with the `Write` tool, but doesn't write the `.cmd` sidecar that `product-manager.md` step 7 writes. Step 8 currently says "run the exact printed command," meaning the agent reconstructs the path from the preview string — weaker than reading the stored sidecar value. This partially re-opens the same class of gap the fix intended to close.

**Suggested fix:** Add to step 7: "Also write the exact `gh issue create` command to `/tmp/capture-<repo-slug>-<unix-timestamp>.cmd` using the `Write` tool." Update step 8 to say "read and run the command from the `.cmd` sidecar" rather than "the exact printed command."

> Note:  line 42 (step 8) is outside this diff — posting here rather than inline.

lugassawan:
**Missing `.cmd` sidecar write — step 7/8 inconsistency**

`commands/capture.md` step 7 now correctly writes the body file with the `Write` tool, but doesn't write the `.cmd` sidecar that `product-manager.md` step 7 writes. Step 8 currently says "run the exact printed command," meaning the agent reconstructs the path from the preview string — weaker than reading the stored sidecar value. This partially re-opens the same class of gap the fix intended to close.

**Suggested fix:** Add to step 7: "Also write the exact `gh issue create` command to `/tmp/capture-<repo-slug>-<unix-timestamp>.cmd` using the `Write` tool." Update step 8 to say "read and run the command from the `.cmd` sidecar" rather than "the exact printed command."


> Note: commands/capture.md line 42 (step 8) is outside this diff — posting here rather than inline.

lugassawan (commented):
Reviewed by `reviewer` (swe-workbench). Posted 2 comments (1 inline, 1 PR-level — step 8 line out of diff range), deduped 0.

Issue #181, opened 2026-05-10
[feat] PreToolUse hook: secret-detection on Write/Edit operations (AWS keys, GitHub tokens, .env content)


lugassawan:
## Goal

Add a PreToolUse hook that intercepts Write and Edit operations and rejects them when the content contains common secret patterns (AWS keys, GitHub tokens, `.env`-style assignments with sensitive names).

## Motivation

`hooks/hooks.json` currently has one PreToolUse matcher: destructive shell commands (`rm -rf`, force-push, hard-reset on protected branches). Secrets in source code are caught only post-hoc by `.githooks/pre-commit` — if that hook is installed. A regex-based PreToolUse on Write/Edit would catch secret leakage at *authoring time*, before the file is written, making it impossible to accidentally commit a secret even if the git hook is absent.

## Surface

- [x] Hook
- [x] Documentation

## Sketch

**Trigger:** `PreToolUse` on `Write` and `Edit` tool calls.

**Patterns to detect (initial set):**

| Pattern | Example |
|---|---|
| AWS Access Key ID | `AKIA[0-9A-Z]{16}` |
| AWS Secret Access Key | `[0-9a-zA-Z/+]{40}` adjacent to `aws_secret` |
| GitHub personal access token | `ghp_[a-zA-Z0-9]{36}` |
| GitHub fine-grained token | `github_pat_[a-zA-Z0-9_]{82}` |
| Generic API key assignment | `API_KEY\s*=\s*["\'][^"\']{16,}["\']` |
| Generic secret/password assignment | `(SECRET|PASSWORD|PASSWD|TOKEN)\s*=\s*["\'][^"\']{8,}["\']` |
| `.env` lines with sensitive names | Lines matching `^(SECRET|API_KEY|TOKEN|PASSWORD)=` in any file |

**On match:** block the Write/Edit, print which pattern matched and which line, and suggest either using an environment variable reference or adding the file to `.gitignore`.

**False-positive escape hatch:** allow an inline `# nosecret` comment (or equivalent) to suppress the check for a specific line (for test fixtures, documentation examples, etc.).

## Acceptance

- [ ] `hooks/hooks.json` includes a PreToolUse entry matching `Write` and `Edit`.
- [ ] Hook script rejects writes containing any of the listed patterns.
- [ ] `# nosecret` (or equivalent) suppresses the check for that line.
[...]

PR #216, opened 2026-05-13
[feat] workflow-commit-and-pr: scan staged files for likely secrets

### LINKED ISSUE #203: [feat] workflow-commit-and-pr: scan staged files for likely secrets before commit
## Goal

Before running `git commit`, scan the staged file set for filenames that commonly
contain secrets (`.env`, `*.pem`, `*.key`, `credentials.json`, `secrets.*`) and
warn the user before the commit proceeds.

## Motivation

This is a different threat model from existing issue #181 (PreToolUse hook on
Write/Edit). #181 catches secrets that Claude itself introduces at write time. The
commit-layer scan catches secrets staged by *anyone* — a human running `git add`,
an IDE that auto-stages, or another tool — before Claude is asked to commit.

The two checks are complementary and ship independently:

| Threat model                              | Layer      | Caught by  |
| ----------------------------------------- | ---------- | ---------- |
| Claude writes a `.env` via Write/Edit     | Write/Edit | #181       |
| User runs `git add .env` then says commit | Commit     | This issue |

Without the commit-layer scan, the `workflow-commit-and-pr` skill happily commits a
staged `.env` or `*.pem` that slipped past the hook layer.

## Surface

- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [x] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch

Before the commit preview block:

```bash
SUSPICIOUS=$(git diff --staged --name-only | grep -E \
  '(^|/)(\.env(\.|$)|.+\.pem$|.+\.key$|credentials\.json$|secrets?\.[a-z]+$)' \
  || true)
if [ -n "$SUSPICIOUS" ]; then
  echo "WARNING: staged files look like they may contain secrets:"
  echo "$SUSPICIOUS"
  echo "Reply 'proceed' to commit anyway, or 'cancel' to abort."
fi
```

Matches should be conservative — false-positives on `*.example` / `*.sample` /
`*.template` / `*.dist` must be excluded.

## Acceptance

- [ ] Skill scans `git diff --staged --name-only` against a curated pattern set
      before the commit preview.
- [ ] When matches found: prints the file list, asks `proceed` vs `cancel`,
      aborts on `cancel` (no commit, no unstage — leave staging untouched).
- [ ] Patterns exclude obvious non-secret variants (`*.example`, `*.sample`,
      `*.template`, `*.dist`).
- [ ] Failure modes table updated with the suspicious-staged-files row.
- [ ] PR body cross-references #181 to clarify the layering.

lugassawan:
## Summary
- Add `## Pre-commit gate: suspicious staged files` section to `workflow-commit-and-pr` skill, inserted between `## Branch-naming check` and `## Doc-only [no ci] rule`
- Gate scans staged filenames with a `grep -iE` heuristic before the commit preview runs; matches `.env*`, `*.pem`, `*.key`, `credentials.json`, `secrets?.*`, and variants
- On a match, presents an `AskUserQuestion` prompt (`Commit anyway` / `Cancel`); `Cancel` aborts with no `git restore --staged` — staging is left untouched
- Exclusion pass uses terminal-suffix matching (`\.(example|sample|template|dist)$`) so `.env.example` / `secrets.sample.yaml` pass silently; `my.sample.pem` is correctly flagged
- Append rows to the Failure-modes and Common-mistakes tables documenting the new behaviour

## Layering note
Complements #181 (PreToolUse Write/Edit hook). #181 catches secrets Claude introduces at authoring time; this PR catches secrets staged by humans, IDEs, or other tools before commit. Ships independently. Related: #181.

## Test Plan
- [x] `pytest tests/test_workflow_commit_and_pr_secret_scan.py -v` — 2 new tests green (prose-presence + regex matrix with 16 positive / 11 negative filenames)
- [x] `pytest tests/test_skill_repo_ambiguity.py -v` — no regression on "this repo" constraint
- [x] `pytest tests/test_skill_triggers.py -v` — no BM25 drift (no new trigger line added)
- [x] Full `pytest tests/` — 364 passed, 0 failed

Closes #203

### DIFF
--- a/skills/workflow-commit-and-pr/SKILL.md
+++ b/skills/workflow-commit-and-pr/SKILL.md
@@ -86,9 +86,10 @@ Behaviour:
 
 Before the commit preview (or before running `git commit`), scan the staged
 file set for filenames that commonly hold secrets. This is a commit-layer
-twin of #181's write-time hook: #181 catches secrets the agent introduces
-via Write/Edit; this gate catches secrets staged by anyone (a human running
-`git add`, an IDE that auto-stages, or any other tool) before commit.
+twin of the PreToolUse Write/Edit hook: that hook catches secrets the agent
+introduces at authoring time; this gate catches secrets staged by anyone (a
+human running `git add`, an IDE that auto-stages, or any other tool) before
+commit.
 
 The scan is a filename heuristic, not a content scan — it cannot catch a
 secret pasted inside an otherwise innocuous `config.yaml`. False negatives
--- a/skills/workflow-commit-and-pr/SKILL.md
+++ b/skills/workflow-commit-and-pr/SKILL.md
@@ -106,8 +106,10 @@ SUSPICIOUS=$(git diff --staged --name-only \
 ```
 
 The exclusion pass (`*.example`, `*.sample`, `*.template`, `*.dist`)
-prevents false-positives on sanitised template files that are conventional
-to commit (e.g. `.env.example`, `secrets.sample.yaml`).
+prevents false-positives on `.env` template variants (e.g. `.env.example`,
+`.env.sample`). Files like `secrets.example.yaml` are already excluded by
+the positive pattern (which requires a single trailing extension), so the
+exclusion pass is not their gate.
 
 **On no matches** → silent; continue to `## Doc-only [no ci] rule`.
 
--- a/skills/workflow-commit-and-pr/SKILL.md
+++ b/skills/workflow-commit-and-pr/SKILL.md
@@ -82,6 +82,58 @@ Behaviour:
 - If branch doesn't match `<type>/<kebab>` → **warn-only** (don't auto-rename): "Branch `<name>` doesn't match the `<type>/<kebab>` convention. Continue anyway? Reply `yes` or rename first."
 - If branch matches → silent.
 
+## Pre-commit gate: suspicious staged files
+
+Before the commit preview (or before running `git commit`), scan the staged
+file set for filenames that commonly hold secrets. This is a commit-layer
+twin of #181's write-time hook: #181 catches secrets the agent introduces
+via Write/Edit; this gate catches secrets staged by anyone (a human running
+`git add`, an IDE that auto-stages, or any other tool) before commit.
+
+The scan is a filename heuristic, not a content scan — it cannot catch a
+secret pasted inside an otherwise innocuous `config.yaml`. False negatives
+are expected; treat a clean scan as "no obvious filename red flags", not
+"safe to commit".
+
+**Scan command** (run before `[no ci]` is computed):
+
+```bash
+SUSPICIOUS=$(git diff --staged --name-only \
+  | grep -iE '(^|/)([^/]*\.env(\.|$)|.+\.pem$|.+\.key$|credentials\.json$|secrets?\.[a-z]+$)' \
+  | grep -vE '\.(example|sample|template|dist)$' \
+  || true)
+```
+
+The exclusion pass (`*.example`, `*.sample`, `*.template`, `*.dist`)
+prevents false-positives on sanitised template files that are conventional
+to commit (e.g. `.env.example`, `secrets.sample.yaml`).
+
+**On no matches** → silent; continue to `## Doc-only [no ci] rule`.
+
+**On match**, print the file list verbatim to the user, then call
+`AskUserQuestion`:
+
+```json
+{
+  "questions": [{
+    "question": "Staged files have names that commonly contain secrets. Commit anyway, or cancel?",
+    "header": "Suspicious",
+    "multiSelect": false,
+    "options": [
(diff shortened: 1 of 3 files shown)


skills/workflow-commit-and-pr/SKILL.md
```diff
@@ -82,6 +82,58 @@ Behaviour:
...
+secret pasted inside an otherwise innocuous `config.yaml`. False negatives
+are expected; treat a clean scan as "no obvious filename red flags", not
+"safe to commit".
+
+**Scan command** (run before `[no ci]` is computed):
+
+```bash
+SUSPICIOUS=$(git diff --staged --name-only \
+  | grep -iE '(^|/)([^/]*\.env(\.|$)|.+\.pem$|.+\.key$|credentials\.json$|secrets?\.[a-z]+$)' \
+  | grep -vE '\.(example|sample|template|dist)$' \
```
lugassawan: **Medium — case-sensitivity mismatch on exclusion grep**

The positive pass uses `-iE` (case-insensitive) but the exclusion pass uses `-vE` (case-sensitive). A file named `.env.EXAMPLE` or `.env.Sample` will pass the positive filter and then not be excluded, resulting in a false-positive alert.

```suggestion
  | grep -ivE '\.(example|sample|template|dist)$' \\
```

tests/test_workflow_commit_and_pr_secret_scan.py
```diff
@@ -0,0 +1,160 @@
...
+    pos_pattern = pos_match.group(1)
+
+    excl_match = re.search(r"grep\s+-vE\s+'([^']+)'", bash_block)
+    assert excl_match, "Could not extract grep -vE pattern from bash block"
+    excl_pattern = excl_match.group(1)
+
+    def should_flag(filename: str) -> bool:
+        if not re.search(pos_pattern, filename, re.IGNORECASE):
+            return False
+        if re.search(excl_pattern, filename, re.IGNORECASE):
```
lugassawan, line 147: **Medium — test simulation diverges from bash on exclusion case-sensitivity**

`should_flag()` applies `re.IGNORECASE` to the exclusion pattern, but the bash `grep -vE` is case-sensitive. This masks the divergence: `.env.EXAMPLE` would be flagged by real bash but not by this test.

Fix: remove `re.IGNORECASE` from the exclusion check to mirror bash faithfully, and add uppercase variants to the negative matrix:

```suggestion
        if re.search(excl_pattern, filename):
```

Also add to `NEGATIVE_FILENAMES`:
```python
".env.EXAMPLE",   # case-insensitive exclusion must fire
".env.Sample",    # mixed-case variant
```

(Pair with adding `-i` to the bash exclusion grep.)

skills/workflow-commit-and-pr/SKILL.md
```diff
@@ -82,6 +82,58 @@ Behaviour:
...
+  || true)
+```
+
+The exclusion pass (`*.example`, `*.sample`, `*.template`, `*.dist`)
```
lugassawan, line 108: **Low — prose overstates what the exclusion pass does**

The comment says the exclusion pass prevents false-positives on `secrets.sample.yaml` style files, but the positive regex `secrets?\.[a-z]+$` doesn't match `secrets.sample.yaml` in the first place (single-segment extension only). The exclusion never fires for those. The prose is correct about `.env.example` but misleading about the compound-extension case.

Suggested tightening:
```suggestion
The exclusion pass (`*.example`, `*.sample`, `*.template`, `*.dist`)
prevents false-positives on .env template variants (e.g. `.env.example`,
`.env.sample`). Files like `secrets.example.yaml` are already excluded by
the positive pattern (which requires a single trailing extension).
```

lugassawan (commented):
Reviewed by `reviewer` (swe-workbench). Posted 3 inline comments, deduped 0.

**Findings summary:**
- **Medium (×2):** Case-sensitivity mismatch — positive grep uses `-iE` but exclusion uses `-vE`, so `.env.EXAMPLE` slips through; Python test mirrors this bug via `re.IGNORECASE` on the exclusion. Fix: add `-i` to the bash exclusion grep and drop `re.IGNORECASE` from the test's exclusion check.
- **Low (×1):** Prose overstates what the exclusion pass catches — `secrets.example.yaml` is already excluded by the positive regex, not the exclusion grep.
- **Notable gaps (non-blocking):** `id_rsa`, `*.p12`, `*service-account*.json` not in the pattern set; the section documents false negatives as expected, so these belong in a follow-up issue.

The no-auto-unstage invariant is well-encoded in three places (prose + AskUserQuestion description + Common mistakes table). The `my.sample.pem` positive case is a sharp edge correctly handled. The `|| true` pipeline tail is correct.

PR #242, opened 2026-05-16
[fix] Suppress posted Review Summary on author self-review

lugassawan:
## Summary

- Extends the existing `CURRENT_USER`/`AUTHOR_LOGIN` self-review gate (already used for address-feedback CTA suppression) to also suppress `## Review Summary` from the GitHub-posted review body when the reviewer is the PR author
- Adds `HAS_NARRATIVE` and `IS_SELF_REVIEW` named predicates to Step 7's `$SUMMARY` builder in both `workflow-pr-review` and `workflow-pr-review-followup` (Extract Variable + Decompose Conditional per Fowler); `IS_SELF_REVIEW` guards against empty-variable false-positives
- Updates Step 4 narrative instruction wording to document the two-path behaviour (posted vs in-session only); updates both Common mistakes tables with a new self-review row cross-referencing the address-feedback CTA gate

## Test Plan

- [x] Changed skills/commands/agents load without errors (`bash scripts/validate.sh` — exit 0)
- [x] Examples in the diff were exercised manually (fixture script: cross-author `$SUMMARY` still contains `## Review Summary`; self-review `$SUMMARY` = BYLINE-only)
- [x] `python -m pytest tests/` — 428/428 pass (no regressions in routing + followup dispatch tests)
- [ ] Manual: `/swe-workbench:review <N>` on a cross-author PR — posted body still contains `## Review Summary`
- [ ] Manual: `/swe-workbench:review <N>` on own PR — posted body is BYLINE-only; narrative still shown in Claude session

- [ ] Reproduce: run self-review before patch → confirm `## Review Summary` appears in posted body
- [ ] Verify: run self-review after patch → confirm posted body is BYLINE-only; narrative still surfaced in session
- [ ] Regression: run cross-author review → confirm no change in posted body

Issue: N/A — UX refinement reported in-session by maintainer

### DIFF
--- a/skills/workflow-pr-review-followup/SKILL.md
+++ b/skills/workflow-pr-review-followup/SKILL.md
@@ -189,15 +189,20 @@ BYLINE="_Re-reviewed by \`reviewer\` ([swe-workbench](https://github.com/lugassa
 if [ "$HAS_NARRATIVE" = true ] && [ "$IS_SELF_REVIEW" = false ]; then
   SUMMARY=$(printf '## Review Summary\n\n%s\n\nDetailed feedback in inline comments.\n\n**Review Decision: %s**\n\n---\n%s\n' \
     "$NARRATIVE" "$DECISION" "$BYLINE")
-else
-  # Reused for empty-narrative AND self-review.
+elif [ "$IS_SELF_REVIEW" = false ]; then
+  # No narrative but cross-author: post just the byline.
   SUMMARY="$BYLINE"
+else
+  # Self-review: nothing posted to GitHub; inline comments speak for themselves.
+  SUMMARY=""
 fi
 ```
 
-Submit per the parsed decision:
+Submit only when `IS_SELF_REVIEW = false` — GitHub blocks self-approval, and for self-review the Step 6 inline comments are sufficient without a review-event body:
 - `APPROVE` → `gh pr review "$PR" --approve --body "$SUMMARY"`
 - `COMMENT` → `gh pr review "$PR" --comment --body "$SUMMARY"`
+
+When `IS_SELF_REVIEW = true`, skip the review-event submission entirely.
 
 **Never** use `--request-changes`.
 
--- a/skills/workflow-pr-review-followup/SKILL.md
+++ b/skills/workflow-pr-review-followup/SKILL.md
@@ -85,7 +85,7 @@ Pass the agent:
 - Footer instruction (load-bearing — opt-in per the agent's `## Decision footer (when instructed)` block):
   > "End the review with EXACTLY ONE of `**Review Decision: APPROVE**` or `**Review Decision: COMMENT**` on its own line, no prefix or trailing text. Never `REQUEST_CHANGES`."
 - Narrative instruction (load-bearing — opt-in per the agent's `## Review Summary (when instructed)` block):
-  > "Begin the review with a `## Review Summary` section: 2–4 sentences capturing overall posture, the strongest positives, and the most important concerns. The orchestrator uses these paragraphs as the top-level PR review body. Do not repeat per-finding detail there — that goes in the severity-grouped findings below."
+  > "Begin the review with a `## Review Summary` section: 2–4 sentences capturing overall posture, the strongest positives, and the most important concerns. When the reviewer is not the PR author, the orchestrator uses these paragraphs as the top-level PR review body; otherwise the narrative is shown in the Claude session only and not posted to GitHub. Do not repeat per-finding detail there — that goes in the severity-grouped findings below."
 - Ticket-context prelude (if Step 3 produced one).
 
 Store the agent's complete text response as `REVIEWER_OUTPUT`.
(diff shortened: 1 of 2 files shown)


skills/workflow-pr-review/SKILL.md
```diff
@@ -172,15 +172,17 @@ NARRATIVE=$(awk '
   /^\*\*Review Decision:/ { exit }
   { print }
 ' <<< "$REVIEWER_OUTPUT" | sed -e '/^[[:space:]]*$/d' -e '/^## Review Summary[[:space:]]*$/d')
+HAS_NARRATIVE="$([ -n "$(echo "$NARRATIVE" | tr -d '[:space:]')" ] && echo true || echo false)"
+IS_SELF_REVIEW="$([ -n "$CURRENT_USER" ] && [ -n "$AUTHOR_LOGIN" ] && [ "$CURRENT_USER" = "$AUTHOR_LOGIN" ] && echo true || echo false)"
```
lugassawan: **Medium** — `IS_SELF_REVIEW` uses a compound `&&/||` chain inside `$()` which is harder to read and has a subtle pitfall: if `echo true` fails in a restricted environment, `|| echo false` fires for the wrong reason. A plain `if/then/else` block is unambiguous:

```bash
if [ -n "$CURRENT_USER" ] && [ -n "$AUTHOR_LOGIN" ] && [ "$CURRENT_USER" = "$AUTHOR_LOGIN" ]; then
  IS_SELF_REVIEW=true
else
  IS_SELF_REVIEW=false
fi
```

The fix also applies to the same line in `workflow-pr-review-followup/SKILL.md`.

lugassawan: **Medium** — When `CURRENT_USER` or `AUTHOR_LOGIN` is empty (e.g. `gh api /user` returns unexpected shape, or a bot PR has no `author.login`), the emptiness guard short-circuits to `false`, which silently treats the scenario as "cross-author" and posts the `## Review Summary` to GitHub. This is arguably the safer default, but it is undocumented and could surprise maintainers debugging a bot-authored PR.

Consider adding a `stderr` warning when either variable is empty:

```bash
if [ -z "$CURRENT_USER" ] || [ -z "$AUTHOR_LOGIN" ]; then
  echo "[warn] IS_SELF_REVIEW: identity unknown; treating as cross-author." >&2
  IS_SELF_REVIEW=false
elif [ "$CURRENT_USER" = "$AUTHOR_LOGIN" ]; then
  IS_SELF_REVIEW=true
else
  IS_SELF_REVIEW=false
fi
```

Mirror this in `workflow-pr-review-followup/SKILL.md:177`.

skills/workflow-pr-review/SKILL.md
```diff
@@ -251,5 +253,6 @@ Match against ANY author (User Decision 2). On match, skip posting AND add 👍
 | Parse threads from REST `pulls/{N}/comments` | REST returns review-comment-by-comment; threading is reconstructed by the GraphQL `reviewThreads` shape. Use GraphQL to fetch, REST to post. |
 | Force-add 👍 to your own existing comment | Check `reactions.nodes[].user.login` first; skip if you've already reacted. |
 | Block on cleanup | Cleanup runs in background `(... ) &`. Don't `wait` for it. |
-| Skip the narrative instruction in Step 4 | Without it, the reviewer does NOT emit `## Review Summary` (per its `## Review Summary (when instructed)` block). Step 7 falls back to the legacy one-liner silently — body is not wrong but loses the prose narrative. |
+| Skip the narrative instruction in Step 4 | Without it, the reviewer does NOT emit `## Review Summary` (per its `## Review Summary (when instructed)` block). Step 7 falls back to the BYLINE-only branch silently — body is not wrong but loses the prose narrative for cross-author reviews (self-review intentionally produces BYLINE-only; see row below). |
```
lugassawan: **Low** — "see row below" is a fragile forward-reference: if a future editor inserts a row between these two, the cross-reference silently breaks. Reference the target row by its title instead:

> `(self-review intentionally produces BYLINE-only; see "Post ``## Review Summary`` on self-review" row)`

Same fix needed at `workflow-pr-review-followup/SKILL.md:252`.

skills/workflow-pr-review-followup/SKILL.md
```diff
@@ -246,4 +249,6 @@ Match against ANY author. On match, skip posting AND add 👍 to the thread head
 | Use `--task "pr-review-$PR"` for the worktree | Use `--task "pr-followup-$PR"` to avoid colliding with a still-active primary-review worktree. |
 | Omit the footer instruction in Step 4 | Without it, the agent does NOT emit the footer. Step 5 will then abort. |
 | Forget repo-relative-path instruction | GitHub comment positioning requires repo-relative paths. The agent will emit `$WT/...` paths otherwise. |
+| Skip the narrative instruction in Step 4 | Without it, the reviewer does NOT emit `## Review Summary` (per its `## Review Summary (when instructed)` block). Step 7 falls back to the BYLINE-only branch silently — body is not wrong but loses the prose narrative for cross-author reviews (self-review intentionally produces BYLINE-only; see row below). |
```
lugassawan: **Low** — Same fragile "see row below" forward-reference as in `workflow-pr-review/SKILL.md:256`. Reference the target row by name to survive future table edits:

> `(self-review intentionally produces BYLINE-only; see "Post ``## Review Summary`` on self-review" row)`

lugassawan (commented):
## Review Summary

This is a tight, well-scoped change that correctly suppresses the `## Review Summary` narrative from GitHub's PR review body on self-reviews, symmetrically across both orchestration skills. The logic is readable and the documentation is kept in lockstep. One correctness concern: the `IS_SELF_REVIEW` computation uses a compound `&&/||` chain inside `$()` that is more complex than necessary and has a subtle failure mode in restricted environments. A second Medium concern covers the silent cross-author fallback when either identity variable is empty — safe default but undocumented. Overall the PR is lean and the policy intent is sound.

**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 4 inline comments, deduped 0._

skills/workflow-pr-review/SKILL.md
lugassawan: Addressed in f7290ec: refactored `IS_SELF_REVIEW` to a plain `if/then/else` block — no more `&&/||` chain inside `$()`.

lugassawan: Addressed in f7290ec: the `if/then/else` refactor includes a dedicated empty-identity branch that emits a `stderr` warning and defaults to `IS_SELF_REVIEW=false` (cross-author), making the fallback explicit and observable.

skills/workflow-pr-review/SKILL.md
lugassawan: Addressed in f7290ec: replaced "see row below" with explicit title `see "Post ## Review Summary on self-review" row` in both files.

skills/workflow-pr-review-followup/SKILL.md
lugassawan: Addressed in f7290ec: same fix applied to workflow-pr-review-followup/SKILL.md.

lugassawan (commented):
## Review Summary

All four prior findings addressed cleanly. `IS_SELF_REVIEW` is now a plain `if/then/else` with an explicit empty-identity warn branch. Forward-references updated to stable row titles. Both orchestrators remain in lock-step. No new findings — **LGTM**.

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comments, deduped 0._

PR #251, opened 2026-05-16
[fix] Preserve pre-existing hooks and core.hooksPath in setup.sh

### LINKED ISSUE #231: [bug] setup.sh silently overwrites existing git hooks and clears core.hooksPath
## Surface affected

- [x] Build / CI / release

**Specific surface:** `scripts/setup.sh:7-14`

## Claude Code version

N/A — defect is in repo automation, not Claude Code itself.

## Plugin version

0.1.11

## Reproduction steps

1. Have a working `lefthook` (or similar) install in some other repo — it sets `core.hooksPath` per-repo and installs hooks at a custom path.
2. Clone `swe-workbench`. In a fresh clone, set up a personal `pre-commit` hook in `.git/hooks/pre-commit` (e.g., a linter you run locally).
3. Run `bash scripts/setup.sh` per CONTRIBUTING.md.
4. Observe:
   - Line 8: `git -C "$ROOT" config --unset core.hooksPath 2>/dev/null || true` — wipes the repo-local `core.hooksPath` you may have set deliberately.
   - Line 11-14: `ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"` — silently overwrites your existing `.git/hooks/pre-commit` symlink/file with no backup, no warning, no opt-in.

## Expected behavior

`setup.sh` should:
1. Detect pre-existing hooks in `.git/hooks/` that point somewhere other than `.githooks/`. Warn and require `--force` to overwrite.
2. Detect a pre-existing repo-local `core.hooksPath` set to something other than `.githooks`. Warn similarly.
3. Document scope intent: `git -C "$ROOT" config --local --unset core.hooksPath` to make the local-vs-global scope self-documenting (even though `-C` without `--global` is already repo-local).

## Actual behavior

Bulk-replace strategy with no detection and no opt-in. Developer's pre-existing hook setup is silently destroyed on a fresh clone.

## Environment

Any developer who runs `bash scripts/setup.sh` on a clone where they already have hooks configured.

---

## Audit reasoning (full 11-field schema)

| Field | Value |
|-------|-------|
| **title** | `setup.sh` removes user's `core.hooksPath` and overwrites existing hook symlinks |
| **severity** | Medium |
| **domain** | reliability |
| **file_line** | `scripts/setup.sh:7-14` |
| **symptom** | Unconditional `git config --unset core.hooksPath`; `ln -sfn` overwrites every hook in `.git/hooks/` with no backup, warning, or opt-in. |
| **root_cause** | Bulk-replace strategy with no detection of pre-existing hooks and no idempotency check that existing target equals desired target. |
| **reasoning_chain** | 1. Line 8 unsets `core.hooksPath` regardless of user intent. 2. Loop at 11 iterates `.githooks/*`; `ln -sfn` overwrites silently. 3. No check like `readlink "$hook" != "$expected" && warn`. 4. Contributors using lefthook elsewhere silently lose that setup in this clone. |
| **counter_evidence_considered** | Is `git config --unset` repo-local by default? Yes — `git -C "$ROOT"` scopes to local config, so global is safe. But repo-local `core.hooksPath` set deliberately is still wiped. Is overwrite caught? No — `ln -sfn` silent on success. |
| **confidence** | 0.70 |
| **effort** | low |
| **suggested_fix** | Detect existing hooks and existing `core.hooksPath`; warn and require `--force` to overwrite. Add `--local` explicitly to the `git config --unset` for self-documenting scope. |

lugassawan:
## Summary
- `scripts/setup.sh`: rewrote with `--force`/`-h` arg parser; conflict preflight detects pre-existing regular-file hooks, foreign symlinks, and non-default repo-local `core.hooksPath`; fail-fast on conflicts without `--force`; idempotency (correct symlinks skipped silently); POSIX-safe `printf '%s'` (no `%b`); empty-glob guard for empty `.githooks/`; trailing-slash normalization for `core.hooksPath`; explicit `git config --local` scope
- `tests/test_setup.py`: adds `TestSetupShPreservation` with 5 regression tests — idempotent rerun, regular-file conflict, foreign-symlink conflict, `core.hooksPath` conflict, `--force` override (all 7 setup tests pass alongside the full 436-test suite)
- `CONTRIBUTING.md`: Setup section updated to document `--force` usage, conflict behavior, and idempotent re-run guarantee

## Test Plan
- [x] `pytest tests/test_setup.py -v` — 7/7 passed (5 new + 2 existing)
- [x] `pytest tests/ -q` — 436/436 passed, no regressions
- [x] `shellcheck --severity=warning scripts/setup.sh` — clean
- [ ] Manual: `./scripts/setup.sh` exits 0, prints "Git hooks linked: ..."
- [ ] Manual: re-running `./scripts/setup.sh` exits 0, no warnings on stderr (idempotent)
- [ ] Manual: `ln -sfn /tmp/foo .git/hooks/pre-push && ./scripts/setup.sh` exits 1, lists conflict, leaves symlink intact
- [ ] Manual: `./scripts/setup.sh --force` exits 0, symlink replaced with correct `.githooks/pre-push` target
- [ ] Manual: `git config --local core.hooksPath /tmp/x && ./scripts/setup.sh` exits 1; `--force` unsets it

- [ ] Reproduce the original bug: confirm bare `./scripts/setup.sh` overwrites a pre-existing hook on `main` before this fix
- [ ] Verify fix: same setup with this branch exits 1 and leaves the hook intact

Closes #231

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,15 +1,69 @@
 #!/bin/sh
 set -e
+
+FORCE=0
+for arg in "$@"; do
+  case "$arg" in
+    --force|-f) FORCE=1 ;;
+    --help|-h)
+      echo "Usage: $0 [--force]"
+      echo
+      echo "Installs symlinks in .git/hooks/ -> .githooks/."
+      echo "Pre-existing hooks and a non-default repo-local core.hooksPath"
+      echo "are preserved unless --force is passed."
+      exit 0 ;;
+    *) printf 'Error: unknown argument: %s\n' "$arg" >&2; exit 2 ;;
+  esac
+done
+
 ROOT="$(git rev-parse --show-toplevel)"
 GIT_DIR="$(git rev-parse --git-dir)"
 case "$GIT_DIR" in /*) ;; *) GIT_DIR="$ROOT/$GIT_DIR" ;; esac
 
-# Remove any stale core.hooksPath — symlinks in the default location handle it.
-git -C "$ROOT" config --unset core.hooksPath 2>/dev/null || true
+# --- Preflight: collect conflicts ------------------------------------------
+conflicts=""
+
+existing_hp="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
+existing_hp="${existing_hp%/}"   # normalize trailing slash
+if [ -n "$existing_hp" ] \
+   && [ "$existing_hp" != ".githooks" ] \
+   && [ "$existing_hp" != "$ROOT/.githooks" ]; then
+  conflicts="${conflicts}  core.hooksPath = $existing_hp (expected unset or .githooks)
+"
+fi
 
 mkdir -p "$GIT_DIR/hooks"
 for src in "$ROOT"/.githooks/*; do
-  name=$(basename "$src")
+  [ -e "$src" ] || [ -L "$src" ] || break   # empty glob — nothing to do
+  name="$(basename "$src")"
+  target="$GIT_DIR/hooks/$name"
+  expected="$ROOT/.githooks/$name"
+  if [ -L "$target" ]; then
+    [ "$(readlink "$target")" = "$expected" ] && continue   # already correct
+    conflicts="${conflicts}  $target -> $(readlink "$target") (expected $expected)
+"
+  elif [ -e "$target" ]; then
+    conflicts="${conflicts}  $target (regular file, not a symlink)
+"
+  fi
+done
+
+if [ -n "$conflicts" ]; then
+  if [ "$FORCE" -eq 0 ]; then
+    printf 'Error: setup.sh would overwrite existing hook state:\n' >&2
+    printf '%s' "$conflicts" >&2
+    printf 'Re-run with --force to overwrite.\n' >&2
+    exit 1
+  fi
+  printf 'warning: --force: overwriting:\n' >&2
+  printf '%s' "$conflicts" >&2
+fi
+
+# --- Apply -----------------------------------------------------------------
+git -C "$ROOT" config --local --unset core.hooksPath 2>/dev/null || true
+for src in "$ROOT"/.githooks/*; do
+  [ -e "$src" ] || [ -L "$src" ] || break   # empty glob — nothing to do
+  name="$(basename "$src")"
   ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
 done
 
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set -eu
 
 FORCE=0
 for arg in "$@"; do
@@ -34,7 +34,7 @@ fi
 
 mkdir -p "$GIT_DIR/hooks"
 for src in "$ROOT"/.githooks/*; do
-  [ -e "$src" ] || [ -L "$src" ] || break   # empty glob — nothing to do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
   name="$(basename "$src")"
   target="$GIT_DIR/hooks/$name"
   expected="$ROOT/.githooks/$name"
@@ -60,9 +60,14 @@ if [ -n "$conflicts" ]; then
 fi
 
 # --- Apply -----------------------------------------------------------------
-git -C "$ROOT" config --local --unset core.hooksPath 2>/dev/null || true
+existing_hp_final="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
+existing_hp_final="${existing_hp_final%/}"
+case "$existing_hp_final" in
+  ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
+  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
+esac
 for src in "$ROOT"/.githooks/*; do
-  [ -e "$src" ] || [ -L "$src" ] || break   # empty glob — nothing to do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
   name="$(basename "$src")"
   ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
 done
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -6,11 +6,11 @@ for arg in "$@"; do
   case "$arg" in
     --force|-f) FORCE=1 ;;
     --help|-h)
-      echo "Usage: $0 [--force]"
-      echo
-      echo "Installs symlinks in .git/hooks/ -> .githooks/."
-      echo "Pre-existing hooks and a non-default repo-local core.hooksPath"
-      echo "are preserved unless --force is passed."
+      printf 'Usage: %s [--force]\n' "$0"
+      printf '\n'
+      printf 'Installs symlinks in .git/hooks/ -> .githooks/.\n'
+      printf 'Pre-existing hooks and a non-default repo-local core.hooksPath\n'
+      printf 'are preserved unless --force is passed.\n'
       exit 0 ;;
     *) printf 'Error: unknown argument: %s\n' "$arg" >&2; exit 2 ;;
   esac
@@ -60,9 +60,7 @@ if [ -n "$conflicts" ]; then
 fi
 
 # --- Apply -----------------------------------------------------------------
-existing_hp_final="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
-existing_hp_final="${existing_hp_final%/}"
-case "$existing_hp_final" in
+case "$existing_hp" in
   ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
   *) git -C "$ROOT" config --local --unset core.hooksPath ;;
 esac
--- a/tests/test_setup.py
+++ b/tests/test_setup.py
@@ -84,7 +84,7 @@ class TestSetupShPreservation:
         assert result.returncode != 0
         assert not hook.is_symlink(), "hook must not have been replaced"
         assert str(hook) in result.stderr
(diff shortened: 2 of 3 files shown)


scripts/setup.sh
```diff
@@ -1,15 +1,69 @@
...
+    printf '%s' "$conflicts" >&2
+    printf 'Re-run with --force to overwrite.\n' >&2
+    exit 1
+  fi
+  printf 'warning: --force: overwriting:\n' >&2
+  printf '%s' "$conflicts" >&2
+fi
+
+# --- Apply -----------------------------------------------------------------
+git -C "$ROOT" config --local --unset core.hooksPath 2>/dev/null || true
```
lugassawan: **High:** `core.hooksPath` is unconditionally unset here even when the preflight whitelisted it as non-conflicting (`.githooks` or `$ROOT/.githooks`).

A developer who explicitly set `core.hooksPath=.githooks` has it silently cleared on every re-run — no stderr, no warning, breaking the stated idempotency guarantee.

Suggested fix — only unset when it was not already in the whitelisted post-state:
```sh
existing_hp_final="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
existing_hp_final="${existing_hp_final%/}"
case "$existing_hp_final" in
  ""|.githooks|"$ROOT/.githooks") ;; # unset or already correct — leave it alone
  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
esac
```

scripts/setup.sh
```diff
@@ -1,15 +1,69 @@
...
+   && [ "$existing_hp" != ".githooks" ] \
+   && [ "$existing_hp" != "$ROOT/.githooks" ]; then
+  conflicts="${conflicts}  core.hooksPath = $existing_hp (expected unset or .githooks)
+"
+fi
 
 mkdir -p "$GIT_DIR/hooks"
 for src in "$ROOT"/.githooks/*; do
-  name=$(basename "$src")
+  [ -e "$src" ] || [ -L "$src" ] || break   # empty glob — nothing to do
```
lugassawan: **Medium:** The comment "empty glob — nothing to do" is misleading — this guard also fires when `.githooks/` does not exist at all. More accurate:
```sh
[ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
```
(Same comment applies to line 65 in the Apply section.)

scripts/setup.sh
```diff
@@ -1,15 +1,69 @@
 #!/bin/sh
 set -e
```
lugassawan: **Low:** `set -e` without `set -u` means a typo in `$FORCE` (or any variable) silently expands to empty string. `set -eu` is cheap insurance and is consistent with the defensive posture of the rest of the script.

tests/test_setup.py
```diff
@@ -47,6 +47,80 @@ def repo_with_worktree(tmp_path):
...
+            cwd=repo, check=False, capture_output=True, text=True, env=_CLEAN_ENV,
+        )
+
+    def test_idempotent_rerun(self, repo_with_worktree):
+        repo, _ = repo_with_worktree
+        r1 = self._run_setup(repo)
+        assert r1.returncode == 0
+        r2 = self._run_setup(repo)
+        assert r2.returncode == 0
+        assert r2.stderr == ""
```
lugassawan, line 63: **High:** No test covers the `core.hooksPath=.githooks` (whitelisted) case. The unconditional unset on `scripts/setup.sh:63` is invisible to CI because the fixture never pre-sets `core.hooksPath`, so the unset is always a no-op in tests.

Suggested test to add:
```python
def test_idempotent_with_whitelisted_hookspath(self, repo_with_worktree):
    repo, _ = repo_with_worktree
    _git("config", "--local", "core.hooksPath", ".githooks", cwd=repo)
    result = self._run_setup(repo)
    assert result.returncode == 0
    assert result.stderr == ""
    val = subprocess.check_output(
        ["git", "config", "--local", "--get", "core.hooksPath"],
        cwd=repo, text=True, env=_CLEAN_ENV,
    ).strip()
    assert val == ".githooks", "whitelisted core.hooksPath must not be silently cleared"
```

lugassawan (commented):
## Review Summary

The PR correctly addresses the core bug: setup.sh now collects conflicts before mutating anything, exits with a clear error when conflicts exist, and requires \`--force\` to proceed. The test suite is well-structured and covers the key acceptance criteria (idempotency, regular-file conflict, foreign-symlink conflict, hooksPath preservation, force override). The overall approach is sound and the CONTRIBUTING.md documentation is accurate.

The most important concern is a silent config mutation that contradicts the fix's stated goal: the Apply section unconditionally runs \`git config --local --unset core.hooksPath\` (line 63) even when the preflight phase accepted \`.githooks\` as a non-conflicting value — a developer who explicitly set \`core.hooksPath=.githooks\` will have it silently cleared on every re-run with no stderr output. There is no test covering this scenario, so the regression is invisible to CI.

**Review Decision: COMMENT**

---
_Reviewed by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 5 inline comments, deduped 0._

scripts/setup.sh
lugassawan: Addressed in f6540dee: replaced unconditional `--unset` with a `case` guard that leaves `.githooks` and `$ROOT/.githooks` values in place; only unsets when the value is genuinely non-default. Also added `test_idempotent_with_whitelisted_hookspath` to lock in the contract.

scripts/setup.sh
lugassawan: Addressed in f6540dee: updated both occurrences (lines 37 and 65) to `# glob unexpanded — .githooks is empty or missing`.

scripts/setup.sh
lugassawan: Addressed in f6540dee: changed `set -e` to `set -eu`.

tests/test_setup.py
lugassawan: Addressed in f6540dee: added `test_idempotent_with_whitelisted_hookspath` which sets `core.hooksPath=.githooks`, runs setup, and asserts the config key is preserved.

scripts/setup.sh
```diff
@@ -1,15 +1,74 @@
...
+    printf '%s' "$conflicts" >&2
+    printf 'Re-run with --force to overwrite.\n' >&2
+    exit 1
+  fi
+  printf 'warning: --force: overwriting:\n' >&2
+  printf '%s' "$conflicts" >&2
+fi
+
+# --- Apply -----------------------------------------------------------------
+existing_hp_final="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
```
lugassawan: **Medium:** `existing_hp_final` re-reads `core.hooksPath` from git config, but `existing_hp` was already captured and normalized identically at lines 26–27. Nothing between preflight and Apply mutates this value, so the two reads are always equal. The double-read makes readers wonder if a mutation happened in between.

Simpler: reuse `existing_hp` directly:
```sh
# --- Apply -----------------------------------------------------------------
case "$existing_hp" in
  ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
esac
```

tests/test_setup.py
```diff
@@ -47,6 +47,92 @@ def repo_with_worktree(tmp_path):
...
+        repo, _ = repo_with_worktree
+        hooks_dir = repo / ".git" / "hooks"
+        hooks_dir.mkdir(parents=True, exist_ok=True)
+        hook = hooks_dir / "pre-push"
+        hook.write_text("#!/bin/sh\nexit 0\n")
+        result = self._run_setup(repo)
+        assert result.returncode != 0
+        assert not hook.is_symlink(), "hook must not have been replaced"
+        assert str(hook) in result.stderr
+        assert not (hooks_dir / "commit-msg").exists(), "no hooks should be installed after a conflict"
```
lugassawan: **Low:** The assertion message "no hooks should be installed after a conflict" is correct but could be more precise about _why_ — it implicitly relies on the script never partially applying hooks before detecting all conflicts. Tightening the message protects future readers if the preflight logic is ever refactored:

```python
assert not (hooks_dir / "commit-msg").exists(), "Apply section must not run if preflight found conflicts"
```

scripts/setup.sh
```diff
@@ -1,15 +1,74 @@
 #!/bin/sh
-set -e
+set -eu
+
+FORCE=0
+for arg in "$@"; do
+  case "$arg" in
+    --force|-f) FORCE=1 ;;
+    --help|-h)
+      echo "Usage: $0 [--force]"
```
lugassawan: **Low:** The `--help` block uses `echo` while the rest of the script uses `printf`. No behavioral impact, but replacing with `printf '%s\n'` would be consistent with the rest of the script.

lugassawan (commented):
## Review Summary

All four previously-flagged issues are correctly addressed. The preflight-then-apply separation in `setup.sh` is clean and the test suite now provides solid behavioral coverage of the preserve/overwrite contract — including the newly-added `test_idempotent_with_whitelisted_hookspath` which locks in the whitelisted-hooksPath invariant that was invisible to CI before. The only remaining note is a cosmetic redundancy in the Apply section (a second `git config` read that could reuse the already-captured `existing_hp`), which carries no correctness risk.

**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comments, deduped 0._

scripts/setup.sh
lugassawan: Addressed in ef0d40b: removed the redundant re-read of `core.hooksPath`; the Apply section now reuses `existing_hp` directly, eliminating the extra `git config` call and making it clear no mutation happened between preflight and apply.

tests/test_setup.py
lugassawan: Addressed in ef0d40b: updated assertion message to "Apply section must not run if preflight found conflicts".

scripts/setup.sh
lugassawan: Addressed in ef0d40b: replaced all five `echo` calls in the `--help` block with `printf` for consistency with the rest of the script.

Issue #99, opened 2026-05-02
[feat] New agent: dependency-auditor


lugassawan:
## Goal

A `dependency-auditor` subagent audits project dependencies for vulnerabilities, outdated versions, license compatibility issues, and transitive bloat.

## Motivation

Dependency hygiene spans multiple concerns (security, legal, maintenance burden) that no current agent addresses in a unified way. The `security-auditor` focuses on code vulnerabilities; no agent reads `package.json`, `Cargo.toml`, `go.mod`, or `requirements.txt` to evaluate the dependency graph itself.

## Surface

- [x] Subagent

## Sketch (optional)

`agents/dependency-auditor.md`

Scope:
- Outdated dependencies: semver age, deprecation signals
- Known vulnerability cross-reference (CVE/advisory databases)
- License compatibility: GPL/MIT/Apache compatibility matrix, copyleft risks
- Transitive dependency bloat: unused transitive deps, duplicate versions
- Lockfile drift: `package-lock.json` vs `package.json` inconsistencies

Manifest support: `package.json`, `Cargo.toml`, `go.mod`, `requirements.txt`, `pyproject.toml`

## Acceptance (optional)

- [ ] `agents/dependency-auditor.md` exists with valid frontmatter and tools list
- [ ] `scripts/validate.py` passes
- [ ] Agent is registered in `plugin.json`

PR #252, opened 2026-05-16
[chore] Script and test hygiene (closes #236)

### LINKED ISSUE #236: [chore] Script and test hygiene — setup glob edge case, validate re-reads, regex drift
## Product framing

**Problem.** Three scripts and a test file have low-severity but real edge-case bugs: `setup.sh` creates a dangling `hooks/*` symlink when `.githooks/` is empty; `test_pr_validation.py` reproduces the `pr.yml` PR-title regex as a Python string literal with no programmatic tie, so the two can silently drift; and `validate.py` re-reads every skill and agent file 2–3× per run.

**Value.** Fixing these avoids setup surprises in fresh clones with no githooks, prevents CI from silently accepting PR titles that the test regex no longer validates, and speeds up `validate.py` on large catalogs.

**Acceptance criteria.**
- [ ] `setup.sh` works correctly when `.githooks/` is empty (no dangling symlink created)
- [ ] `test_pr_validation.py` derives its title regex from the workflow YAML (or asserts structural equivalence) rather than maintaining a parallel copy
- [ ] `validate.py` reads each skill/agent file at most once per run (cached in a dict)
- [ ] `scripts/release.sh` empty-PR-list path is explicit (comment or assertion) to signal intent, not silent fallthrough

**Impact / Effort.** Impact: S — edge cases unlikely to hit in normal use but confusing when they do. Effort: S — each fix is localized to one function/block.

---

## Type

- [x] Release / packaging (`.claude-plugin/`, `scripts/`)

## Surface

**Files to change:**
- `scripts/setup.sh:10-13` — glob loop with no nullglob guard
- `tests/test_pr_validation.py` — hardcoded regex copy of `pr.yml` pattern
- `scripts/validate.py` — repeated file reads per check function
- `scripts/release.sh:128` — `gh pr list` empty-array edge path

## Description

**B6 — `setup.sh` glob creates dangling symlink when `.githooks/` is empty**
`for src in "$ROOT"/.githooks/*` in POSIX `sh` (no `nullglob`) expands to the literal string `…/.githooks/*` when the directory is empty. `basename` produces `*`; `ln -sfn` creates `$GIT_DIR/hooks/*` pointing nowhere. Fix: add a `[ -e "$src" ] || continue` guard, or switch to `find "$ROOT/.githooks" -maxdepth 1 -type f`.

**B7 — `test_pr_validation.py` duplicates `pr.yml` regex without a tie**
The test maintains a standalone Python regex string that mirrors the `validate-pr` step in `.github/workflows/pr.yml`. When the workflow regex is updated, the test silently diverges. Fix: have the test parse and extract the regex from `pr.yml` at runtime (e.g. `yaml.safe_load` + key path), so one source of truth drives both validation and test.

**O5 — `validate.py` re-reads skill/agent files 2–3× per run**
`check_skills()`, `check_agents()`, and related functions each open and read the same Markdown files independently. For a catalog with 40+ skills and 15 agents, this means 100+ redundant file opens per `validate.py` run. Fix: load all skill/agent files into a `dict[path, content]` once at startup and pass it as a parameter to each check function.

**B5 — `release.sh:128` empty-PR-list path**
`EXISTING_PR=$(gh pr list ... --jq '.[0]' 2>/dev/null || echo "null")` — when `gh pr list` returns `[]`, `.[0]` silently produces the JSON `null` string, which causes the `!= "null"` guard to trigger the else-branch without a visible signal. The logic is effectively correct but should be documented with an inline comment explaining the `null`-string contract, to avoid future regressions when the check is edited.

## Acceptance

- [ ] `bash scripts/setup.sh` in a repo with an empty `.githooks/` directory produces no symlinks and no error
- [ ] `pytest tests/test_pr_validation.py` fails (not passes) when the `pr.yml` title regex is changed without updating the test
- [ ] `python scripts/validate.py` reads each file path at most once (verifiable via `strace` or instrumentation)
- [ ] `release.sh:128` block has a comment documenting the `null`-string contract for `gh pr list`

lugassawan:
## Summary
- `scripts/setup.sh`: add `[ -e "$src" ] || continue` guard — POSIX sh leaves the literal `*` when `.githooks/` is empty, which previously created a dangling symlink named `*`
- `scripts/release.sh`: comment explaining the two independent paths that produce the `"null"` sentinel (jq empty-array path and gh-failure fallback)
- `tests/test_pr_validation.py`: new `TestPrYamlSync` class — extracts live regexes from `.github/workflows/pr.yml` via `re.DOTALL` anchors and asserts behavioural equivalence against the Python helpers; now covers all 4 patterns including the previously-unmirrored title regex
- `scripts/validate.py`: `_build_cache()` reads every `agents/*.md` and `skills/*/SKILL.md` exactly once; `cache=None` kwarg threaded through 6 check functions so per-file read count drops from 3–4× to 1×
- `tests/test_setup_sh.py`: new — subprocess-based tests for the glob guard (empty dir, absent dir, happy path, multi-hook)
- `tests/test_validate.py`: `TestFileReadCaching` monkeypatches `Path.read_text` to assert ≤1 read per path after the cache refactor

## Test Plan
- [x] `pytest tests/ -v` — 441 passed, 0 failed
- [x] `python3 scripts/validate.py` — All checks passed
- [x] `shellcheck scripts/setup.sh scripts/release.sh -S warning` — clean
- [x] Manual smoke test: empty `.githooks/` repo produces no `*` symlink
- [x] Mutation test: changing `pr.yml` title regex causes `TestPrYamlSync::test_title_pattern_locks_allowed_types` to fail noisily

### Type-tailored checks ([fix] / [refactor])
- [x] `setup.sh` — before fix: `ls .git/hooks/` shows dangling `*` symlink on empty `.githooks/`; after fix: symlink absent
- [x] `validate.py` — read count confirmed ≤1 per agent file and ≤1 per SKILL.md via `TestFileReadCaching`

Closes #236

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -46,8 +46,11 @@ def parse_frontmatter(path, text=None):
 def _build_cache():
     """Read every agent .md and every skills/*/SKILL.md exactly once.
 
-    Returns (agents, skills) where each is a dict[Path, str]. Paths that
-    cannot be read are omitted so per-check OSError handling is preserved.
+    Returns (agents, skills) where each is a dict[Path, str | None].
+    rglob("*.md") is intentional: it covers check_unwired_principle_skills
+    (which uses rglob) in addition to the flat-glob consumers.
+    Unreadable files are stored as None so consumers that track failures
+    (e.g. check_catalog_completeness) can report them without re-reading.
     ROOT is resolved inside this function so test monkeypatching of ROOT works.
     """
     agents_dir = ROOT / "agents"
@@ -58,12 +61,12 @@ def _build_cache():
         try:
             agents[p] = p.read_text(encoding="utf-8")
         except OSError:
-            pass
+            agents[p] = None  # sentinel: present but unreadable
     for p in skills_dir.glob("*/SKILL.md"):
         try:
             skills[p] = p.read_text(encoding="utf-8")
         except OSError:
-            pass
+            skills[p] = None  # sentinel: present but unreadable
     return agents, skills
 
 
@@ -327,9 +330,11 @@ def check_catalog_completeness(cache=None):
              f"stale entry 'swe-workbench:{sid}' has no skills/{sid}/ on disk")
 
     for agent_md in sorted(agents_dir.glob("*.md")):
-        cached = agents_cache.get(agent_md) if agents_cache is not None else None
-        if cached is not None:
-            agent_text = cached
+        if agents_cache is not None and agent_md in agents_cache:
+            agent_text = agents_cache[agent_md]
+            if agent_text is None:
+                fail(agent_md.relative_to(ROOT), "could not read file")
+                continue
         else:
             try:
                 agent_text = agent_md.read_text(encoding="utf-8")
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -52,6 +52,9 @@ def _build_cache():
     Unreadable files are stored as None so consumers that track failures
     (e.g. check_catalog_completeness) can report them without re-reading.
     ROOT is resolved inside this function so test monkeypatching of ROOT works.
+
+    Note: skills/*/templates/*.md files are NOT cached here; check_template_placeholders
+    reads each template file directly (one read_text() call per template).
     """
     agents_dir = ROOT / "agents"
     skills_dir = ROOT / "skills"
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -40,6 +40,34 @@ def parse_frontmatter(path, text=None):
 
 
 # ──────────────────────────────────────────────
+# File-read cache
+# ──────────────────────────────────────────────
+
+def _build_cache():
+    """Read every agent .md and every skills/*/SKILL.md exactly once.
+
+    Returns (agents, skills) where each is a dict[Path, str]. Paths that
+    cannot be read are omitted so per-check OSError handling is preserved.
+    ROOT is resolved inside this function so test monkeypatching of ROOT works.
+    """
+    agents_dir = ROOT / "agents"
+    skills_dir = ROOT / "skills"
+    agents: dict = {}
+    skills: dict = {}
+    for p in agents_dir.rglob("*.md"):
+        try:
+            agents[p] = p.read_text(encoding="utf-8")
+        except OSError:
+            pass
+    for p in skills_dir.glob("*/SKILL.md"):
+        try:
+            skills[p] = p.read_text(encoding="utf-8")
+        except OSError:
+            pass
+    return agents, skills
+
+
+# ──────────────────────────────────────────────
 # Validators
 # ──────────────────────────────────────────────
 
@@ -116,12 +144,14 @@ def check_hooks_json():
                     fail(path.relative_to(ROOT), f"hooks.{event}[{i}].hooks[{j}].command must be a string")
 
 
-def check_skills():
+def check_skills(cache=None):
     skills_dir = ROOT / "skills"
+    skills_cache = cache[1] if cache is not None else None
     # glob("*/SKILL.md") matches exactly depth-2 paths; no need for a post-hoc depth guard.
     for skill_md in sorted(skills_dir.glob("*/SKILL.md")):
         skill_dir_name = skill_md.parent.name
-        text = skill_md.read_text(encoding="utf-8")
+        text = (skills_cache.get(skill_md) if skills_cache is not None else None) \
+            or skill_md.read_text(encoding="utf-8")
         line_count = len(text.splitlines())  # total file length including frontmatter
         fm = parse_frontmatter(skill_md, text=text)
         if fm is None:
@@ -146,10 +176,12 @@ def check_skills():
             )
 
 
-def check_agents():
+def check_agents(cache=None):
     agents_dir = ROOT / "agents"
+    agents_cache = cache[0] if cache is not None else None
     for agent_md in sorted(agents_dir.glob("*.md")):
-        text = agent_md.read_text(encoding="utf-8")
+        text = (agents_cache.get(agent_md) if agents_cache is not None else None) \
+            or agent_md.read_text(encoding="utf-8")
         fm = parse_frontmatter(agent_md, text=text)
         if fm is None:
             fail(agent_md.relative_to(ROOT), "missing or malformed frontmatter")
@@ -172,13 +204,15 @@ def check_commands():
             fail(cmd_md.relative_to(ROOT), "frontmatter missing required field: 'description'")
 
 
-def check_agent_skill_refs():
+def check_agent_skill_refs(cache=None):
     """Every `swe-workbench:<id>` in agents/*.md must resolve to skills/<id>/ on disk."""
     agents_dir = ROOT / "agents"
     skills_dir = ROOT / "skills"
+    agents_cache = cache[0] if cache is not None else None
     pattern = re.compile(r'`swe-workbench:([\w-]+)`')
     for agent_md in sorted(agents_dir.glob("*.md")):
-        text = agent_md.read_text(encoding="utf-8")
+        text = (agents_cache.get(agent_md) if agents_cache is not None else None) \
+            or agent_md.read_text(encoding="utf-8")
         for skill_id in set(pattern.findall(text)):
             if not (skills_dir / skill_id).is_dir():
                 fail(
@@ -205,15 +239,17 @@ def check_command_skill_refs():
 TEMPLATE_MARKER_RE = re.compile(r'\[\[detect:([a-z][a-z0-9-]*)\]\]')
 
 
-def check_template_placeholders():
+def check_template_placeholders(cache=None):
     """Every [[detect:KEY]] in skills/*/templates/*.md must be documented in the
     adjacent SKILL.md's '## Project Detection' section as a backtick-wrapped key."""
     skills_dir = ROOT / "skills"
+    skills_cache = cache[1] if cache is not None else None
     for template in sorted(skills_dir.glob("*/templates/*.md")):
         skill_md = template.parent.parent / "SKILL.md"
         if not skill_md.is_file():
             continue
-        skill_text = skill_md.read_text(encoding="utf-8")
+        skill_text = (skills_cache.get(skill_md) if skills_cache is not None else None) \
+            or skill_md.read_text(encoding="utf-8")
         pd_idx = skill_text.find("## Project Detection")
         if pd_idx >= 0:
             next_h2 = skill_text.find("\n## ", pd_idx + 4)
@@ -261,11 +297,12 @@ def check_skill_trigger_fixtures():
                 )
 
 
-def check_catalog_completeness():
+def check_catalog_completeness(cache=None):
     """Catalog at agents/shared/skills.md must list every skill, and every agent must include it."""
     catalog = ROOT / "agents" / "shared" / "skills.md"
     skills_dir = ROOT / "skills"
     agents_dir = ROOT / "agents"
+    agents_cache = cache[0] if cache is not None else None
 
     if not catalog.is_file():
         fail(catalog.relative_to(ROOT), "missing — required catalog file")
@@ -275,7 +312,8 @@ def check_catalog_completeness():
         fail(skills_dir.relative_to(ROOT), "missing — required skills directory")
         return
 
-    text = catalog.read_text(encoding="utf-8")
+    text = (agents_cache.get(catalog) if agents_cache is not None else None) \
+        or catalog.read_text(encoding="utf-8")
     # — = EM DASH; [^\r\n]* avoids capturing CRLF carriage returns in description
     entry_re = re.compile(r'^-\s+`swe-workbench:([\w-]+)`\s+—\s+(\S[^\r\n]*)$', re.MULTILINE)
     catalog_ids = {sid for sid, _ in entry_re.findall(text)}
@@ -289,11 +327,15 @@ def check_catalog_completeness():
              f"stale entry 'swe-workbench:{sid}' has no skills/{sid}/ on disk")
 
     for agent_md in sorted(agents_dir.glob("*.md")):
-        try:
-            agent_text = agent_md.read_text(encoding="utf-8")
-        except OSError as e:
-            fail(agent_md.relative_to(ROOT), f"could not read file: {e}")
-            continue
+        cached = agents_cache.get(agent_md) if agents_cache is not None else None
+        if cached is not None:
+            agent_text = cached
+        else:
+            try:
+                agent_text = agent_md.read_text(encoding="utf-8")
+            except OSError as e:
+                fail(agent_md.relative_to(ROOT), f"could not read file: {e}")
+                continue
         if "@./shared/skills.md" not in agent_text:
             fail(agent_md.relative_to(ROOT),
                  "missing required '@./shared/skills.md' include"
@@ -318,7 +360,7 @@ def check_examples():
             )
 
 
-def check_unwired_principle_skills():
+def check_unwired_principle_skills(cache=None):
     """Every skills/principle-*/ must be referenced by at least one agent.
 
     agents/shared/skills.md (the catalog) is explicitly excluded — it lists
@@ -327,6 +369,7 @@ def check_unwired_principle_skills():
     skills_dir = ROOT / "skills"
     agents_dir = ROOT / "agents"
     catalog = agents_dir / "shared" / "skills.md"
+    agents_cache = cache[0] if cache is not None else None
 
     principle_skills = sorted(
         p.name for p in skills_dir.glob("principle-*")
@@ -337,7 +380,13 @@ def check_unwired_principle_skills():
 
     for skill_id in principle_skills:
         needle = f"`swe-workbench:{skill_id}`"
-        wired = any(needle in f.read_text(encoding="utf-8") for f in agent_files)
+        wired = any(
+            needle in (
+                (agents_cache.get(f) if agents_cache is not None else None)
+                or f.read_text(encoding="utf-8")
+            )
+            for f in agent_files
+        )
         if not wired:
             fail(
                 Path("agents") / "<unwired>",
@@ -355,18 +404,20 @@ def main():
     print("Validating swe-workbench plugin files...")
     print()
 
+    cache = _build_cache()
+
     plugin_data = check_plugin_json()
     check_marketplace_json(plugin_data)
     check_hooks_json()
-    check_skills()
+    check_skills(cache=cache)
     check_skill_trigger_fixtures()
-    check_agents()
+    check_agents(cache=cache)
     check_commands()
-    check_agent_skill_refs()
+    check_agent_skill_refs(cache=cache)
     check_command_skill_refs()
-    check_catalog_completeness()
-    check_template_placeholders()
-    check_unwired_principle_skills()
+    check_catalog_completeness(cache=cache)
+    check_template_placeholders(cache=cache)
+    check_unwired_principle_skills(cache=cache)
     check_examples()
 
     if FAILURES:
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -335,8 +335,17 @@ def check_catalog_completeness(cache=None):
         fail(skills_dir.relative_to(ROOT), "missing — required skills directory")
         return
 
-    text = (agents_cache.get(catalog) if agents_cache is not None else None) \
-        or catalog.read_text(encoding="utf-8")
+    if agents_cache is not None and catalog in agents_cache:
+        text = agents_cache[catalog]
+        if text is None:
+            fail(catalog.relative_to(ROOT), "could not read file")
+            return
+    else:
+        try:
+            text = catalog.read_text(encoding="utf-8")
+        except OSError as e:
+            fail(catalog.relative_to(ROOT), f"could not read file: {e}")
+            return
     # — = EM DASH; [^\r\n]* avoids capturing CRLF carriage returns in description
     entry_re = re.compile(r'^-\s+`swe-workbench:([\w-]+)`\s+—\s+(\S[^\r\n]*)$', re.MULTILINE)
     catalog_ids = {sid for sid, _ in entry_re.findall(text)}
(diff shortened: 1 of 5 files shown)


scripts/validate.py
```diff
@@ -39,6 +39,34 @@ def parse_frontmatter(path, text=None):
...
+    """
+    agents_dir = ROOT / "agents"
+    skills_dir = ROOT / "skills"
+    agents: dict = {}
+    skills: dict = {}
+    for p in agents_dir.rglob("*.md"):
+        try:
+            agents[p] = p.read_text(encoding="utf-8")
+        except OSError:
+            pass
```
lugassawan: **Medium** — `_build_cache` silently drops unreadable files with a bare `except OSError: pass`, breaking `check_catalog_completeness`'s original contract.

The original agent loop had an explicit `OSError` handler that called `fail(...)` and `continue` — collecting the error and moving on. With the cache in place, if a file was unreadable during `_build_cache`, it is absent from `agents_cache`. The consumer falls through to the bare `agent_md.read_text()`, which re-raises the `OSError` uncaught — killing the validator with a traceback instead of continuing past the failure.

Suggested fix — store `None` as a sentinel so consumers can handle the failure gracefully:
```python
for p in agents_dir.rglob("*.md"):
    try:
        agents[p] = p.read_text(encoding="utf-8")
    except OSError as e:
        agents[p] = None  # cache the failure; let the consumer decide
```
Then in `check_catalog_completeness`, treat `cached is None` as a read failure.

scripts/validate.py
```diff
@@ -39,6 +39,34 @@ def parse_frontmatter(path, text=None):
...
+
+    Returns (agents, skills) where each is a dict[Path, str]. Paths that
+    cannot be read are omitted so per-check OSError handling is preserved.
+    ROOT is resolved inside this function so test monkeypatching of ROOT works.
+    """
+    agents_dir = ROOT / "agents"
+    skills_dir = ROOT / "skills"
+    agents: dict = {}
+    skills: dict = {}
+    for p in agents_dir.rglob("*.md"):
```
lugassawan, line 63: **Medium** — `_build_cache` walks `agents_dir.rglob("*.md")` (recursive) but every flat-glob consumer (`check_agents`, `check_agent_skill_refs`, `check_catalog_completeness`) iterates `agents_dir.glob("*.md")` (depth-1 only).

There is no immediate correctness bug — `rglob` is a superset of `glob`, so all flat-glob paths are cached. But the cache silently pre-reads files no flat-glob consumer will ever look up (e.g. `agents/shared/skills.md`), and the depth asymmetry will confuse future authors comparing cache size vs. consumer iteration.

Options:
1. Switch to `glob("*.md")` if the intent is to cache only what flat-glob consumers see.
2. Keep `rglob` and update the docstring to document that it intentionally also covers `check_unwired_principle_skills` (which uses `rglob`).

tests/test_pr_validation.py
```diff
@@ -142,3 +169,125 @@ def test_fixes_with_number_matches(self):
...
+                f"yml_re and has_closing_keyword() disagree on {text!r}: "
+                f"yml={yml_result}, py={py_result}"
+            )
+            assert yml_result == expected, (
+                f"yml regex gave {yml_result!r} for {text!r}, expected {expected}"
+            )
+
+    def test_title_pattern_locks_allowed_types(self):
+        """PR title regex in pr.yml must accept only defined types."""
+        pattern = _extract_pr_yml_pattern(r"PATTERN='([^']+)'")
```
lugassawan: **Low** — The `PATTERN='([^']+)'` anchor has no surrounding context, making it fragile to a second `PATTERN=` variable in `pr.yml`.

Today there is exactly one `PATTERN=` assignment in the file. The inline comment on line 16 of `pr.yml` ("Keep in sync with .githooks/commit-msg") hints that a second `PATTERN=` for the commit-msg hook could appear in the future. `re.search` returns the first match and would silently lock the test to the wrong pattern.

The other three anchors are already scoped with context strings — this one should be too:
```suggestion
        pattern = _extract_pr_yml_pattern(r"Validate PR title.*?PATTERN='([^']+)'")
```

tests/test_setup_sh.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TestSetupShGlobGuard:
+    def test_empty_githooks_creates_no_symlinks(self, tmp_path):
+        """When .githooks/ exists but is empty, no symlinks are created."""
+        (tmp_path / ".githooks").mkdir()
+        result = _run_setup(tmp_path)
+        assert result.returncode == 0, result.stderr
+        hooks_dir = tmp_path / ".git" / "hooks"
+        symlinks = [p for p in hooks_dir.iterdir() if p.is_symlink()]
```
lugassawan: **Low** — `hooks_dir.iterdir()` will raise `FileNotFoundError` if git initialises without creating `.git/hooks/`.

Some minimal git installations (especially on CI runners or containers without a template directory) do not create `.git/hooks` during `git init`. If the directory is absent, the test crashes with `FileNotFoundError` rather than producing a useful assertion failure.

```suggestion
        symlinks = [p for p in hooks_dir.iterdir() if p.is_symlink()] if hooks_dir.exists() else []
```

tests/test_validate.py
```diff
@@ -908,3 +908,81 @@ def test_principle_dir_without_skill_md_is_ignored(self, reset_validate):
...
+
+        return read_counts
+
+    def test_each_agent_md_read_at_most_once(self, reset_validate, monkeypatch):
+        root = reset_validate
+        _make_full_valid_tree(root)
+        read_counts = self._count_reads(root, monkeypatch)
+
+        agents_dir = root / "agents"
+        agent_files = [p for p in agents_dir.glob("*.md")]
```
lugassawan: **Low** — `test_each_agent_md_read_at_most_once` collects agent files with `glob("*.md")` (depth-1) but the cache is built with `rglob("*.md")` (recursive).

If `agents/shared/skills.md` or any other subdirectory file is read more than once, the test misses it — it only validates the subset of files that flat-glob consumers would iterate. This is the same depth asymmetry noted in `_build_cache`; the test coverage follows the bug.

lugassawan (commented):
## Review Summary

This PR delivers all four hygiene fixes called for in #236 with solid test coverage: the `setup.sh` glob guard, the `pr.yml` sync tests, the `validate.py` read-caching layer, and the `release.sh` null-sentinel comment. The approach on each fix is conservative and well-reasoned. The sync-test anchors extract the correct patterns from `pr.yml` at runtime; the caching refactor correctly handles the cache-miss fallback path for the common case.

Two medium findings worth addressing before merge: `_build_cache` uses a bare `except OSError: pass` that silently drops unreadable files, which breaks `check_catalog_completeness`'s original contract of collecting failures and continuing (a bare re-raise now propagates instead); and the `rglob` vs `glob` depth asymmetry between the cache builder and its flat-glob consumers creates latent confusion about what is and isn't cached.

**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 5 inline comments, deduped 0._

scripts/validate.py
```diff
@@ -116,12 +147,14 @@ def check_hooks_json():
...
 
-def check_skills():
+def check_skills(cache=None):
     skills_dir = ROOT / "skills"
+    skills_cache = cache[1] if cache is not None else None
     # glob("*/SKILL.md") matches exactly depth-2 paths; no need for a post-hoc depth guard.
     for skill_md in sorted(skills_dir.glob("*/SKILL.md")):
         skill_dir_name = skill_md.parent.name
-        text = skill_md.read_text(encoding="utf-8")
+        text = (skills_cache.get(skill_md) if skills_cache is not None else None) \
```
lugassawan: **High** — Five `or`-fallback cache lookups bypass the `None` sentinel introduced to track unreadable files, defeating the fix for the previous Medium finding.

`_build_cache()` stores `None` for files that raise `OSError`. But the consumers in `check_skills`, `check_agents`, `check_agent_skill_refs`, `check_template_placeholders`, and `check_unwired_principle_skills` all use:
```python
text = (cache.get(k) if cache is not None else None) or file.read_text(...)
```
Since `None or expr` evaluates `expr`, a cached `None` sentinel silently triggers a re-read on every consumer call — breaking the single-read guarantee. Worse, if the re-read also fails, the `OSError` propagates uncaught. In `check_unwired_principle_skills` (line 390), `needle in None` raises `TypeError` before even reaching `read_text`.

`check_catalog_completeness` already shows the correct pattern: explicit `if agent_md in agents_cache` + `if agent_text is None: fail(); continue`. The five other consumers should be updated to match.

For `check_unwired_principle_skills` the generator makes inline guarding awkward — consider pre-filtering `agent_files` to skip `None`-valued entries:
```python
agent_files = [f for f in sorted(agents_dir.rglob("*.md")) if f != catalog
               and (agents_cache is None or agents_cache.get(f) is not None)]
```

tests/test_validate.py
```diff
@@ -908,3 +908,81 @@ def test_principle_dir_without_skill_md_is_ignored(self, reset_validate):
...
+        monkeypatch.setattr(Path, "read_text", counting_read_text)
+
+        try:
+            validate.main()
+        except SystemExit:
+            pass  # we care about read counts, not pass/fail
+
+        return read_counts
+
+    def test_each_agent_md_read_at_most_once(self, reset_validate, monkeypatch):
```
lugassawan, line 958: **Medium** — `TestFileReadCaching` only exercises readable files; the `None`-sentinel bypass (High finding above) is not caught by any test.

Both caching tests use `_make_full_valid_tree()`, where all files are readable. `_build_cache()` therefore never stores `None`, so the `or`-fallback in the five consumer functions is never exercised with a sentinel value.

A test variant that makes one agent file unreadable (e.g. patch `Path.read_text` to raise `OSError` for one specific path during the cache build, then allow it to succeed on subsequent calls) would have caught the regression and would document the expected behavior:
- `main()` reports exactly one failure for the unreadable file.
- No unhandled `OSError` or `TypeError` escapes.
- All other agent files are still checked normally.

tests/test_pr_validation.py
```diff
@@ -142,3 +169,125 @@ def test_fixes_with_number_matches(self):
...
+        )
+        yml_re = re.compile(raw, re.IGNORECASE)
+
+        samples = [
+            ("Closes #42",         True),
+            ("closes #1",          True),
+            ("Fixes #99",          True),
+            ("Fixed #7",           True),
+            ("Resolves #100",      True),
+            ("resolved #99",       True),   # resolve[sd]? matches 'resolved'
```
lugassawan: **Low** — The comment on line 239 notes that `resolve[sd]?` matches `'resolved'`, but doesn't mention that the same regex also matches `'resolves'`. A future reader updating the sample table might miss this and omit it.

Consider adding a corresponding positive sample to make coverage explicit:
```suggestion
            ("resolved #99",       True),   # resolve[sd]? matches 'resolved' and 'resolves'
            ("resolves #99",       True),
```

lugassawan (commented):
## Review Summary

All five prior findings were addressed and the test additions — `TestPrYamlSync`, `TestSetupShGlobGuard`, and `TestFileReadCaching` — represent real quality improvements. However, the `None`-sentinel fix introduced a correctness regression: five of the six cache consumers use an `or`-fallback that treats a cached `None` identically to a cache miss, silently re-reading unreadable files on every call and propagating uncaught `OSError` (or `TypeError` in the generator case). `check_catalog_completeness` already demonstrates the correct explicit-`in`-check pattern; the other five consumers need the same treatment. The existing caching tests only exercise readable files and therefore do not catch this regression.

**Review Decision: COMMENT**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comments, deduped 0._

scripts/validate.py
```diff
@@ -275,7 +335,8 @@ def check_catalog_completeness():
         fail(skills_dir.relative_to(ROOT), "missing — required skills directory")
         return
 
-    text = catalog.read_text(encoding="utf-8")
+    text = (agents_cache.get(catalog) if agents_cache is not None else None) \
+        or catalog.read_text(encoding="utf-8")
```
lugassawan: **High** — One `or`-fallback survived the fix: the catalog file read in `check_catalog_completeness` (line 338–339).

`_build_cache()` uses `rglob("*.md")` so `agents/shared/skills.md` lands in `agents_cache` as `None` when unreadable. The surviving expression `(agents_cache.get(catalog) if ...) or catalog.read_text(...)` treats `None` as a cache miss and re-reads — raising an uncaught `OSError` that crashes `main()` instead of recording a tidy `fail()` entry. This is the same class of bug as the five that were fixed this round.

Replace with the explicit guard already used for the per-agent loop below it:
```python
if agents_cache is not None and catalog in agents_cache:
    text = agents_cache[catalog]
    if text is None:
        fail(catalog.relative_to(ROOT), "could not read file")
        return
else:
    try:
        text = catalog.read_text(encoding="utf-8")
    except OSError as e:
        fail(catalog.relative_to(ROOT), f"could not read file: {e}")
        return
```

tests/test_validate.py
```diff
@@ -908,3 +908,112 @@ def test_principle_dir_without_skill_md_is_ignored(self, reset_validate):
...
+        assert skill_files, "Expected at least one SKILL.md"
+
+        for skill_path in skill_files:
+            count = read_counts.get(str(skill_path.resolve()), 0)
+            assert count <= 1, (
+                f"{skill_path.parent.name}/SKILL.md was read {count} times; expected ≤1 "
+                "(check_skills and check_template_placeholders share the same cache)"
+            )
+
+    def test_unreadable_agent_cached_as_failure(self, reset_validate, monkeypatch):
```
lugassawan, line 991: **Medium** — `test_unreadable_agent_cached_as_failure` patches the first file from `agents_dir.glob("*.md")` (flat, depth-1), which never yields `agents/shared/skills.md`. The surviving High bug above (catalog `or`-fallback) therefore has zero test coverage.

A companion test that patches the catalog file's `read_text` to raise `OSError` would both document the expected behavior and have caught this regression:
```python
def test_unreadable_catalog_cached_as_failure(self, reset_validate, monkeypatch):
    root = reset_validate
    _make_full_valid_tree(root)
    catalog = root / "agents" / "shared" / "skills.md"
    original = Path.read_text
    read_count = {"n": 0}
    def patched(self_path, *args, **kwargs):
        if self_path.resolve() == catalog.resolve():
            read_count["n"] += 1
            raise OSError("simulated")
        return original(self_path, *args, **kwargs)
    monkeypatch.setattr(Path, "read_text", patched)
    try:
        validate.main()
    except SystemExit:
        pass
    assert read_count["n"] == 1
    rel = str(catalog.relative_to(root))
    assert any(rel in entry for entry in validate.FAILURES)
```

tests/test_pr_validation.py
```diff
@@ -5,6 +5,33 @@
...
 """
 import re
+from pathlib import Path
+
+import pytest
+
+_PR_YML_PATH = Path(__file__).parent.parent / ".github" / "workflows" / "pr.yml"
+try:
+    _PR_YML_TEXT: str | None = _PR_YML_PATH.read_text(encoding="utf-8")
+except FileNotFoundError:
```
lugassawan: **Low** — `except FileNotFoundError` is narrower than necessary. Any other `OSError` subclass (e.g. `PermissionError` on a locked CI runner) would propagate at import time and surface as a confusing collection error instead of a skip.

```suggestion
except OSError:
```

tests/test_setup_sh.py
```diff
@@ -0,0 +1,76 @@
...
+# Strip GIT_* vars so hook-context env doesn't leak into ephemeral test repos.
+# When tests run inside a git pre-push hook, GIT_DIR points to the host repo —
+# setup.sh's `git rev-parse` would then target the host instead of tmp_path.
+_CLEAN_ENV = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+
+
+def _run_setup(tmp_path: Path) -> subprocess.CompletedProcess:
+    """Run setup.sh inside a minimal git repo rooted at tmp_path."""
+    subprocess.run(
+        ["git", "init", str(tmp_path)], check=True, capture_output=True, env=_CLEAN_ENV
```
lugassawan: **Low** — `subprocess.run(["git", "init", ...], check=True, capture_output=True)` suppresses stderr on failure; if `git` is absent from `PATH` the test crashes with a generic `CalledProcessError` that doesn't show the underlying cause. The existing `check=True` is correct (it raises), but `capture_output=True` hides the reason.

```suggestion
    init = subprocess.run(
        ["git", "init", str(tmp_path)], capture_output=True, env=_CLEAN_ENV
    )
    assert init.returncode == 0, init.stderr.decode(errors="replace")
```

lugassawan (commented):
## Review Summary

The three round-2 findings were addressed correctly: all five `or`-fallback lookups in per-agent/skill loops now use explicit `if k in cache` guards, the sentinel test correctly verifies exactly one `read_text` call and a `FAILURES` entry, and `("resolves #99", True)` was added to the closing-keyword table. One instance of the `or`-fallback survived in `check_catalog_completeness` — the catalog file read at line 338–339 — where `agents_cache.get(catalog)` returning `None` still falls through to an uncaught `catalog.read_text()` re-raise; the sentinel test does not cover this path because it patches flat-glob agent files, not the subdirectory catalog.

**Review Decision: COMMENT**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 4 inline comments, deduped 0._

scripts/validate.py
```diff
@@ -39,6 +39,37 @@ def parse_frontmatter(path, text=None):
...
+
+def _build_cache():
+    """Read every agent .md and every skills/*/SKILL.md exactly once.
+
+    Returns (agents, skills) where each is a dict[Path, str | None].
+    rglob("*.md") is intentional: it covers check_unwired_principle_skills
+    (which uses rglob) in addition to the flat-glob consumers.
+    Unreadable files are stored as None so consumers that track failures
+    (e.g. check_catalog_completeness) can report them without re-reading.
+    ROOT is resolved inside this function so test monkeypatching of ROOT works.
```
lugassawan, line 54: **Medium** — `_build_cache()` docstring says it reads "every agent `.md` and every `skills/*/SKILL.md`" but does not mention that `skills/*/templates/*.md` files remain uncached. `check_template_placeholders` (line 282) still reads each template file directly, one `read_text()` per template.

This is not a bug — there are only 2 templates today — but the docstring creates a false impression that all skill-related reads go through the cache, which will mislead the next person auditing reads or extending coverage.

**Fix:** extend the docstring:
```python
# Note: skills/*/templates/*.md files are NOT cached here;
# check_template_placeholders reads each template file directly.
```
Alternatively, add a `templates` dict to the cache tuple and extend `test_each_skill_md_read_at_most_once` — more robust if templates proliferate.

tests/test_validate.py
```diff
@@ -908,3 +908,142 @@ def test_principle_dir_without_skill_md_is_ignored(self, reset_validate):
...
+class TestFileReadCaching:
+    """After the cache refactor, main() reads each agent and skill file exactly once."""
+
+    def _count_reads(self, root: Path, monkeypatch):
+        """Patch Path.read_text, run main(), return per-path read counts."""
+        read_counts: dict[str, int] = {}
+        original = Path.read_text
+
+        def counting_read_text(self_path, *args, **kwargs):
+            key = str(self_path.resolve())
```
lugassawan, line 944: **Low** — `_count_reads()` monkeypatches `Path.read_text` globally, so every `read_text` call during `validate.main()` — including any reads by pytest internals or future code paths outside `tmp_path` — lands in `read_counts`. Today the test tree assertions only look up paths under `tmp_path`, so this is harmless, but the count dict accumulates noise that can confuse a debugging session.

**Fix:** filter inside `counting_read_text` to only count reads under the test root:
```python
def counting_read_text(self_path, *args, **kwargs):
    key = str(self_path.resolve())
    if str(root) in key:  # only count reads under the test tree
        read_counts[key] = read_counts.get(key, 0) + 1
    return original(self_path, *args, **kwargs)
```

tests/test_pr_validation.py
```diff
@@ -142,3 +169,126 @@ def test_fixes_with_number_matches(self):
...
+    (or vice-versa), these tests fail with a clear diagnostic.
+    """
+
+    def test_html_stripper_matches_pr_yml(self):
+        """strip_html_comments() must use the exact same regex as pr.yml."""
+        # Anchor includes 'python3 -c' context so a future re.sub addition
+        # above line 61 in pr.yml doesn't silently redirect extraction.
+        extracted = _extract_pr_yml_pattern(r"python3 -c.*?re\.sub\(r'([^']+)'")
+        # No POSIX classes — pattern is byte-identical between shell and Python.
+        assert extracted == r"<!--.*?(?:-->|$)", (
```
lugassawan, line 187: **Low** — The assertion message on divergence (`"pr.yml HTML-stripper pattern {extracted!r} diverged from the pattern used in strip_html_comments() — sync one of them."`) does not indicate which side needs to change. When this test fails in CI, the engineer has to infer whether to update the Python mirror or the pr.yml source.

**Fix:** append a directional hint, e.g.:
```python
f"pr.yml HTML-stripper pattern {extracted!r} diverged from the "
"pattern used in strip_html_comments() — "
"update strip_html_comments() or the hardcoded expected string in this test."
```

lugassawan (commented):
## Review Summary

This PR is clean. The caching refactor in `validate.py` is well-designed — `_build_cache()` correctly uses `rglob(".md")` to pull the catalog into `agents_cache` alongside flat-glob consumers, the `None` sentinel is properly threaded through all six consumers with explicit `if key in cache` guards (no `or`-fallback regressions), and the fallback-to-direct-read paths remain correct for the no-cache call sites. The three new test files provide meaningful behavioral coverage: the `setup.sh` tests use real subprocess isolation with proper `GIT_*` env-var stripping, the `pr.yml` sync tests anchor extraction to stable textual patterns and cross-validate both the yml regex and Python mirror against the same sample suite, and the caching tests directly verify the "at most once" contract using a counting monkeypatch.

**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comments, deduped 0._

PR #253, opened 2026-05-16
[feat] Add /swe-workbench:report-issue command

### LINKED ISSUE #226: [feat] /report-issue command to file feedback into the swe-workbench plugin repo from any workspace
## Product framing

- **Problem.** Users running the swe-workbench plugin inside their own projects hit plugin bugs or have feature ideas, but `/swe-workbench:capture` auto-detects the *current* repo and files there — so feedback about the plugin lands in the user's own private project tracker, not in the plugin's issue tracker. Switching to the plugin checkout to file an issue is enough friction that most feedback never gets reported.
- **Value.** Plugin maintainers receive structured user feedback from real-world usage without users having to context-switch. Users get a one-line "report this upstream" affordance.
- **Acceptance criteria.**
  - Given a user is in any working directory, when they invoke `/report-issue <thought>`, the command produces a preview-confirm flow identical to `/capture` but with `Filing into: lugassawan/swe-workbench` hardcoded.
  - The command refuses to run if `gh auth status` shows the user lacks push/issue scope on `lugassawan/swe-workbench` (or surfaces a clear "you may not have access" warning and proceeds).
  - Duplicate scan and template discovery both run against `lugassawan/swe-workbench`, not the local repo.
  - The filed issue's footer includes plugin version (`.claude-plugin/plugin.json`) and Claude Code version, captured automatically so users don't have to fill them in.
- **Impact / Effort (RICE-lite).** Impact: M — unlocks a steady stream of in-the-wild feedback that's currently lost. Effort: S — `commands/report-issue.md` can be a 30-line variant of `commands/capture.md` that overrides repo detection.

## Goal

Add a `/report-issue` slash command that files a GitHub issue into `lugassawan/swe-workbench` regardless of the user's current working directory, so plugin feedback can be reported from anywhere.

## Motivation

`/swe-workbench:capture` is repo-agnostic by design (`gh repo view` in step 1 of `commands/capture.md`). That's correct for capturing thoughts about the *user's* project, but it means feedback about the plugin itself either gets filed in the wrong repo or doesn't get filed at all.

## Surface

- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Skill — `workflow-*`
- [x] Subagent (reuses `product-manager`)
- [x] Slash command (`/report-issue`)
- [ ] Hook
- [ ] Documentation (README mention + command catalog)
- [ ] Build / CI / release

## Sketch (optional)

```
$ pwd
/home/me/my-private-project
$ /report-issue The /capture preview eats my command history when the timestamp collides
[product-manager preview]
Filing into: lugassawan/swe-workbench
Template: bug_report.md
Title: [bug] /capture preview overwrites existing /tmp file on timestamp collision
...
Reply 'confirm' to file
```

Implementation outline:
- New `commands/report-issue.md` that delegates to `product-manager` with the same 8-step protocol as `/capture`, but replaces step 1 with: `repo="lugassawan/swe-workbench"; gh repo view "$repo" --json nameWithOwner` — no auto-detection.
- All subsequent `gh issue list` / `gh issue create` calls pass `--repo "$repo"` explicitly.
- Auto-append a footer block with `claude --version` and the plugin version from `.claude-plugin/plugin.json` (read via `gh api`).

## Acceptance (optional)

- [ ] `commands/report-issue.md` exists and is discoverable via the plugin's command catalog
- [ ] Invoking from outside the swe-workbench checkout files into `lugassawan/swe-workbench`
- [ ] Preview shows the hardcoded repo, not the user's current repo
- [ ] Filed issue includes plugin + CLI version footer

lugassawan:
## Summary
- Add `/swe-workbench:report-issue` command — a sibling to `/capture` that hardcodes `lugassawan/swe-workbench` as the filing target, so users can report plugin bugs or ideas from any working directory without switching repos.
- Auto-appends a version footer (plugin version from cache glob + `claude --version`) to every drafted body so maintainers get version-tagged reports without user effort.
- Supports blank-argument mode: scans the current conversation for plugin-related thoughts, falls back to `MEMORY.md`, proposes 1–3 candidates, and exits cleanly when nothing relevant is found.
- Suspends the `product-manager` agent's no-`--repo` rule for this command only (override documented in command body; agent file unchanged so its existing tests remain green).
- Adds 6 static-text tests in `tests/test_report_issue_command.py` and updates `README.md` / `docs/catalog.md`.

## Test Plan
- [x] `bash scripts/validate.sh` — exits 0, no error mentioning `commands/report-issue.md`
- [x] `pytest tests/test_report_issue_command.py -v` — 6/6 passed
- [x] `pytest tests/ -q` — 466 passed, zero regressions in `test_issue_filing_label_discovery.py`
- [ ] Manual smoke: `/swe-workbench:report-issue test thought` from `/tmp` — preview shows `Filing into: lugassawan/swe-workbench` and `--repo lugassawan/swe-workbench` in the Command line
- [ ] Manual smoke: reply `confirm` to preview — issue filed into `lugassawan/swe-workbench` with version footer present
- [ ] Manual smoke: `/swe-workbench:report-issue` (no arg) in a session with plugin-related context — candidate thoughts surface; selecting one proceeds to restatement

- [ ] New behaviour (`report-issue` blank-arg flow) tested in a session with real plugin-related conversation context
- [ ] No regression on `/capture` — the agent's no-`--repo` rule still applies for that command (verify by checking `product-manager.md` is unmodified)

Closes #226


commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
+---
+description: File a plugin bug or feature request directly into lugassawan/swe-workbench from any directory
+argument-hint: "[<one-line thought> — omit to draft from conversation/memory]"
+---
+
+> **Override (this command only):** The `product-manager` agent's default rules "Does not pass `--repo` explicitly to `gh issue create`" (Decision boundaries) and "Never pass `--repo`" (Mutation rules) are **suspended** for `/report-issue`. You MUST pass `--repo lugassawan/swe-workbench` on every `gh` invocation in this flow. No other rule is changed.
```
lugassawan, line 6: **High:** The override block cannot reliably suspend `product-manager.md`'s hard Mutation Rule.

The agent file `agents/product-manager.md` contains an authoritative `## Mutation rule` that reads: *"Never pass `--repo` — rely on the detected current-repo context."* This override note in `report-issue.md` attempts to cancel it from the caller side, but the agent loads its own system prompt intact. Depending on how the runtime resolves conflicting directives, the agent may silently ignore the override and omit `--repo`, causing `gh issue create` to target the user's local repo instead of `lugassawan/swe-workbench` — exactly the failure mode this command is designed to prevent.

**Suggested fix:** Update `agents/product-manager.md`'s Mutation rule to allow `--repo` when the calling command explicitly specifies a target repo — e.g. *"Never pass `--repo` unless the calling command explicitly directs a specific target repo."* The override block here is a useful belt-and-suspenders hint but cannot substitute for the agent's own rule permitting the flag.

commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
...
+   ```
+   I found a few plugin-related thoughts from this session/memory. Pick one or type your own:
+
+   1. <one-line framing>
+   2. <one-line framing>
+   3. <one-line framing>
+
+   Reply with 1/2/3 or type your own thought.
+   ```
+   Wait for the user's reply. The selected or typed thought becomes `$ARGUMENTS` and flow continues at **Step 2**.
```
lugassawan: **Medium:** "flow continues at **Step 2**" references a non-existent step label.

The blank-argument routing's sub-step 3 says flow continues at **Step 2**, but the delegation block below uses plain numbered items (`1.`, `2.`, …), not labelled as "Step 1", "Step 2", etc. An agent following this literally may loop back into blank-arg sub-step 2 (memory scan) rather than jumping to the product-manager delegation's `Auth + repo check` item.

**Suggested fix:** Replace the reference with something unambiguous, e.g. *"flow continues at the **Auth + repo check** step below"*, or add an explicit `## Step 1 — Auth + repo check` header to the delegation block to match the `## Step 0` style.

commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
...
+
+   Append the following footer to the body's last free-text section (or as a final `---` block):
+
+   ```markdown
+   ---
+   _Reported via `/swe-workbench:report-issue` — plugin v<version>, Claude Code <cli-version>._
+   ```
+
+   **Version capture (run once, before drafting):**
+   - **Plugin version:** list `~/.claude/plugins/cache/swe-workbench/swe-workbench/` version directories, sort semantically (`sort -V`), take the highest with `tail -1`, then read `plugin.json` from it: `ls ~/.claude/plugins/cache/swe-workbench/swe-workbench/ | sort -V | tail -1 | xargs -I{} python3 -c "import json; print(json.load(open('$HOME/.claude/plugins/cache/swe-workbench/swe-workbench/{}/.claude-plugin/plugin.json'))['version'])"`. If this returns no output (not installed from cache), fall back to: `gh api repos/lugassawan/swe-workbench/contents/.claude-plugin/plugin.json --repo lugassawan/swe-workbench --jq '.content' | base64 -d | python3 -c "import sys,json; print(json.load(sys.stdin)['version'])"`.
```
lugassawan: **Medium:** The `ls` call for the cache path emits an error to stderr in non-cache installs, which is confusing.

When the plugin is installed from source rather than the cache, `ls ~/.claude/plugins/cache/swe-workbench/swe-workbench/ 2>&1` will print `ls: cannot access ...: No such file or directory` to stderr before the pipeline silently falls back to the API path. Users see a misleading error even though the fallback works correctly.

**Suggested fix:** Add `2>/dev/null` to suppress the directory-not-found error so the silent fallback is truly silent:
```bash
ls ~/.claude/plugins/cache/swe-workbench/swe-workbench/ 2>/dev/null | sort -V | tail -1 | ...
```

commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
...
+
+2. **Restatement.** One sentence in the user's domain language confirming the thought. If the thought is ambiguous, ask exactly one clarifying question before continuing.
+
+3. **Product framing** — four lenses applied with brevity:
+   - **Problem.** User pain or constraint, stated as the user's pain — not the feature.
+   - **Value.** Who benefits and how. One "so-what" sentence.
+   - **Acceptance criteria.** 2–4 bullets (Given/When/Then or simple bullets).
+   - **Impact / Effort (RICE-lite).** `Impact: S/M/L` and `Effort: S/M/L`, one sentence each.
+
+4. **Template discovery.** Fetch templates via `gh api repos/lugassawan/swe-workbench/contents/.github/ISSUE_TEMPLATE --repo lugassawan/swe-workbench --jq '.[].name'` to list template file names (skip `config.yml`). For each `.md` template, fetch its body with `gh api repos/lugassawan/swe-workbench/contents/.github/ISSUE_TEMPLATE/<name> --repo lugassawan/swe-workbench --jq '.content' | base64 -d`. Read the frontmatter and first ~20 body lines. Classify the thought into the closest-fit template with a one-sentence reason, or note "No issue templates found; using default body shape" when none exist.
```
lugassawan: **Low:** `--repo` is redundant (and potentially mispositioned) on `gh api` calls where the repo is already encoded in the URL path.

The `gh api repos/lugassawan/swe-workbench/contents/.github/ISSUE_TEMPLATE --repo lugassawan/swe-workbench --jq '.[].name'` call passes `--repo` after the positional URL argument. Some older `gh` 2.x versions require `--repo` before the path. More importantly, the flag is redundant here — the full path already contains the owner and repo. Same applies to the template body fetch on the next line.

**Suggested fix:** Drop `--repo lugassawan/swe-workbench` from the two `gh api repos/...` calls (the URL is self-contained). Keep it on `gh issue create`, `gh issue list`, `gh label list`, and `gh repo view` where it's required.

commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
...
+   Append the following footer to the body's last free-text section (or as a final `---` block):
+
+   ```markdown
+   ---
+   _Reported via `/swe-workbench:report-issue` — plugin v<version>, Claude Code <cli-version>._
+   ```
+
+   **Version capture (run once, before drafting):**
+   - **Plugin version:** list `~/.claude/plugins/cache/swe-workbench/swe-workbench/` version directories, sort semantically (`sort -V`), take the highest with `tail -1`, then read `plugin.json` from it: `ls ~/.claude/plugins/cache/swe-workbench/swe-workbench/ | sort -V | tail -1 | xargs -I{} python3 -c "import json; print(json.load(open('$HOME/.claude/plugins/cache/swe-workbench/swe-workbench/{}/.claude-plugin/plugin.json'))['version'])"`. If this returns no output (not installed from cache), fall back to: `gh api repos/lugassawan/swe-workbench/contents/.claude-plugin/plugin.json --repo lugassawan/swe-workbench --jq '.content' | base64 -d | python3 -c "import sys,json; print(json.load(sys.stdin)['version'])"`.
+   - **CLI version:** `claude --version` — strip the leading `Claude Code ` prefix to get the bare semver.
```
lugassawan, line 75: **Low:** `base64 -d` in the API fallback is not portable on macOS.

macOS ships BSD `base64` which uses `-D` (capital D); GNU coreutils (Linux, Homebrew) uses `-d`. The primary cache path avoids this, but the fallback `gh api ... | base64 -d | python3 ...` will fail silently on macOS when the cache path is absent.

**Suggested fix:** Use a pure-Python decoder instead, which is portable across all platforms:
```bash
gh api repos/lugassawan/swe-workbench/contents/.claude-plugin/plugin.json --repo lugassawan/swe-workbench --jq '.content' | python3 -c "import base64,sys,json; print(json.loads(base64.b64decode(sys.stdin.read()))['version'])"
```

lugassawan (commented):
## Review Summary

This PR is approve-lean overall. The core feature is well-scoped and correctly delivers against all five acceptance criteria: hardcoded repo target, auth warning, duplicate scan and template discovery against the plugin repo, version footer, and catalog discoverability. The preview-confirm gate, `.cmd` sidecar pattern, and blank-argument memory-scan are thoughtful additions that go beyond the spec. The most important concern is a silent runtime conflict: `product-manager.md`'s `## Mutation rule` still reads "Never pass `--repo`", which the override block in `report-issue.md` attempts to cancel at the calling-command level — but an LLM interpreting the agent's own constraints may refuse to comply regardless of the caller's directive. The second concern is a step-numbering gap in the main flow that leaves the agent without a clear resume point after blank-argument routing.

**Review Decision: COMMENT**

---
_Reviewed by `reviewer` (swe-workbench). Posted 5 inline comments, deduped 0._

commands/report-issue.md
lugassawan: Addressed in 07dd37f8f0c0e7fb69cd4ad24a30cf4b0f4846d4: updated `agents/product-manager.md` Decision boundaries (line 88) and Mutation rule (line 108) to permit `--repo` when the calling command explicitly directs a specific target repo. The agent now has its own authority to comply — the override block in `report-issue.md` is belt-and-suspenders on top.

commands/report-issue.md
lugassawan: Addressed in 07dd37f8f0c0e7fb69cd4ad24a30cf4b0f4846d4: changed "flow continues at **Step 2**" to "flow continues at the **Auth + repo check** step below" — unambiguously points to the first item of the product-manager delegation block.

commands/report-issue.md
lugassawan: Addressed in 07dd37f8f0c0e7fb69cd4ad24a30cf4b0f4846d4: added `2>/dev/null` to the `ls` call so non-cache installs fall through silently. Also replaced both `base64 -d` occurrences with the portable Python decoder while editing this line (covers thread #5 too).

commands/report-issue.md
lugassawan: Addressed in 07dd37f8f0c0e7fb69cd4ad24a30cf4b0f4846d4: dropped `--repo lugassawan/swe-workbench` from both `gh api repos/...` calls in step 4 (template listing and body fetch) — the full URL path already identifies the repo. Also took the opportunity to replace the `base64 -d` pipe in the template body fetch with the portable Python decoder (covers thread #5 there too).

commands/report-issue.md
lugassawan: Addressed in 07dd37f8f0c0e7fb69cd4ad24a30cf4b0f4846d4: replaced all `base64 -d` occurrences with `python3 -c "import base64,sys,..."` — portable across macOS (BSD base64) and Linux (GNU base64) without external tooling.

commands/report-issue.md
```diff
@@ -0,0 +1,94 @@
...
+
+2. **Restatement.** One sentence in the user's domain language confirming the thought. If the thought is ambiguous, ask exactly one clarifying question before continuing.
+
+3. **Product framing** — four lenses applied with brevity:
+   - **Problem.** User pain or constraint, stated as the user's pain — not the feature.
+   - **Value.** Who benefits and how. One "so-what" sentence.
+   - **Acceptance criteria.** 2–4 bullets (Given/When/Then or simple bullets).
+   - **Impact / Effort (RICE-lite).** `Impact: S/M/L` and `Effort: S/M/L`, one sentence each.
+
+4. **Template discovery.** Fetch templates via `gh api repos/lugassawan/swe-workbench/contents/.github/ISSUE_TEMPLATE --jq '.[].name'` to list template file names (skip `config.yml`). For each `.md` template, fetch its body with `gh api repos/lugassawan/swe-workbench/contents/.github/ISSUE_TEMPLATE/<name> --jq '.content' | python3 -c "import base64,sys; print(base64.b64decode(sys.stdin.read()).decode())"`. Read the frontmatter and first ~20 body lines. Classify the thought into the closest-fit template with a one-sentence reason, or note "No issue templates found; using default body shape" when none exist.
```
lugassawan: **Medium (improvement, non-blocking):** The Python base64 decoder used for template body fetch and the step 7 API fallback lacks `.strip()` before decoding.

`gh api --jq '.content'` emits the raw base64 value with a trailing newline and embedded `\n` every 60 chars. Python's `b64decode` ignores non-base64 characters by default (`validate=False`), so this works today — but the pattern relies on that permissive default silently discarding whitespace.

**Suggested fix** — add `.strip()` before decoding (applies here and to the step 7 fallback at line 75):
```python
# template body fetch (step 4)
python3 -c "import base64,sys; print(base64.b64decode(sys.stdin.read().strip()).decode())"

# plugin version fallback (step 7)
python3 -c "import base64,sys,json; print(json.loads(base64.b64decode(sys.stdin.read().strip()))['version'])"
```
Not a blocker — current behaviour is correct — but `.strip()` makes the intent explicit and guards against any future GitHub API change that tightens whitespace handling.

tests/test_report_issue_command.py
```diff
@@ -0,0 +1,81 @@
...
+        "commands/report-issue.md must include a gh issue create call"
+    )
+    assert "gh issue list" in text, (
+        "commands/report-issue.md must include a gh issue list call for duplicate scan"
+    )
+    assert "gh label list" in text, (
+        "commands/report-issue.md must include a gh label list call for label discovery"
+    )
+
+
```
lugassawan, line 47: **Low:** The `count >= 4` floor on `--repo lugassawan/swe-workbench` is now too conservative — there are currently 8 occurrences, of which 2 are prose-only (the override declaration and a diagnostic instruction), not functional `gh` invocations.

If a future edit removes `--repo` from `gh issue list` or `gh label list`, the count drops to 6 and still clears the `>= 4` bar, silently masking the regression.

**Suggested fix** — either tighten the floor to `>= 6` to match current actual count, or use per-invocation substring assertions:
```python
for expected_cmd in [
    "gh repo view --repo",
    "gh label list --repo lugassawan/swe-workbench",
    "gh issue list --repo lugassawan/swe-workbench",
    "gh issue create --repo lugassawan/swe-workbench",
]:
    assert expected_cmd in text, f"Missing expected gh invocation: {expected_cmd}"
```
The per-invocation form is self-documenting and catches any single removal immediately.

lugassawan (commented):
## Review Summary

All five prior review threads were correctly addressed. The `product-manager.md` Mutation rule was tightened with a conditional clause (\"unless the calling command explicitly directs a specific target repo\") rather than left as an absolute prohibition, which makes the caller-side override in `report-issue.md` semantically sound. Base64 decoding was ported to Python for macOS/Linux portability, stale line-number references in the override block were replaced with rule-text references, and keyword sanitization in the dup-scan step was corrected. The 466/466 test suite passes and `validate.sh` exits clean.

**Review Decision: APPROVE**

Two minor non-blocking improvement opportunities noted via inline comments (do not block merge):
1. Medium — add `.strip()` before `base64.b64decode(sys.stdin.read())` in `commands/report-issue.md` Step 4 and Step 7 version capture to guard against trailing whitespace in `gh api` output.
2. Low — tighten `test_report_issue_passes_repo_flag_to_gh` assertion from `count >= 4` to `>= 6` or per-invocation assertions for stronger guarantees.

---
_Re-reviewed by `reviewer` (swe-workbench). All prior threads resolved; 2 new minor findings posted inline._

PR #254, opened 2026-05-16
[feat] Subagent skill-usage telemetry surfaced by the orchestrator

### LINKED ISSUE #227: [feat] Subagent skill-usage telemetry surfaced by the orchestrator
## Product framing

- **Problem.** When the orchestrator dispatches a subagent (via the Agent / Task tool), any skills the subagent invokes are invisible — the orchestrator only sees the subagent's final text response. There's no signal about which skills shaped the work, so it's impossible to spot when (a) a subagent forgot to invoke a relevant skill, (b) the same skill is being re-loaded across many subagents (cost signal), or (c) a skill is never being used in practice (dead code signal).
- **Value.** Developers iterating on the swe-workbench plugin get free, lightweight observability into skill usage across delegated work. Catches "subagent skipped the required skill" mistakes that are otherwise only visible by inspecting tool-call traces.
- **Acceptance criteria.**
  - Given an orchestrator dispatches a subagent that invokes one or more skills, when the subagent returns, the orchestrator prints a single line: `Skills used by <subagent-type>: <skill-1>, <skill-2>, ...`.
  - Given an orchestrator (top-level) invokes a skill directly, no skill-usage line is printed — the metric is for sub-invocations only.
  - Given a subagent runs and invokes zero skills, no line is printed (avoid noise).
  - The mechanism is opt-out at the subagent level if needed (some agents may legitimately want to stay silent), but on by default.
- **Impact / Effort (RICE-lite).** Impact: M — high-leverage observability for an otherwise opaque part of the system; pays off every time someone debugs subagent behavior. Effort: M — requires either a hook (PostToolUse on Task / SubagentStop) or a convention in each subagent's prompt that emits a structured trailer the orchestrator parses.

## Goal

After a dispatched subagent finishes, the orchestrator emits a one-line summary of which skills the subagent invoked, so skill usage in delegated work becomes visible.

## Motivation

Skills are central to how swe-workbench enforces process (TDD, debugging, brainstorming, etc.). The `superpowers:using-superpowers` skill explicitly tells every agent — including subagents — to invoke applicable skills proactively. Today, the orchestrator has no way to verify whether a subagent actually did so. Adding a low-cost telemetry line:
1. Lets the user see at a glance whether `swe-workbench:debugger` actually ran `systematic-debugging`, etc.
2. Surfaces missed-invocation patterns over time (e.g., "the test-writer subagent never invokes `test-driven-development`").
3. Is information-only — no behavior change, no enforcement.

## Surface

- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Skill — `workflow-*`
- [x] Subagent (all subagent prompts updated to emit the trailer)
- [ ] Slash command
- [x] Hook (PostToolUse on `Task` / `SubagentStop` to parse + print)
- [x] Documentation (catalog of the convention)
- [ ] Build / CI / release

## Sketch (optional)

Two implementation paths to evaluate (pick one in the implementation PR):

**Path A — convention + hook** (lighter touch):
- Each subagent prompt gets a closing instruction: "Before returning, emit a final line of the form `__skills_used__: <comma-separated-skill-names>` (omit the line if none).".
- A `SubagentStop` hook reads the subagent's final message, extracts the trailer if present, strips it from what the orchestrator sees, and prints `Skills used by <subagent-type>: ...` to the user.

**Path B — runtime instrumentation** (more invasive, more accurate):
- The plugin registers a `PreToolUse` hook on `Skill` that, when the calling context is a subagent (not the top-level orchestrator), records the skill name into a per-task buffer. On `SubagentStop`, the hook flushes the buffer.
- Requires the harness to expose enough context to distinguish "subagent invoked skill" from "orchestrator invoked skill" — verify this is available before committing to Path B.

Either way, the print is **suppressed for orchestrator-level skill invocations** — the metric exists to make *delegated* skill usage visible, not to narrate the top-level loop.

## Acceptance (optional)

- [ ] Dispatching any swe-workbench subagent that invokes at least one skill results in a single `Skills used by <subagent-type>: ...` line in the orchestrator's output
- [ ] Subagent runs with zero skill invocations produce no line
- [ ] Top-level (orchestrator) skill invocations produce no line
- [ ] The trailer (if using Path A) is stripped from the message the orchestrator otherwise sees, so it doesn't pollute downstream parsing
- [ ] Documentation in the README or `agents/shared/` explains the convention

lugassawan:
## Summary
- Add two runtime-instrumentation hooks (`skill_usage_record.sh` + `skill_usage_flush.sh`) that surface subagent skill invocations into the orchestrator transcript after each dispatched subagent finishes.
- Emit `Skills used by <agent-type>: skill-1, skill-2, ...` via `systemMessage`; top-level calls and zero-skill runs produce no output.
- Update `hooks.json` with `PreToolUse:Skill` and `SubagentStop` entries; update `scripts/validate.py` to allow lifecycle events without a `matcher` field.
- Add 15 subprocess-based pytest tests (TDD, red → green) covering no-ops, buffer write/flush, dedup, midnight straddle, opt-out, sweep, and path-traversal rejection.
- Add `docs/skill-usage-telemetry.md` opt-out guide; link from README.

## Test Plan
- [x] `python scripts/validate.py` — exit 0
- [x] `pytest tests/` — 475/475 pass (15 new telemetry tests)
- [x] Record hook: top-level no-op, subagent writes buffer, unknown agent no-op, opted-out no-op, missing skill no-op, 24h sweep
- [x] Flush hook: formatted systemMessage, buffer deleted, missing buffer `{}`, opted-out `{}`, midnight straddle
- [x] Path-traversal rejection in both hooks (agent_type sanitisation guard)
- [x] Existing rm -rf / Bash guard regression test still passes
- [ ] Live smoke test: dispatch a subagent that invokes a skill; verify "Skills used by …" line appears and cache dir is empty afterwards

- [x] New behaviour is information-only: both hooks exit 0 unconditionally; no tool call is blocked or modified
- [x] Opt-out mechanism confirmed by test: `skill_telemetry: false` in agent frontmatter suppresses telemetry in both hooks independently
- [x] No `agents/*.md` files modified (Path B — zero prompt changes)

Closes #227

### DIFF
--- a/docs/skill-usage-telemetry.md
+++ b/docs/skill-usage-telemetry.md
@@ -39,7 +39,7 @@ model: sonnet
 ---
 ```
 
-Both hooks check this flag independently, so a crash between `PreToolUse` and `SubagentStop` cannot cause a buffer to accumulate for an opted-out agent.
+Both hooks check this flag independently. In the common case — where the flag is present before the subagent starts — no buffer is ever written. If the flag is added mid-run, any in-flight buffer is cleaned up by the 24-hour sweep.
 
 ## Scope
 
--- a/docs/skill-usage-telemetry.md
+++ b/docs/skill-usage-telemetry.md
@@ -1,0 +1,58 @@
+# Skill-usage telemetry
+
+When the orchestrator dispatches a subagent via the `Agent` tool, any skills that subagent invokes are invisible to the orchestrator — it only receives the subagent's final text output.  The telemetry hooks surface those invocations by printing a single informational line into the orchestrator's transcript after each subagent finishes:
+
+```
+Skills used by <agent-type>: <skill-1>, <skill-2>, ...
+```
+
+## How it works
+
+Two hooks cooperate:
+
+| Hook | Script | Event |
+|---|---|---|
+| `PreToolUse:Skill` | `hooks/skill_usage_record.sh` | Fires on every `Skill` tool call.  When inside a subagent (`agent_id` present in stdin), appends the skill name to a date-stamped buffer: `.claude/cache/skill-usage/YYYYMMDD-<agent_id>.txt`. |
+| `SubagentStop` | `hooks/skill_usage_flush.sh` | Fires when a subagent finishes.  Reads the buffer, deduplicates (preserving first-seen order), formats the line, emits it via `systemMessage`, then deletes the buffer. |
+
+Top-level orchestrator skill calls produce **no line**. Zero-skill subagent runs produce **no line**.  The feature is information-only — it never blocks or modifies tool calls.
+
+## Where buffers live
+
+```
+<project-root>/.claude/cache/skill-usage/
+  20260517-abc-123.txt      ← one file per in-flight subagent
+  .errors.log               ← hook errors (only present when something went wrong)
+```
+
+Buffers older than 24 hours are swept opportunistically on each `PreToolUse:Skill` call.  `SubagentStop` deletes the buffer immediately after flushing.  Under normal operation the directory stays empty between agent runs.
+
+## Opting out
+
+An agent can suppress telemetry for itself by adding `skill_telemetry: false` to its YAML frontmatter (within the first 20 lines of the file):
+
+```markdown
+---
+name: my-agent
+skill_telemetry: false
+model: sonnet
+---
+```
+
+Both hooks check this flag independently, so a crash between `PreToolUse` and `SubagentStop` cannot cause a buffer to accumulate for an opted-out agent.
+
+## Scope
+
+The hooks only track agents whose `.md` file lives under `<plugin-root>/agents/`.  Other plugins ship their own equivalent hooks if they want telemetry for their subagents.
+
+## Troubleshooting
+
+If you expect a "Skills used by …" line but don't see one:
+
+1. Check `.claude/cache/skill-usage/.errors.log` for hook errors.
+2. Confirm the subagent's `.md` file exists under `agents/` and does **not** have `skill_telemetry: false`.
+3. Confirm `CLAUDE_PLUGIN_ROOT` resolves to the plugin root (hooks use this to locate agent files).
+
+## Security note
+
+Skill names land in the `systemMessage` field visible in the orchestrator's transcript.  If a future plugin author ever encodes sensitive data into skill names, those names would appear in the transcript.  Skill names in this plugin are all short, non-sensitive identifiers (e.g. `swe-workbench:principle-code-review`).
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -9,6 +9,25 @@
             "command": "cmd=$(jq -r '.tool_input.command // \"\"'); if echo \"$cmd\" | grep -Eq '(^|[[:space:]])rm[[:space:]]+-[a-zA-Z]*r[a-zA-Z]*f?[[:space:]]+(/|/\\*|~|\\$HOME)([[:space:]]|$)'; then echo 'BLOCKED: destructive rm against root or home' >&2; exit 2; fi; if echo \"$cmd\" | grep -Eq 'git[[:space:]]+push.*(--force([[:space:]]|$)|(^|[[:space:]])-f([[:space:]]|$))' && echo \"$cmd\" | grep -Eq '(^|[[:space:]]|:)(main|master)([[:space:]]|:|$)'; then echo 'BLOCKED: force push to main/master' >&2; exit 2; fi; if echo \"$cmd\" | grep -Eq 'git[[:space:]]+reset[[:space:]]+--hard'; then branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null); case \"$branch\" in main|master|release/*) echo \"BLOCKED: git reset --hard on protected branch '$branch'\" >&2; exit 2 ;; esac; fi; exit 0"
           }
         ]
+      },
+      {
+        "matcher": "Skill",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "$CLAUDE_PLUGIN_ROOT/hooks/skill_usage_record.sh"
+          }
+        ]
+      }
+    ],
+    "SubagentStop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "$CLAUDE_PLUGIN_ROOT/hooks/skill_usage_flush.sh"
+          }
+        ]
       }
     ]
   }
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -12,7 +12,8 @@ agent_id=$(printf '%s' "$input" | jq -r '.agent_id // empty')
 agent_type=$(printf '%s' "$input" | jq -r '.agent_type // empty')
 { [ -z "$agent_id" ] || [ -z "$agent_type" ]; } && { printf '{}'; exit 0; }
 
-# Sanitize agent_type: plain identifiers only — reject path-traversal attempts.
+# Sanitize agent_id and agent_type: plain identifiers only — reject path-traversal attempts.
+[[ "$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || { printf '{}'; exit 0; }
 [[ "$agent_type" =~ ^[A-Za-z0-9_-]+$ ]] || { printf '{}'; exit 0; }
 
 # Scope + opt-out (same gates as the record hook).
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+# SubagentStop handler — flushes the per-subagent buffer when a dispatched
+# subagent finishes.  Emits a single telemetry line via systemMessage so the
+# orchestrator's transcript shows which skills the subagent invoked.
+#
+# Exits 0 unconditionally; emits {} when there is nothing to report.
+set -u
+
+input=$(cat)
+
+agent_id=$(printf '%s' "$input" | jq -r '.agent_id // empty')
+agent_type=$(printf '%s' "$input" | jq -r '.agent_type // empty')
+{ [ -z "$agent_id" ] || [ -z "$agent_type" ]; } && { printf '{}'; exit 0; }
+
+# Sanitize agent_type: plain identifiers only — reject path-traversal attempts.
+[[ "$agent_type" =~ ^[A-Za-z0-9_-]+$ ]] || { printf '{}'; exit 0; }
+
+# Scope + opt-out (same gates as the record hook).
+plugin_root="${CLAUDE_PLUGIN_ROOT:-$(cd "$(dirname "$0")/.." && pwd)}"
+agent_file="$plugin_root/agents/$agent_type.md"
+[ -f "$agent_file" ] || { printf '{}'; exit 0; }
+if head -20 "$agent_file" | grep -Eq '^skill_telemetry:[[:space:]]*false[[:space:]]*$'; then
+  printf '{}'; exit 0
+fi
+
+cache_dir="${CLAUDE_PROJECT_DIR:-$PWD}/.claude/cache/skill-usage"
+
+# Buffer may span today's and yesterday's date-stamped file (straddle midnight).
+# Use a glob loop so paths with spaces are handled correctly (compatible with bash 3.2+).
+buffers=()
+for f in "$cache_dir"/*-"$agent_id".txt; do
+  [ -f "$f" ] && buffers+=("$f")
+done
+[ "${#buffers[@]}" -eq 0 ] && { printf '{}'; exit 0; }
+
+# Dedupe preserving first-seen order; join with ", ".
+skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++' | paste -sd, - | sed 's/,/, /g')
+[ -z "$skills" ] && { printf '{}'; exit 0; }
+
+# Clean up all matching buffers regardless of whether the emit succeeds below.
+rm -f "${buffers[@]}" 2>/dev/null || true
+
+# Safely JSON-encode the message.
+msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .)
+printf '{"systemMessage": %s, "suppressOutput": true}\n' "$msg"
+exit 0
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -33,8 +33,8 @@ for f in "$cache_dir"/*-"$agent_id".txt; do
 done
 [ "${#buffers[@]}" -eq 0 ] && { printf '{}'; exit 0; }
 
-# Dedupe preserving first-seen order; join with ", ".
-skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++' | paste -sd, - | sed 's/,/, /g')
+# Dedupe preserving first-seen order; join with ", " in one pass (POSIX awk, no paste/sed).
+skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++ { out = (out ? out ", " : "") $0 } END { print out }')
 [ -z "$skills" ] && { printf '{}'; exit 0; }
 
 # Clean up all matching buffers regardless of whether the emit succeeds below.
--- a/hooks/skill_usage_record.sh
+++ b/hooks/skill_usage_record.sh
@@ -11,9 +11,15 @@ input=$(cat)
 agent_id=$(printf '%s' "$input" | jq -r '.agent_id // empty')
 [ -z "$agent_id" ] && exit 0          # top-level orchestrator → no-op
 
+# Sanitize agent_id: plain identifiers only — same rule as agent_type.
+[[ "$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || exit 0
+
 agent_type=$(printf '%s' "$input" | jq -r '.agent_type // empty')
 skill=$(printf '%s' "$input" | jq -r '.tool_input.skill // empty')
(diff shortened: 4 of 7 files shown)


hooks/skill_usage_record.sh
```diff
@@ -0,0 +1,42 @@
...
+[ -f "$agent_file" ] || exit 0
+
+# Opt-out: frontmatter `skill_telemetry: false` in the first 20 lines.
+if head -20 "$agent_file" | grep -Eq '^skill_telemetry:[[:space:]]*false[[:space:]]*$'; then
+  exit 0
+fi
+
+cache_dir="${CLAUDE_PROJECT_DIR:-$PWD}/.claude/cache/skill-usage"
+mkdir -p "$cache_dir" 2>/dev/null || exit 0
+buffer="$cache_dir/$(date +%Y%m%d)-$agent_id.txt"
```
lugassawan, line 39: **High: `agent_id` not sanitized before use in buffer file path**

`agent_type` is validated with `^[A-Za-z0-9_-]+$` on line 19, but `agent_id` on this line is embedded directly into the file path with no equivalent check. A crafted value like `../../somewhere/evil` causes the buffer file write to land outside the cache directory — the `mkdir -p` on line 32 would silently create the traversal path.

While `agent_id` is supplied by the Claude Code runtime and unlikely to be attacker-controlled today, the asymmetric treatment creates a latent inconsistency.

**Suggested fix** — add after line 12:
```bash
[[ "$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || exit 0
```

hooks/skill_usage_record.sh
```diff
@@ -0,0 +1,42 @@
...
+# which is unacceptable for an observation-only hook. Exit 0 always.
+set -u
+
+input=$(cat)
+
+agent_id=$(printf '%s' "$input" | jq -r '.agent_id // empty')
+[ -z "$agent_id" ] && exit 0          # top-level orchestrator → no-op
+
+agent_type=$(printf '%s' "$input" | jq -r '.agent_type // empty')
+skill=$(printf '%s' "$input" | jq -r '.tool_input.skill // empty')
```
lugassawan, line 18: **Medium: unsanitized `skill` field written verbatim to buffer**

`jq -r` interprets JSON string escapes. A skill name of `"foo\nbar"` in the raw JSON input becomes two lines in the buffer file. At flush time `awk '!seen[$0]++'` treats each line as a distinct skill, producing `Skills used by agent: foo, bar` instead of one entry. `jq -Rs .` in the flush hook safely encodes the final JSON payload, so there is no injection risk — but the output is semantically wrong.

Real Claude Code skill names never contain newlines, but without enforcement this is a silent correctness trap.

**Suggested fix** — add after line 15:
```bash
# Reject skill names that contain newlines (jq -r interprets \n escapes in JSON).
[[ "$skill" =~ ^[A-Za-z0-9_:/-]+$ ]] || exit 0
```

tests/test_skill_usage_telemetry.py
```diff
@@ -0,0 +1,350 @@
...
+                "agent_type": "reviewer",
+                "tool_input": {"skill": "swe-workbench:principle-code-review"},
+            },
+            optout_plugin_root,
+            cache_dir,
+        )
+        assert result.returncode == 0
+        assert _buffer_files(cache_dir) == []
+
+    def test_path_traversal_agent_type_is_noop(self, plugin_root, cache_dir):
```
lugassawan, line 158: **Medium: test suite has no end-to-end record→flush pipeline test; no `agent_id` path-traversal test**

All 15 tests drive `record.sh` or `flush.sh` in isolation, or pre-populate the buffer manually. The acceptance criterion ("given a subagent invokes skills, orchestrator prints the line") maps to a sequential record→flush call that is never exercised in the test suite.

Additionally, `test_path_traversal_agent_type_is_noop` (line 158) tests `agent_type` traversal but there is no parallel test for a malicious `agent_id` value like `../../evil`.

**Suggested additions:**
1. An integration test that calls `_run_record` then `_run_flush` with matching `agent_id`/`agent_type` and asserts the flush output contains the skill name.
2. A `test_path_traversal_agent_id_is_noop` test in `TestRecordHook` mirroring the existing `agent_type` variant.

hooks/skill_usage_flush.sh
```diff
@@ -0,0 +1,46 @@
...
+# Buffer may span today's and yesterday's date-stamped file (straddle midnight).
+# Use a glob loop so paths with spaces are handled correctly (compatible with bash 3.2+).
+buffers=()
+for f in "$cache_dir"/*-"$agent_id".txt; do
+  [ -f "$f" ] && buffers+=("$f")
+done
+[ "${#buffers[@]}" -eq 0 ] && { printf '{}'; exit 0; }
+
+# Dedupe preserving first-seen order; join with ", ".
+skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++' | paste -sd, - | sed 's/,/, /g')
```
lugassawan: **Low: `paste -sd, -` trailing hyphen is a GNU extension on some platforms**

The trailing `-` (read from stdin) is widely accepted on GNU coreutils and macOS BSD, but is not guaranteed by POSIX. Given the bash 3.2 compatibility care taken elsewhere in this PR (the `mapfile` → glob-for-loop fix), it is worth replacing the `paste | sed` pipeline with a single pure-`awk` pass that deduplicates and joins in one step:

```bash
skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++ { if (out) out = out ", " $0; else out = $0 } END { print out }')
```

This removes the `paste` and `sed` dependencies and is fully POSIX-compatible.

hooks/hooks.json
```diff
@@ -9,6 +9,25 @@
...
+        "matcher": "Skill",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "$CLAUDE_PLUGIN_ROOT/hooks/skill_usage_record.sh"
+          }
+        ]
+      }
+    ],
+    "SubagentStop": [
```
lugassawan, line 23: **Low: the absence of a `matcher` field on the `SubagentStop` entry is correct but undocumented**

The validator exemption for lifecycle events is in `_LIFECYCLE_HOOK_EVENTS` in `scripts/validate.py`. However, the frozenset also silently permits future entries under `Stop`, `Notification`, and `PreCompact` without a matcher — none of those are validated. A future contributor adding a tool-use event (e.g. `PostToolUse`) to that frozenset by mistake would bypass the matcher check with no warning.

Consider adding a comment to `scripts/validate.py:13` clarifying the distinction:
```python
# Hook events that fire unconditionally (no tool name to match against).
# Do NOT add PreToolUse / PostToolUse here — those ARE tool-matchers and must have a matcher field.
_LIFECYCLE_HOOK_EVENTS = frozenset({"SubagentStop", "PreCompact", "Stop", "Notification"})
```

docs/skill-usage-telemetry.md
```diff
@@ -0,0 +1,58 @@
...
+
+```markdown
+---
+name: my-agent
+skill_telemetry: false
+model: sonnet
+---
+```
+
+Both hooks check this flag independently, so a crash between `PreToolUse` and `SubagentStop` cannot cause a buffer to accumulate for an opted-out agent.
```
lugassawan: **Low: doc claim about crash-safety overstates the guarantee**

The current text: "Both hooks check this flag, so a crash between `PreToolUse` and `SubagentStop` cannot cause a buffer to accumulate for an opted-out agent."

This is accurate when `skill_telemetry: false` is already set at the time `PreToolUse` fires. But if the flag is added to the agent file *after* a recording session has started (e.g. during a long-running subagent), the record hook has already written a buffer that the flush hook will then skip — leaving the buffer to be swept by the 24h cleanup rather than immediately removed.

**Suggested wording:**
> "Both hooks check this flag independently. In the common case — where the flag is present before the subagent starts — no buffer is ever written. If the flag is added mid-run, any in-flight buffer is cleaned up by the 24-hour sweep."

lugassawan (commented):
## Review Summary

This PR ships a clean, well-scoped telemetry feature: two shell hooks (record on `PreToolUse:Skill`, flush on `SubagentStop`), a validator fix, 15 subprocess-based tests, and clear docs. The design is sound — double-quoting prevents glob expansion on untrusted values, `jq -Rs` correctly encodes the final JSON payload, and the sweep/flush lifecycle is safe for the normal code path. The strongest positives are the bash 3.2 compatibility care (the `mapfile` → glob-for-loop fix caught by subprocess tests), the opportunistic sweep to avoid buffer accumulation, and the independent opt-out check in both hooks.

The most important concern is a **missing `agent_id` sanitization in `skill_usage_record.sh`**: `agent_type` is regex-validated in both hooks, but `agent_id` is embedded directly into the buffer file path with no equivalent guard. A crafted value containing `../` can write outside the intended cache directory. There is also a medium-severity correctness issue where a skill name containing a JSON `\n` escape passes through `jq -r` as a real newline, splitting one skill into two apparent entries at flush time.

**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 6 inline comments, deduped 0._

hooks/skill_usage_record.sh
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: added `[[ "\$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || exit 0` after the empty-check on line 12, matching the existing `agent_type` guard.

hooks/skill_usage_record.sh
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: added `[[ "\$skill" =~ ^[A-Za-z0-9_:/-]+$ ]] || exit 0` after skill extraction — rejects values that jq -r would expand to newlines or shell metacharacters.

tests/test_skill_usage_telemetry.py
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: added `test_path_traversal_agent_id_is_noop` for the agent_id guard, and a new `TestRecordFlushIntegration` class with two end-to-end tests (single skill + dedup). Test count: 15 → 18.

hooks/skill_usage_flush.sh
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: replaced `paste -sd, - | sed 's/,/, /g'` with a single-pass `awk '!seen[$0]++ { out = (out ? out ", " : "") $0 } END { print out }'` — no paste or sed dependency, POSIX-compatible.

hooks/hooks.json
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: expanded the comment above `_LIFECYCLE_HOOK_EVENTS` to explicitly warn against adding PreToolUse/PostToolUse tool-matcher events to the frozenset.

docs/skill-usage-telemetry.md
lugassawan: Addressed in a1f1dd26122580b84b4ee30c6fcd9e91b4e3d05c: softened the claim — 'In the common case — where the flag is present before the subagent starts — no buffer is ever written. If the flag is added mid-run, any in-flight buffer is cleaned up by the 24-hour sweep.'

hooks/skill_usage_flush.sh
```diff
@@ -0,0 +1,46 @@
...
+# orchestrator's transcript shows which skills the subagent invoked.
+#
+# Exits 0 unconditionally; emits {} when there is nothing to report.
+set -u
+
+input=$(cat)
+
+agent_id=$(printf '%s' "$input" | jq -r '.agent_id // empty')
+agent_type=$(printf '%s' "$input" | jq -r '.agent_type // empty')
+{ [ -z "$agent_id" ] || [ -z "$agent_type" ]; } && { printf '{}'; exit 0; }
```
lugassawan, line 13: **Medium: `flush.sh` still has no `agent_id` sanitization guard (asymmetric with `record.sh`)**

The fix commit added `[[ "$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || exit 0` to `record.sh:15`, but `flush.sh` was not updated symmetrically. `agent_id` is interpolated into the glob pattern on line 31 (`"$cache_dir"/*-"$agent_id".txt`). Double-quoting prevents active exploitation today, but the asymmetry means a future refactor could reintroduce the vulnerability without a guard already in place.

**Suggested fix** — add after line 13:
```bash
# Sanitize agent_id: plain identifiers only — same rule as agent_type.
[[ "$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || { printf '{}'; exit 0; }
```

tests/test_skill_usage_telemetry.py
```diff
@@ -0,0 +1,428 @@
...
+        result = _run_flush(
+            {"agent_id": "flush-003", "agent_type": "reviewer"},
+            optout_plugin_root,
+            cache_dir,
+        )
+        assert result.returncode == 0
+        assert json.loads(result.stdout) == {}
+        assert buf.exists(), "Buffer should remain when agent is opted out"
+
+    def test_path_traversal_agent_type_is_noop(self, plugin_root, cache_dir):
```
lugassawan, line 300: **Medium: `TestFlushHook` has no `agent_id` path-traversal test**

`TestRecordHook` gained `test_path_traversal_agent_id_is_noop` in the fix commit, but `TestFlushHook` only has `test_path_traversal_agent_type_is_noop`. If the sanitization guard above is added to `flush.sh`, it will need a corresponding test — and the gap exists either way.

**Suggested addition to `TestFlushHook`:**
```python
def test_path_traversal_agent_id_is_noop(self, plugin_root, cache_dir):
    """Flush rejects agent_id values with path-traversal characters."""
    for malicious in ["../../evil", "../sensitive", "foo/bar", "foo bar"]:
        result = _run_flush(
            {"agent_id": malicious, "agent_type": "reviewer"},
            plugin_root,
            cache_dir,
        )
        assert result.returncode == 0
        assert json.loads(result.stdout) == {}
```

lugassawan (commented):
## Review Summary

All 6 findings from the first review were correctly addressed: the `agent_id` and `skill` sanitization guards land in `record.sh`, the awk replacement in `flush.sh` is semantically equivalent to the old `paste | sed` pipeline (including the empty-input and single-item edge cases), the integration tests exercise the full record→flush path, the `_LIFECYCLE_HOOK_EVENTS` comment is clear, and the doc claim is accurately softened.

One structural gap remains: `flush.sh` did not receive the `agent_id` sanitization guard that `record.sh` did. The double-quoted glob pattern prevents active exploitation today, but the asymmetry is inconsistent with the defense-in-depth posture the rest of the PR establishes, and `TestFlushHook` has no `agent_id` traversal counterpart to match the new `TestRecordHook` test.

**Review Decision: COMMENT**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 2 inline comments, deduped 0._

hooks/skill_usage_flush.sh
lugassawan: Addressed in baa0be2: added `[[ "\$agent_id" =~ ^[A-Za-z0-9_-]+$ ]] || { printf '{}'; exit 0; }` immediately after the empty-check on line 13, symmetric with the `agent_type` guard.

tests/test_skill_usage_telemetry.py
lugassawan: Addressed in baa0be2: added `test_path_traversal_agent_id_is_noop` to `TestFlushHook` mirroring the `TestRecordHook` variant. Test count: 18 → 19.

tests/test_skill_usage_telemetry.py
```diff
@@ -0,0 +1,439 @@
...
+        """Flush rejects agent_type values with path-traversal characters."""
+        for malicious in ["../../etc/passwd", "../sensitive", "foo/bar"]:
```
lugassawan: **Medium** | `tests/test_skill_usage_telemetry.py:302` | Missing `"foo bar"` from `TestFlushHook.test_path_traversal_agent_type_is_noop` | The analogous `TestRecordHook.test_path_traversal_agent_type_is_noop` (line 160) tests four inputs: `["../../etc/passwd", "../sensitive", "foo/bar", "foo bar"]`. The flush variant at line 302 tests only three — `"foo bar"` is absent. The guard in `flush.sh` correctly rejects space-containing strings, but the omission leaves the test matrix asymmetric. If the regex is ever relaxed, the space-injection vector would go undetected. **Suggested fix:** add `"foo bar"` to the list on line 302: `for malicious in ["../../etc/passwd", "../sensitive", "foo/bar", "foo bar"]:`

lugassawan (commented):
## Review Summary

This is a well-executed follow-up that completes the previously flagged security gap symmetrically: `flush.sh` now has the same `agent_id` sanitization guard as `record.sh`, the guard ordering is correct (sanitize before any path construction), and `set -u` is safe throughout. The 19-test suite is comprehensive, the awk dedup pipeline is correct, and `jq -Rs .` encoding handles all valid skill name characters. One minor test-matrix asymmetry remains (see inline comment). The implementation is solid and ready to merge.

**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comments, deduped 0._

tests/test_skill_usage_telemetry.py
lugassawan: Addressed in 1529b0ac5c20d0f2eaad840d1915a1d23236dc48: added "foo bar" to the traversal input list.

PR #263, opened 2026-05-17
[chore] Add post-merge cleanup hint + clarify pr-review stale-SHA recovery

lugassawan:
## Summary

- **E7 — `commands/implement.md`**: Appended a post-merge cleanup hint to Phase 5. After `/implement` opens a PR, new users had no signal that `/swe-workbench:cleanup-merged <N>` is the natural follow-up once the PR merges, leaving stale worktrees and branches.
- **O8 — `skills/workflow-pr-review/SKILL.md`**: Reworded the `All POSTs returned 422` failure-modes table row to make stale-recovery intent explicit. The original text said "Re-fetch HEAD_SHA ... and retry" without explaining why — a future reader could mistake the re-fetch for redundancy. The new wording leads with the diagnosis: cached `$HEAD_SHA` is *provably* stale when 422 fires on every POST (the PR head advanced between Step 1 and Step 6). **Note:** AC #2 of the original ticket ("error-recovery block reuses the captured variable") was amended after `senior-engineer` review — the re-fetch at line 257 is intentional stale-recovery, not redundant; removing it would defeat the recovery path. The fix is clarification, not deletion.

## Test Plan

- [x] Changed skills/commands/agents load without errors
- [x] Examples in the diff were exercised manually
- [x] Diff is exactly two prose-only hunks; `gh pr view` count in SKILL.md unchanged (3 → 3); `cleanup-merged` reference appears exactly once in `implement.md`; table pipe characters verified intact; SKILL.md frontmatter confirmed intact

### Type-tailored checks ([chore])
- [x] Both edited files build/load correctly in the plugin (no syntax errors in Markdown)
- [x] No behavioural change introduced — diff is prose-only rewording

Closes #237

PR #264, opened 2026-05-17
[fix] workflow-pr-review: emit address-feedback CTA on self-review too

lugassawan:
## Summary
- The address-feedback CTA (`"Want me to help address this feedback? Reply yes to start /address-feedback N"`) was silently suppressed on self-review (`CURRENT_USER == AUTHOR_LOGIN`), breaking the `/implement → /review → /address-feedback` chain observed during PR #263.
- Root cause: the CTA gate incorrectly AND-combined two orthogonal policy axes — identity (who authored the PR) and outcome (whether findings are actionable). Only the outcome axis should gate the CTA.
- Fix: remove the identity-axis bullet from the CTA suppression block in both `workflow-pr-review/SKILL.md` and `workflow-pr-review-followup/SKILL.md`. The clean-approval suppression (`DECISION=APPROVE AND posted=0 AND deduped=0`) is preserved.
- Also fixes a stale "same policy axis" cross-reference in the Common Mistakes table that would have caused future LLM readers to re-introduce the identity suppression.
- Adds 8 pytest assertions pinning the CTA contract for both skill files.

## Test Plan
- [x] `pytest tests/test_workflow_pr_review_cta_suppression.py -v` — 8/8 pass
- [x] `pytest tests/ -k workflow_pr_review` — 14/14 pass (no regressions)
- [x] `grep -n "CURRENT_USER == AUTHOR_LOGIN" skills/workflow-pr-review/SKILL.md skills/workflow-pr-review-followup/SKILL.md` — 0 matches
- [x] `grep -n "CURRENT_USER != AUTHOR_LOGIN" skills/workflow-pr-review/SKILL.md skills/workflow-pr-review-followup/SKILL.md` — 0 matches

- [x] Reproducing scenario confirmed: self-review with posted findings previously dropped the CTA
- [x] Non-regression: cross-author review CTA still fires when findings are actionable
- [x] Non-regression: clean-approval CTA suppression (`APPROVE + posted=0 + deduped=0`) still in place
- [x] Deliberately untouched: `IS_SELF_REVIEW` gate on GitHub review-event submission (lines 200-205) — GitHub blocks self-approval, correct to preserve
- [x] Deliberately untouched: `IS_SELF_REVIEW` gate on `## Review Summary` narrative — separate axis, out of scope

Issue: N/A — bug observed during PR #263 self-review workflow

### DIFF
--- a/tests/test_workflow_pr_review_cta_suppression.py
+++ b/tests/test_workflow_pr_review_cta_suppression.py
@@ -17,13 +17,16 @@ PR_REVIEW_FOLLOWUP_SKILL = ROOT / "skills" / "workflow-pr-review-followup" / "SK
 
 def _suppression_block(text: str) -> str:
     """Extract the paragraph(s) immediately following the CTA quote block."""
-    # Grab text from the CTA header through to the next ## section or end
     match = re.search(
         r"Address-feedback CTA \(conditional\):.*?(?=\n## |\Z)",
         text,
         re.DOTALL,
     )
-    return match.group(0) if match else ""
+    assert match is not None, (
+        "Could not locate 'Address-feedback CTA (conditional):' section. "
+        "The CTA section header was renamed or removed — update the regex in _suppression_block."
+    )
+    return match.group(0)
 
 
 # ── workflow-pr-review/SKILL.md ───────────────────────────────────────────────
--- a/tests/test_workflow_pr_review_cta_suppression.py
+++ b/tests/test_workflow_pr_review_cta_suppression.py
@@ -1,0 +1,118 @@
+# tests/test_workflow_pr_review_cta_suppression.py
+
+"""Pin the address-feedback CTA suppression contract for both pr-review skills.
+
+Bug: IS_SELF_REVIEW caused the CTA to be silently dropped when the user ran
+/implement → /review on their own PR, even when findings were posted. The fix
+removes the identity axis from the CTA gate — only the outcome axis gates it.
+"""
+
+import re
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+PR_REVIEW_SKILL = ROOT / "skills" / "workflow-pr-review" / "SKILL.md"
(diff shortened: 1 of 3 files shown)


tests/test_workflow_pr_review_cta_suppression.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _suppression_block(text: str) -> str:
+    """Extract the paragraph(s) immediately following the CTA quote block."""
+    # Grab text from the CTA header through to the next ## section or end
+    match = re.search(
+        r"Address-feedback CTA \(conditional\):.*?(?=\n## |\Z)",
+        text,
+        re.DOTALL,
+    )
```
lugassawan, line 24: **Medium — `_suppression_block` returns `""` silently on a header rename; negative assertions pass vacuously**

If a future edit changes `"Address-feedback CTA (conditional):"` to any other phrasing, the regex returns `""`. The two negative tests (`no_identity_suppression`, `no_self_review_suppression`) would then pass — `"CURRENT_USER == AUTHOR_LOGIN" not in ""` is trivially `True` — while the identity gate has actually been reintroduced under a new header.

Suggested fix — assert non-empty at the top of the helper:

```python
def _suppression_block(text: str) -> str:
    match = re.search(
        r"Address-feedback CTA \(conditional\):.*?(?=\n## |\Z)",
        text,
        re.DOTALL,
    )
    assert match is not None, (
        "Could not locate 'Address-feedback CTA (conditional):' section. "
        "The CTA section header was renamed or removed — update the regex."
    )
    return match.group(0)
```

Applies symmetrically to both skill files (the same helper is shared).

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

lugassawan: Addressed in 1383bce17f78cc21f6a8b9e64dba7368868ce345: added `assert match is not None` guard to `_suppression_block` so a header rename produces an explicit failure instead of a vacuous pass.

PR #265, opened 2026-05-17
[fix] Hook: close $HOME bypass and short-circuit non-rm/non-git

### LINKED ISSUE #233: [bug] PreToolUse Bash hook: $HOME guard bypass and per-call pipeline overhead
## Product framing

**Problem.** Users who type `rm -rf $HOME/foo` get false safety — the shell expands `$HOME` to an absolute path before Claude Code captures `tool_input.command`, so the guard never sees `$HOME` and never fires. Meanwhile, every Bash call (even `ls .`) pays a full `jq` + 3× `grep` + conditional `git rev-parse` pipeline even when no risky pattern could possibly match.

**Value.** Closing the bypass makes the safety guarantee real, not cosmetic. Short-circuiting the pipeline removes latency tax from the majority of Bash calls that need no guarding. Both fixes land in one JSON patch.

**Acceptance criteria.**
- `rm -rf /Users/foo` (shell-expanded `$HOME`) is blocked
- `rm -rf ~` and `rm -rf $HOME` (literal) remain blocked
- `rm -rf /` remains blocked
- Non-`rm`/non-`git` Bash commands skip all grep invocations entirely
- `git reset --hard` on `main`/`master`/`release/*` remains blocked

**Impact / Effort.** Impact: H — a safety guard that silently no-ops is worse than no guard (false confidence). Effort: S — single JSON string patch in `hooks/hooks.json`.

---

## Surface affected

- [x] Hook

**Specific surface:** `hooks/hooks.json` — PreToolUse Bash hook (line 9)

## Reproduction steps

**B1 — `$HOME` guard bypass:**
1. Start a Claude Code session in this repo.
2. Ask Claude to run: `rm -rf $HOME/Documents`
3. The hook **does not block it** — `$HOME` is expanded to `/Users/<you>` by the shell before `tool_input.command` is captured; the regex only tests for the literal string `$HOME` or `~`.

**O1 — Per-call pipeline overhead:**
1. Ask Claude to run any simple read-only command, e.g. `ls .`
2. The hook fires a full `jq` parse + 3× `grep -E` + conditional `git rev-parse` before the command runs — even though `ls` can never match any guard pattern.

## Expected behavior

- `rm -rf /Users/<any-user>` is blocked (catches shell-expanded `$HOME`).
- Bash commands that do not start with `rm` or `git` skip the pipeline with zero `grep` invocations.

## Actual behavior

- `rm -rf /Users/foo` passes through unguarded.
- Every Bash call pays the full pipeline cost regardless of content.

## Suggested fix sketch

```bash
# 1. Short-circuit first — most commands don't need guarding
cmd=$(jq -r '.tool_input.command // ""')
case "$cmd" in
  *rm\ *|*git\ *) ;;   # fall through to safety checks
  *)              exit 0 ;;
esac

# 2. Expand guard to catch /Users/* (macOS) and /home/* (Linux) paths
# in addition to ~ and $HOME literals
if echo "$cmd" | grep -Eq \
  '(^|[[:space:]])rm[[:space:]]+-[a-zA-Z]*r[a-zA-Z]*f?[[:space:]]+(/(\*|[[:space:]]|$)|~|\$HOME|/Users/[^/[:space:]]+(/[^/[:space:]]+)?[[:space:]]|/home/[^/[:space:]]+(/[^/[:space:]]+)?[[:space:]])'; then
  echo 'BLOCKED: destructive rm against root or home' >&2; exit 2
fi
```

lugassawan:
## Summary

- Extract the PreToolUse:Bash inline one-liner to `hooks/bash_guard.sh`, matching the `skill_usage_record.sh` pattern — shellcheck auto-discovers it, tests invoke it end-to-end.
- Fix `$HOME` guard bypass: shell-expanded paths like `/Users/foo/Documents` were not matched by the old regex `(/|/\*|~|\$HOME)`. New regex covers `/Users/<user>/…`, `/home/<user>/…`, `~`, literal `$HOME`, bare `/Users` and `/home`, quoted paths (via `tr -d` quote stripping), and `-Rf`/`-RF` (BSD/macOS uppercase-R synonym).
- Add `case`-glob short-circuit: commands containing neither `rm` nor `git` exit 0 after one `jq` call — eliminates the per-call 3× grep overhead for `ls`, `echo`, `make`, etc.

## Test Plan

- [x] `bash scripts/validate.sh` → "All checks passed"

- [ ] `shellcheck hooks/bash_guard.sh` → clean
- [ ] `python3 -m pytest tests/test_hooks.py -v` → 56/56 pass (all new e2e cases)
- [ ] `python3 -m pytest tests/ -q` → 529/529 pass
- [ ] Manual: `echo '{"tool_input":{"command":"rm -rf /Users/foo"}}' | hooks/bash_guard.sh; echo "exit=$?"` → `BLOCKED` + `exit=2`
- [ ] Manual: `echo '{"tool_input":{"command":"ls ."}}' | hooks/bash_guard.sh; echo "exit=$?"` → `exit=0` (no grep invoked)
- [ ] Manual: `echo '{"tool_input":{"command":"rm -rf \"/Users/foo\""}}' | hooks/bash_guard.sh; echo "exit=$?"` → `BLOCKED` + `exit=2` (quote-strip verified)
- [ ] Verified: `git ls-files --stage hooks/bash_guard.sh` → mode `100755`

Closes #233

### DIFF
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -2,7 +2,7 @@
 # PreToolUse:Bash guard — block destructive commands and short-circuit safe ones.
 #
 # Blocks:
-#   - rm -rf against /, ~, $HOME (literal or shell-expanded /Users/*|/home/*)
+#   - rm -rf against /, /*, ~, $HOME, /Users[/<user>], /home[/<user>]
 #   - git push --force / -f to main/master
 #   - git reset --hard on main/master/release/*
 #
@@ -10,15 +10,18 @@
 # "git". This is the common case (ls, cat, echo, make, npm, …) and removes the
 # per-call grep tax flagged in #233.
 #
-# Out of scope: ${HOME} brace-form; path normalization via .. traversal.
+# Out of scope: ${HOME} brace-form; ANSI-C $'...' quoting; path normalization via .. traversal.
 
 set -u
 
-cmd=$(jq -r '.tool_input.command // ""')
+if ! cmd=$(jq -r '.tool_input.command // ""' 2>/dev/null); then
+  echo 'bash_guard: jq parse error — blocking by default' >&2
+  exit 2
+fi
 
 case "$cmd" in
-  *rm*|*git*) ;;
-  *)          exit 0 ;;
+  rm\ *|*\ rm\ *|*git*) ;;
+  *)                    exit 0 ;;
 esac
 
 # Strip quotes so quoted paths cannot bypass the rm regex.
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -20,12 +20,12 @@ if ! cmd=$(jq -r '.tool_input.command // ""'); then
 fi
 
 case "$cmd" in
-  rm\ *|*\ rm\ *|*\(rm\ *|*git*) ;;
+  rm\ *|*\ rm\ *|*\(*rm\ *|*git*) ;;
   *)                              exit 0 ;;
 esac
 
-# Strip quotes so quoted paths cannot bypass the rm regex.
-norm=$(printf '%s' "$cmd" | tr -d "'\"")
+# Strip quotes and subshell parens so they cannot mask the rm command or path.
+norm=$(printf '%s' "$cmd" | tr -d "'\"()")
 
 # shellcheck disable=SC2016  # $HOME in single quotes is intentional: matches literal text, not the shell variable
 # [rR] covers both -rf and -Rf (BSD/macOS rm accepts -R as synonym for -r).
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+# PreToolUse:Bash guard — block destructive commands and short-circuit safe ones.
+#
+# Blocks:
+#   - rm -rf against /, ~, $HOME (literal or shell-expanded /Users/*|/home/*)
+#   - git push --force / -f to main/master
+#   - git reset --hard on main/master/release/*
+#
+# Short-circuits (exit 0, no greps) for commands that contain neither "rm" nor
+# "git". This is the common case (ls, cat, echo, make, npm, …) and removes the
+# per-call grep tax flagged in #233.
+#
+# Out of scope: ${HOME} brace-form; path normalization via .. traversal.
+
+set -u
+
+cmd=$(jq -r '.tool_input.command // ""')
+
+case "$cmd" in
+  *rm*|*git*) ;;
+  *)          exit 0 ;;
+esac
+
+# Strip quotes so quoted paths cannot bypass the rm regex.
+norm=$(printf '%s' "$cmd" | tr -d "'\"")
+
+# shellcheck disable=SC2016  # $HOME in single quotes is intentional: matches literal text, not the shell variable
+# [rR] covers both -rf and -Rf (BSD/macOS rm accepts -R as synonym for -r).
+if echo "$norm" | grep -Eq \
+   '(^|[[:space:]])rm[[:space:]]+-[a-zA-Z]*[rR][a-zA-Z]*[fF]?[[:space:]]+(/(\*|[[:space:]]|$)|(~|\$HOME)(/[^[:space:]]*)?([[:space:]]|$)|(/Users|/home)(/[^/[:space:]]+)?([[:space:]]|/|$))'; then
+  echo 'BLOCKED: destructive rm against root or home' >&2
+  exit 2
+fi
+
+if echo "$norm" | grep -Eq 'git[[:space:]]+push.*(--force([[:space:]]|$)|(^|[[:space:]])-f([[:space:]]|$))' \
+   && echo "$norm" | grep -Eq '(^|[[:space:]]|:)(main|master)([[:space:]]|:|$)'; then
+  echo 'BLOCKED: force push to main/master' >&2
+  exit 2
+fi
+
+if echo "$norm" | grep -Eq 'git[[:space:]]+reset[[:space:]]+--hard'; then
+  branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || true)
+  case "$branch" in
+    main|master|release/*)
+      echo "BLOCKED: git reset --hard on protected branch '$branch'" >&2
+      exit 2
+      ;;
+  esac
+fi
+
+exit 0
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -14,14 +14,14 @@
 
 set -u
 
-if ! cmd=$(jq -r '.tool_input.command // ""' 2>/dev/null); then
+if ! cmd=$(jq -r '.tool_input.command // ""'); then
   echo 'bash_guard: jq parse error — blocking by default' >&2
   exit 2
 fi
 
 case "$cmd" in
-  rm\ *|*\ rm\ *|*git*) ;;
-  *)                    exit 0 ;;
+  rm\ *|*\ rm\ *|*\(rm\ *|*git*) ;;
+  *)                              exit 0 ;;
 esac
 
 # Strip quotes so quoted paths cannot bypass the rm regex.
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -10,7 +10,7 @@
 # "git". This is the common case (ls, cat, echo, make, npm, …) and removes the
 # per-call grep tax flagged in #233.
 #
-# Out of scope: ${HOME} brace-form; ANSI-C $'...' quoting; path normalization via .. traversal.
+# Out of scope: ${HOME} brace-form; ANSI-C $'...' quoting; path normalization via .. traversal; compound cd (cwd not in hook payload).
 
 set -u
 
@@ -19,13 +19,16 @@ if ! cmd=$(jq -r '.tool_input.command // ""'); then
   exit 2
 fi
 
-case "$cmd" in
+# Normalise shell separators to spaces so rm/git after ; | & are still detected.
+_norm=$(printf '%s' "$cmd" | tr ';|&' '   ')
+
+case "$_norm" in
   rm\ *|*\ rm\ *|*\(*rm\ *|*git*) ;;
   *)                              exit 0 ;;
 esac
 
-# Strip quotes and subshell parens so they cannot mask the rm command or path.
-norm=$(printf '%s' "$cmd" | tr -d "'\"()")
+# Strip quotes and bracket metacharacters that could mask paths.
+norm=$(printf '%s' "$_norm" | tr -d "'\"()[]{}")
 
 # shellcheck disable=SC2016  # $HOME in single quotes is intentional: matches literal text, not the shell variable
 # [rR] covers both -rf and -Rf (BSD/macOS rm accepts -R as synonym for -r).
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -1,60 +1,36 @@
-"""Tests for hooks/hooks.json regex blockers (issue #81).
+"""Tests for hooks/bash_guard.sh — end-to-end guard invocation (issue #233).
 
-Each of the three embedded grep -Eq patterns is extracted from the real
-hooks.json and tested with subprocess.run(["grep", "-Eq", ...]) — the same
-regex engine the hook uses — so platform-specific POSIX character classes
-(e.g. [[:space:]]) behave identically.
-
-Note on hard-reset: the hook only exits 2 when BOTH the regex matches AND the
-current branch is main/master/release/*. TestHardResetPatternMatch verifies the
-first gate (regex match); branch-protection logic requires a live git environment
-and is intentionally out of scope for unit tests.
+Each test class invokes hooks/bash_guard.sh directly with a JSON payload on
+stdin, mirroring exactly how Claude Code calls the PreToolUse:Bash hook.
+Exit code 2 + "BLOCKED" in stderr → command was blocked.
+Exit code 0, empty stderr              → command was allowed.
 """
 
 import json
-import re
+import os
+import shutil
 import subprocess
 from pathlib import Path
 
 import pytest
 
-HOOKS_PATH = Path(__file__).parent.parent / "hooks" / "hooks.json"
+GUARD = Path(__file__).parent.parent / "hooks" / "bash_guard.sh"
 
 
 @pytest.fixture(scope="module")
-def hook_patterns():
-    """Return a dict of named grep -Eq patterns extracted from hooks.json.
-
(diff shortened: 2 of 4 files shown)


hooks/bash_guard.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+# Short-circuits (exit 0, no greps) for commands that contain neither "rm" nor
+# "git". This is the common case (ls, cat, echo, make, npm, …) and removes the
+# per-call grep tax flagged in #233.
+#
+# Out of scope: ${HOME} brace-form; path normalization via .. traversal.
+
+set -u
+
+cmd=$(jq -r '.tool_input.command // ""')
```
lugassawan: **High** — jq parse failure silently allows all commands through.

`cmd=$(jq -r '.tool_input.command // ""')` swallows jq's non-zero exit under `set -u` (no `-e`). If stdin is malformed JSON or `jq` is absent, `$cmd` is empty, the `case` short-circuit fires `exit 0`, and every block is skipped. A broken hook silently becomes a no-op guard.

**Suggested fix:** fail-closed instead of fail-open:
```bash
if ! cmd=$(jq -r '.tool_input.command // ""' 2>/dev/null); then
  echo 'bash_guard: jq parse error — blocking by default' >&2
  exit 2
fi
```

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,51 @@
...
+
+case "$cmd" in
+  *rm*|*git*) ;;
+  *)          exit 0 ;;
+esac
+
+# Strip quotes so quoted paths cannot bypass the rm regex.
+norm=$(printf '%s' "$cmd" | tr -d "'\"")
```
lugassawan: **Medium** — ANSI-C `$'...'` quoting survives quote stripping.

`tr -d "'\"" removes only straight single/double quotes. A command like `rm -rf $'/Users/foo'` after stripping becomes `rm -rf $/Users/foo` — the leading `$'` prevents the `/Users` regex from anchoring, so the block is missed.

Claude Code likely never emits this form, but it is a documented shell escape hatch. Consider stripping `$'` ANSI-C openers as well:
```bash
norm=$(printf '%s' "$cmd" | tr -d "'\"" | sed "s/\$'//g")
```
Or document explicitly in the header that ANSI-C quoting is out of scope.

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+# PreToolUse:Bash guard — block destructive commands and short-circuit safe ones.
+#
+# Blocks:
+#   - rm -rf against /, ~, $HOME (literal or shell-expanded /Users/*|/home/*)
```
lugassawan: **Medium** — Documentation gap: header comment doesn't mention `/Users` or `/home` as standalone blocked paths.

The regex's optional username component `(/Users|/home)(/[^/[:space:]]+)?` correctly blocks `rm -rf /Users` (catastrophic), but the header comment at lines 5–7 lists only `/, ~, $HOME`. A future maintainer might miss that bare `/Users` and `/home` are also blocked targets.

**Suggested fix:** update the comment at the top:
```
# Blocks:
#   - rm -rf against /, /*, ~, $HOME, /Users[/<user>], /home[/<user>]
```

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,51 @@
...
+# per-call grep tax flagged in #233.
+#
+# Out of scope: ${HOME} brace-form; path normalization via .. traversal.
+
+set -u
+
+cmd=$(jq -r '.tool_input.command // ""')
+
+case "$cmd" in
+  *rm*|*git*) ;;
```
lugassawan: **Low** — `*rm*` short-circuit matches `npm`, `chmod`, `firmware` etc., adding grep overhead for safe commands.

Strings like `npm run build` and `chmod +x script.sh` contain the substring `rm` and fall through to all grep checks. This is a performance false-positive only — no security bypass is possible — but it partially undercuts the #233 optimization goal.

Consider a word-boundary form to avoid substring matches:
```bash
case "$cmd" in
  rm\ *|*\ rm\ *|*git*) ;;
  *)                    exit 0 ;;
esac
```
This avoids matching `npm`, `chmod`, `from`, etc. while still catching `rm -rf ...` in all leading positions.

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

tests/test_hooks.py
```diff
@@ -101,49 +110,130 @@ class TestForcePushBlocker:
...
+        bin_dir.mkdir()
+        real_jq = shutil.which("jq")
+        real_grep = shutil.which("grep")
+        for tool, real in [("jq", real_jq), ("grep", real_grep)]:
+            shim = bin_dir / tool
+            shim.write_text(
+                f'#!/bin/sh\necho "{tool}" >> "{trace}"\nexec "{real}" "$@"\n'
+            )
+            shim.chmod(0o755)
+        env = os.environ.copy()
```
lugassawan: **Low** — `_run_with_shims` uses `os.environ.copy()` instead of `_CLEAN_ENV`.

All other helpers in the file use `_CLEAN_ENV` to strip `GIT_*` variables, but this method does a raw copy. When tests run inside a git pre-push hook, `GIT_DIR` leaks into the shim environment, which could cause the hard-reset branch check inside `bash_guard.sh` to resolve against the wrong repo — exactly the class of bug `_CLEAN_ENV` was added to prevent.

**Suggested fix:**
```python
env = dict(_CLEAN_ENV)  # was: os.environ.copy()
env["PATH"] = f"{bin_dir}:{_CLEAN_ENV.get('PATH', '')}"
```

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -u
+
+if ! cmd=$(jq -r '.tool_input.command // ""' 2>/dev/null); then
+  echo 'bash_guard: jq parse error — blocking by default' >&2
+  exit 2
+fi
+
+case "$cmd" in
+  rm\ *|*\ rm\ *|*git*) ;;
```
lugassawan: **Medium** — Short-circuit misses `rm` directly after `(` or `{` without a space.

The pattern `rm\ *|*\ rm\ *` requires a space before `rm` (the `*\ ` prefix). A command like `(rm -rf /)` or `{rm -rf /;}` does not match any branch and exits 0 without reaching the blocking regex. The original `*rm*` caught these; the narrowed pattern missed them.

**Suggested fix** — add open-paren as a valid left boundary:
```bash
case "$cmd" in
  rm\ *|*[\ (]rm\ *|*git*) ;;
  *)                         exit 0 ;;
esac
```
`*[\ (]rm\ *` matches a space or `(` immediately before `rm` (covering `(rm -rf /)` and `sudo rm`). `{` without a following space is not a valid shell construct in practice, but `*[\ ({]rm\ *` would cover it explicitly if desired.

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# Short-circuits (exit 0, no greps) for commands that contain neither "rm" nor
+# "git". This is the common case (ls, cat, echo, make, npm, …) and removes the
+# per-call grep tax flagged in #233.
+#
+# Out of scope: ${HOME} brace-form; ANSI-C $'...' quoting; path normalization via .. traversal.
+
+set -u
+
+if ! cmd=$(jq -r '.tool_input.command // ""' 2>/dev/null); then
```
lugassawan: **Low** — `2>/dev/null` discards jq's own error message, leaving only the generic blocking notice.

During incident triage, knowing *why* jq failed (e.g. "unexpected token", "parse error at line 1") is useful. The `2>/dev/null` was added to keep the guard's stderr clean in the normal case, but it silences jq even in the error path.

**Simplest fix:** remove `2>/dev/null` — jq's error naturally flows to stderr, and the `if !` already handles the exit code:
```bash
if ! cmd=$(jq -r '.tool_input.command // ""'); then
  echo 'bash_guard: jq parse error — blocking by default' >&2
  exit 2
fi
```
Both the generic message AND jq's diagnostic appear in stderr on failure. Low priority — correctness is unaffected.

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -u
+
+if ! cmd=$(jq -r '.tool_input.command // ""'); then
+  echo 'bash_guard: jq parse error — blocking by default' >&2
+  exit 2
+fi
+
+case "$cmd" in
+  rm\ *|*\ rm\ *|*\(rm\ *|*git*) ;;
```
lugassawan: **Medium** — `((rm -rf /))` bypasses the short-circuit and exits 0 without reaching the grep blockers.

`*\(rm\ *` matches a single leading paren (`(rm`), but `((rm` has two parens: the outer one is consumed by `*`, the inner `(` is not matched by the literal `\(rm\ *` arm. So `((rm -rf /))` falls through to `exit 0` — entirely skipping the grep — rather than being treated as an rm command. This is a security bypass, not just a missed optimisation.

**Suggested fix** — `*\(*rm\ *` covers any number of leading parens:
```bash
case "$cmd" in
  rm\ *|*\ rm\ *|*\(*rm\ *|*git*) ;;
  *)                                exit 0 ;;
esac
```
`*\(*rm\ *` matches any prefix, then one or more `(` chars (via `*`), then `rm ` — catching `(rm`, `((rm`, `(((rm`, etc. Alternatively, `*rm\ *` (trailing-space word boundary only) is simpler and relies on the grep regex for the left-side boundary, which already requires `(^\|[[:space:]])rm`.

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

tests/test_hooks.py
```diff
@@ -64,34 +47,60 @@ def grep_matches(pattern: str, text: str) -> bool:
 class TestRmRfBlocker:
     @pytest.mark.parametrize("cmd", [
```
lugassawan, line 48: **Low** — No tests assert that parenthesized `rm` forms are blocked.

The round-2 fix added `*\(rm\ *` to cover `(rm -rf /)`, but no test in `TestRmRfBlocker.test_blocked` exercises this form. A double-paren form like `((rm -rf /))` would also go undetected as allowed.

**Suggested fix** — add to the `test_blocked` parametrize list:
```python
"(rm -rf /)",
"((rm -rf /))",
```
This closes the gap where short-circuit fixes can regress silently.

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+  exit 2
+fi
+
+case "$cmd" in
+  rm\ *|*\ rm\ *|*\(*rm\ *|*git*) ;;
+  *)                              exit 0 ;;
+esac
+
+# Strip quotes and subshell parens so they cannot mask the rm command or path.
+norm=$(printf '%s' "$cmd" | tr -d "'\"()")
```
lugassawan: **High** — Bracket-glob bypass: `rm -rf /[U]sers/foo` and `rm -rf /[h]ome/foo` evade both the `case` short-circuit and the blocking regex.

The `case` at line 22–25 correctly passes these strings through (they contain `rm `), but after `tr -d "'\"()"` the square brackets are preserved. So `rm -rf /[U]sers/foo` does not match `/Users` in the regex. When bash later executes the command, `[U]sers` glob-expands to `Users`, making this a working destructive bypass.

**Fix:** add `[]` and `{}` to the `tr -d` character class:
```bash
norm=$(printf '%s' "$cmd" | tr -d "'\"()[]{}")
```

**Test cases to add to `TestRmRfBlocker.test_blocked`:**
```python
"rm -rf /[U]sers/foo",
"rm -rf /[h]ome/foo",
"rm -rf /[h]ome",
```

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -u
+
+if ! cmd=$(jq -r '.tool_input.command // ""'); then
+  echo 'bash_guard: jq parse error — blocking by default' >&2
+  exit 2
+fi
+
+case "$cmd" in
+  rm\ *|*\ rm\ *|*\(*rm\ *|*git*) ;;
```
lugassawan, line 26: **Medium** — The `case` short-circuit misses `ls;rm -rf /` and `ls&&rm -rf /` (no space between the shell operator and `rm`).

The pattern `*\ rm\ *` requires a literal space on both sides of `rm`. Commands with no space between a shell operator and `rm` (e.g., `ls;rm`, `ls&&rm`, `ls||rm`) miss all four `case` arms and fall through to `exit 0`. The regex also misses them — `[[:space:]]` does not match `;`. So no second gate exists.

**Fix:** extend the `case` arm with a semicolon/pipe/ampersand alternative:
```bash
case "$cmd" in
  rm\ *|*\ rm\ *|*[;|&]rm\ *|*\(*rm\ *|*git*) ;;
  *)                              exit 0 ;;
esac
```

No test currently covers the `ls;rm -rf /` variant in `TestRmRfBlocker`.

hooks/bash_guard.sh
```diff
@@ -0,0 +1,54 @@
...
+  exit 2
+fi
+
+if echo "$norm" | grep -Eq 'git[[:space:]]+push.*(--force([[:space:]]|$)|(^|[[:space:]])-f([[:space:]]|$))' \
+   && echo "$norm" | grep -Eq '(^|[[:space:]]|:)(main|master)([[:space:]]|:|$)'; then
+  echo 'BLOCKED: force push to main/master' >&2
+  exit 2
+fi
+
+if echo "$norm" | grep -Eq 'git[[:space:]]+reset[[:space:]]+--hard'; then
```
lugassawan, line 47: **Low** — `git rev-parse --abbrev-ref HEAD` runs in the guard's working directory, not necessarily where Claude Code will execute the command.

If the Bash tool runs `cd /some/repo && git checkout -B feature/x && git reset HEAD~1` (or similar compound commands that change `cwd`), the guard's `git rev-parse` reads the branch from the guard's own `cwd`, not from the target repo. This can cause a spurious block on a feature branch or silently allow a reset on a different protected branch.

**Suggestion:** document this limitation in the header 'Out of scope' block:
```bash
# Out of scope: ...; compound commands with cd (cwd not exposed in hook payload)
```

Not a blocker — just worth noting in the documentation.

PR #266, opened 2026-05-17
[polish] workflow-pr-review: use AskUserQuestion for address-feedback CTA

lugassawan:
## Summary
- Replace the free-text `"Reply yes to start /address-feedback N"` CTA in `workflow-pr-review` and `workflow-pr-review-followup` with an `AskUserQuestion` call.
- The user now gets a clickable `Yes — address feedback` / `No thanks` button pair instead of having to type manually.
- `AskUserQuestion` is already used for the draft-vs-ready prompt in `workflow-commit-and-pr`; this change brings the address-feedback handoff in line with that UX pattern.
- Common Mistakes table updated with a new row: "Emit the CTA as plain text instead of AskUserQuestion" to guard against future regression.

## Test Plan
- [x] Changed skills/commands/agents load without errors
- [x] Examples in the diff were exercised manually
- [x] `pytest tests/ -k workflow_pr_review` — 14/14 pass (no regressions)

### Type-tailored checks ([polish])
- [x] `AskUserQuestion` JSON block appears in both SKILL.md files
- [x] `<N>` placeholder is consistently documented (substitute real PR number)
- [x] `Yes — address feedback` option invokes `/address-feedback <N>`
- [x] `No thanks` option performs no further action
- [x] Old free-text `"Reply yes"` pattern is absent from both files
- [x] Common Mistakes table addition accurately describes the failure mode

Depends on: #264 (must merge first — this branch is based on it)

Issue: N/A — UX improvement; no issue filed

PR #267, opened 2026-05-17
[fix] workflow-cleanup-merged: handle bare repos in sync-and-verify.sh

lugassawan:
## Summary
- `sync-and-verify.sh` Block B called `git checkout $DEFAULT_BRANCH && git pull --ff-only origin $DEFAULT_BRANCH`, which aborts with "this operation must be run in a work tree" when the main repo is a bare clone (rimba worktree layout).
- Fix: detect `core.bare=true` and update the local branch ref via a force-fetch refspec (`git fetch origin +refs/heads/$DEFAULT_BRANCH:refs/heads/$DEFAULT_BRANCH`) instead — semantically equivalent to checkout+pull--ff-only but works without a working tree.
- Adds a contract test pinning the bare-repo detection path in `test_skill_repo_ambiguity.py`.

## Test Plan
- [x] `pytest tests/test_skill_repo_ambiguity.py -v` — 6/6 pass (includes new `test_sync_and_verify_handles_bare_repo`)
- [x] `pytest tests/ -q` — 520/520 pass
- [x] Script exercised live against the bare swe-workbench repo during PR #266 cleanup — no more "this operation must be run in a work tree" error; `WORKTREE_GONE` resolved correctly

Issue: N/A — bug observed during PR #266 post-merge cleanup in bare+worktrees repo


skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -15,8 +15,15 @@ MAIN_REPO=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
 cd "$MAIN_REPO"
 
 # Block B: sync local default branch (best-effort — failure warns, does not abort)
-(git checkout "$DEFAULT_BRANCH" && git pull --ff-only origin "$DEFAULT_BRANCH") >/dev/null \
-  || echo "sync-main: best-effort failed — reconcile $DEFAULT_BRANCH manually" >&2
+# Bare repos (rimba worktree layout) have no working tree; use a force-fetch refspec to
+# advance the local ref instead of checkout+pull.
+if git config --get core.bare 2>/dev/null | grep -q '^true$'; then
+  git fetch --quiet origin "+refs/heads/${DEFAULT_BRANCH}:refs/heads/${DEFAULT_BRANCH}" 2>/dev/null \
```
lugassawan, line 25: **Critical — hook fast-path silently broken for bare repos**

`git fetch` does not fire the `post-merge` hook — only `git merge`/`git pull` (which internally calls merge) does. In the rimba bare-repo layout this line now runs instead of `git pull --ff-only`, so Block C will always produce `WORKTREE_GONE=0` even when the rimba post-merge hook is configured. The hook fast-path is silently bypassed on every bare-repo cleanup run without any warning.

This does not break correctness — the fallback strategies (rimba-binary or shell) still handle cleanup — but it contradicts the SKILL.md contract that documents `git pull` firing the hook as a side-effect.

**Suggested fix:** Add a comment here noting that `git fetch` bypasses the post-merge hook, so bare-repo layout always falls through to the rimba-binary or shell strategy in Block C:

```bash
# Note: git fetch does not fire post-merge, so WORKTREE_GONE will always be 0
# for bare repos — Block C falls through to the rimba-binary or shell strategy.
```

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -15,8 +15,18 @@ MAIN_REPO=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
 cd "$MAIN_REPO"
 
 # Block B: sync local default branch (best-effort — failure warns, does not abort)
-(git checkout "$DEFAULT_BRANCH" && git pull --ff-only origin "$DEFAULT_BRANCH") >/dev/null \
-  || echo "sync-main: best-effort failed — reconcile $DEFAULT_BRANCH manually" >&2
+# Bare repos (rimba worktree layout) have no working tree; use a force-fetch refspec to
+# advance the local ref instead of checkout+pull.
+if git config --get core.bare 2>/dev/null | grep -q '^true$'; then
+  # git fetch does not fire post-merge, so WORKTREE_GONE=1 via the hook fast-path
+  # is not achievable in bare-repo layout. Block C will always get WORKTREE_GONE=0
```
lugassawan: **Important — "always" claim is overstated**

Lines 22–23 say: "Block C will always get `WORKTREE_GONE=0` and fall through to the rimba-binary or shell removal strategy."

This is not accurate. Block C sets `WORKTREE_GONE=1` whenever both the worktree and local branch are absent — regardless of *how* they became absent. If a prior run, `git worktree remove`, or a concurrent cleanup already removed them before this script runs, Block C can yield `WORKTREE_GONE=1` even in bare-repo layout.

The correct claim is narrower: the **hook fast-path** cannot set `WORKTREE_GONE=1` in bare-repo layout (because `git fetch` does not fire `post-merge`). Block C itself is not limited.

**Suggested fix:**
```
# git fetch does not fire post-merge, so WORKTREE_GONE=1 via the hook fast-path
# is not achievable in bare-repo layout. Block C may still yield WORKTREE_GONE=1
# if the worktree and branch were already removed by other means; otherwise it
# falls through to the rimba-binary or shell removal strategy.
```

Same wording fix applies to `SKILL.md:78` ("Block C will therefore always yield `WORKTREE_GONE=0`" → "Block C will not get `WORKTREE_GONE=1` via the hook fast-path").

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
lugassawan: Addressed in 60afa9a: added a comment on the bare-repo fetch block explicitly stating that `git fetch` does not fire `post-merge`, so bare-repo layout always falls through to the rimba-binary or shell removal strategy. Also updated SKILL.md:78 to note this constraint. The behaviour is correct — fallback strategies handle cleanup — but the missing documentation was the gap.

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
lugassawan: Addressed in 7beb52e: corrected the comment from 'Block C will always get WORKTREE_GONE=0' to 'Block C may still yield WORKTREE_GONE=1 if the worktree and branch were already removed by other means'. Applied the same narrowing to SKILL.md:78 ('Block C will not get WORKTREE_GONE=1 via the hook fast-path').

lugassawan:
Closing as no longer required. The bare-repo scenario this PR defends against was a test environment artifact — `GIT_DIR` leaking from a git hook process into subprocess calls in tests caused temp repos to inherit `core.bare=true`. The root cause was addressed in #265 and will be further hardened via #268 (`_CLEAN_ENV` centralisation). In production, `sync-and-verify.sh` always runs against a non-bare repo; the `core.bare` branch added here would never be taken.

PR #269, opened 2026-05-17
[chore] Extract _CLEAN_ENV to shared tests/conftest.py

### LINKED ISSUE #268: [chore] Extract _CLEAN_ENV to shared conftest fixture
## Product framing

**Problem:** Test files across the repo (`test_setup.py`, `test_skill_usage_telemetry.py`, `test_sync_and_verify.py`, `test_setup_sh.py`) independently declare and maintain their own `_CLEAN_ENV` constant — a filtered environment that strips `GIT_*` variables to prevent git config leakage when subprocess calls create temp repos. This pattern is duplicated, undocumented, and fragile: the bare-repo + `GIT_DIR` inheritance issue is invisible to new test files.

**Value:** Centralizing this utility in a shared `conftest.py` or module eliminates duplication, ensures all current and future test files get the protection automatically, and documents WHY it's needed (bare-repo environment variable inheritance issue). Teams onboarding to the repo will understand the constraint immediately.

**Acceptance criteria:**
- [ ] `tests/conftest.py` exists and exports `_CLEAN_ENV` as a module-level constant with inline documentation explaining the bare-repo + `GIT_DIR` leakage issue.
- [ ] All four existing test files (`test_setup.py`, `test_skill_usage_telemetry.py`, `test_sync_and_verify.py`, `test_setup_sh.py`) import `_CLEAN_ENV` from conftest instead of declaring it locally.
- [ ] Documentation added to repo docs or `tests/README.md` explaining when and why to use `_CLEAN_ENV` in subprocess calls.
- [ ] No regressions: all tests pass after consolidation.

**Impact / Effort:**
- Impact: **M** — reduces test fragility, improves onboarding clarity, prevents future bug reintroduction across the test suite.
- Effort: **S** — pure refactor; ~30 lines of code + docs; low risk of breaking change.

---

### Type

- [x] Documentation (README, catalog, SECURITY, etc.)
- [ ] CI / workflow (`.github/workflows/`, actions)
- [ ] Dependency / tooling bump
- [ ] Repository config (templates, Dependabot, labels)
- [ ] Release / packaging (`.claude-plugin/`, `scripts/`)

### Surface

**Files to change:**
- `tests/conftest.py` (create)
- `tests/test_setup.py` (import from conftest, remove local `_CLEAN_ENV`)
- `tests/test_skill_usage_telemetry.py` (import from conftest, remove local `_CLEAN_ENV`)
- `tests/test_sync_and_verify.py` (import from conftest, remove local `_CLEAN_ENV`)
- `tests/test_setup_sh.py` (import from conftest, remove local `_CLEAN_ENV`)
- `tests/README.md` (document or create; explain git env leakage + workaround)

### Description

The repo is a bare git repo (`core.bare = true` in `.git/config`). When tests run inside a git hook (e.g., pre-push), the hook process exports `GIT_DIR` pointing to the bare repo. Any subprocess calls (e.g., `subprocess.run(["git", "init", ...])` in a temp directory) inherit this environment variable. Git then treats the temp directory as a bare repo too, unexpectedly setting `core.bare = true` on temporary test repos.

The workaround is simple: strip all `GIT_*` environment variables before subprocess calls. This protects against both the immediate bare-repo issue and future git environment leakage.

Currently, each test file independently declares `_CLEAN_ENV = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}` at module level. This duplication is fragile — new test files won't know to apply the workaround, and the *why* is lost.

Solution: Extract `_CLEAN_ENV` into `tests/conftest.py` with clear documentation of the root cause and usage pattern. All test files import it, and all *future* test files benefit automatically.

### Acceptance

- [x] `tests/conftest.py` created with `_CLEAN_ENV` constant and docstring explaining bare-repo + `GIT_DIR` inheritance.
- [x] All four test files refactored to import `_CLEAN_ENV` from conftest.
- [x] Documentation added to `tests/README.md` (or docs) explaining the git env leakage issue and when to use `_CLEAN_ENV` in subprocess calls.
- [x] All tests pass after refactor (verify via CI or local `pytest tests/`).

lugassawan:
## Summary

- Centralise `_CLEAN_ENV` in `tests/conftest.py` with inline docs explaining the bare-repo / `GIT_DIR` inheritance root cause
- Migrate 5 test files that each had a local copy (`test_hooks.py`, `test_setup.py`, `test_setup_sh.py`, `test_skill_usage_telemetry.py`, `test_sync_and_verify.py`) to `from conftest import _CLEAN_ENV`
- Add net-new protection to `test_release_sh_gh_repo.py` (previously used unfiltered `os.environ` in a `bash -c` subprocess that calls `git remote get-url`)
- Promote `GIT_CONFIG_NOSYSTEM=1` from `test_sync_and_verify.py`-only to a baseline for all 6 subprocess sites
- Document the pattern in `tests/README.md` so future authors learn the constraint without debugging a flaky hook-context CI run

Scope superset vs ticket: ticket #268 named 4 files; also covers `test_hooks.py` (5th duplicate) and `test_release_sh_gh_repo.py` (new protection).

## Test Plan

- [x] Import probe: `python -c "from tests.conftest import _CLEAN_ENV; ..."` → `['GIT_CONFIG_NOSYSTEM']` only
- [x] Grep proof: `grep -rn "_CLEAN_ENV.*=.*{" tests/ --include='*.py'` → exactly 1 match (conftest.py)
- [x] Callers: `grep -rln "from conftest import _CLEAN_ENV" tests/` → 6 files
- [x] Lint: `python scripts/validate.py` → exit 0
- [x] Tests: `pytest tests/ -v` → 547/547 pass (0 regressions; +62 from `test_hooks.py` guard_script tests previously masked by a missing `import os`)

Closes #268

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,8 @@
 import os
 import sys
 from pathlib import Path
+from types import MappingProxyType
+from typing import Final
 
 import pytest
 
@@ -12,11 +14,15 @@ import pytest
 # We strip every GIT_* var, then re-add GIT_CONFIG_NOSYSTEM=1 to ignore the
 # system gitconfig for hermeticity.
 #
+# Snapshot: built once from os.environ at pytest collection time. Session-scoped
+# fixtures that mutate GIT_* vars after import will not be reflected here.
+#
 # Usage:
 #   subprocess.run([...], env=_CLEAN_ENV, ...)
 #   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
-_CLEAN_ENV: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
-_CLEAN_ENV["GIT_CONFIG_NOSYSTEM"] = "1"
+_clean: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+_clean["GIT_CONFIG_NOSYSTEM"] = "1"
+_CLEAN_ENV: Final[MappingProxyType[str, str]] = MappingProxyType(_clean)
 
 # Ensure scripts/ and tests/ are importable from any working directory.
 _HERE = Path(__file__).parent
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,22 @@
+import os
 import sys
 from pathlib import Path
 
 import pytest
+
+# _CLEAN_ENV: subprocess-safe environment for tests that spawn git or bash.
+#
+# swe-workbench is a bare git repo (core.bare=true). When tests run inside a
+# git hook (e.g. pre-push), GIT_DIR is exported and inherited by subprocess
+# children — Git then treats temp test repos as bare too, causing failures.
+# We strip every GIT_* var, then re-add GIT_CONFIG_NOSYSTEM=1 to ignore the
+# system gitconfig for hermeticity.
+#
+# Usage:
+#   subprocess.run([...], env=_CLEAN_ENV, ...)
+#   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
+_CLEAN_ENV: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+_CLEAN_ENV["GIT_CONFIG_NOSYSTEM"] = "1"
 
 # Ensure scripts/ and tests/ are importable from any working directory.
 _HERE = Path(__file__).parent
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,9 +20,10 @@ import pytest
 # Usage:
 #   subprocess.run([...], env=_CLEAN_ENV, ...)
 #   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
-_clean: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
-_clean["GIT_CONFIG_NOSYSTEM"] = "1"
-_CLEAN_ENV: Final[MappingProxyType[str, str]] = MappingProxyType(_clean)
+_CLEAN_ENV: Final[MappingProxyType[str, str]] = MappingProxyType(
+    {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+    | {"GIT_CONFIG_NOSYSTEM": "1"}
+)
 
 # Ensure scripts/ and tests/ are importable from any working directory.
 _HERE = Path(__file__).parent
--- a/tests/test_release_sh_gh_repo.py
+++ b/tests/test_release_sh_gh_repo.py
@@ -4,7 +4,6 @@ Regression tests for the GH_REPO export in scripts/release.sh.
 Guards against "No default remote repository has been set" failures
 on multi-remote clones (e.g. after `gh repo fork` adds sibling remotes).
 """
-import os
 import re
 import shlex
 import shutil
@@ -12,6 +11,8 @@ import subprocess
 import tempfile
 import textwrap
 from pathlib import Path
+
+from conftest import _CLEAN_ENV
 
 RELEASE_SH = Path(__file__).parent.parent / "scripts" / "release.sh"
 
@@ -103,7 +104,7 @@ class TestExtractionRoundTrip:
                 f"#!/bin/sh\nprintf '%s\\n' {shlex.quote(fake_url)}\n"
             )
             stub.chmod(0o755)
-            env = {**os.environ, "PATH": f"{stub_dir}:{os.environ['PATH']}"}
+            env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV['PATH']}"}
             result = subprocess.run(
                 ["bash", "-c", self._SNIPPET],
                 capture_output=True,
--- a/tests/test_release_sh_gh_repo.py
+++ b/tests/test_release_sh_gh_repo.py
@@ -104,7 +104,7 @@ class TestExtractionRoundTrip:
                 f"#!/bin/sh\nprintf '%s\\n' {shlex.quote(fake_url)}\n"
             )
             stub.chmod(0o755)
-            env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV['PATH']}"}
+            env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV.get('PATH', '')}"}
             result = subprocess.run(
                 ["bash", "-c", self._SNIPPET],
                 capture_output=True,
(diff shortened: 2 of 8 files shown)


tests/conftest.py
```diff
@@ -1,8 +1,23 @@
...
+# git hook (e.g. pre-push), GIT_DIR is exported and inherited by subprocess
+# children — Git then treats temp test repos as bare too, causing failures.
+# We strip every GIT_* var, then re-add GIT_CONFIG_NOSYSTEM=1 to ignore the
+# system gitconfig for hermeticity.
+#
+# Usage:
+#   subprocess.run([...], env=_CLEAN_ENV, ...)
+#   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
+_CLEAN_ENV: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+_CLEAN_ENV["GIT_CONFIG_NOSYSTEM"] = "1"
```
lugassawan: **Medium — mutable shared dict is a latent footgun**

`_CLEAN_ENV` is a plain `dict` shared across the entire test session. Any test that accidentally does `_CLEAN_ENV["KEY"] = val` (rather than `{**_CLEAN_ENV, "KEY": val}`) silently pollutes all later tests in the same process. The `dict(_CLEAN_ENV)` copy idiom in `test_hooks.py` is correct, but it is not enforced.

Wrapping with `MappingProxyType` turns accidental mutation into an immediate `TypeError`:

```python
from types import MappingProxyType

_clean = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
_clean["GIT_CONFIG_NOSYSTEM"] = "1"
_CLEAN_ENV: MappingProxyType[str, str] = MappingProxyType(_clean)
```

All callers using `{**_CLEAN_ENV, ...}` keep working unchanged.

tests/conftest.py
```diff
@@ -1,8 +1,23 @@
...
+# swe-workbench is a bare git repo (core.bare=true). When tests run inside a
+# git hook (e.g. pre-push), GIT_DIR is exported and inherited by subprocess
+# children — Git then treats temp test repos as bare too, causing failures.
+# We strip every GIT_* var, then re-add GIT_CONFIG_NOSYSTEM=1 to ignore the
+# system gitconfig for hermeticity.
+#
+# Usage:
+#   subprocess.run([...], env=_CLEAN_ENV, ...)
+#   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
+_CLEAN_ENV: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
```
lugassawan: **Low — consider `Final` annotation and document snapshot-at-import semantics**

Two related observations on this line:

1. **`Final` annotation** — `_CLEAN_ENV` reads as a constant but lacks `Final`, so nothing prevents a module-level reassignment being caught by a type-checker:
   ```python
   from typing import Final
   _CLEAN_ENV: Final[dict[str, str]] = {...}
   ```
   If the `MappingProxyType` fix above is adopted, use `Final[MappingProxyType[str, str]]` instead.

2. **Snapshot semantics** — `_CLEAN_ENV` is built once from `os.environ` at module-import time (during pytest collection). Any session-scoped fixture that sets `GIT_*` vars *before* the first test but *after* import would not be reflected. This is an unlikely edge case in the current suite, but worth a one-liner in the comment block above so future authors don't get surprised.

tests/conftest.py
lugassawan: Addressed in 84535a33101ee564c1adcc105b17ad02fb4534ee: wrapped with `MappingProxyType` so any accidental in-place mutation raises `TypeError` immediately. Also added `Final[MappingProxyType[str, str]]` annotation and a snapshot-at-import note to the comment block.

tests/conftest.py
lugassawan: Addressed in 84535a33101ee564c1adcc105b17ad02fb4534ee: added `Final[MappingProxyType[str, str]]` annotation and a one-liner snapshot-at-import note to the comment block above the declaration.

tests/conftest.py
```diff
@@ -1,8 +1,29 @@
...
+# We strip every GIT_* var, then re-add GIT_CONFIG_NOSYSTEM=1 to ignore the
+# system gitconfig for hermeticity.
+#
+# Snapshot: built once from os.environ at pytest collection time. Session-scoped
+# fixtures that mutate GIT_* vars after import will not be reflected here.
+#
+# Usage:
+#   subprocess.run([...], env=_CLEAN_ENV, ...)
+#   subprocess.run([...], env={**_CLEAN_ENV, "KEY": "val"}, ...)
+_clean: dict[str, str] = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
```
lugassawan: **Low — `_clean` is still a mutable module-level name**

`MappingProxyType` wraps a live dict — any code that imports `conftest` and writes `conftest._clean["FOO"] = "bar"` would silently corrupt `_CLEAN_ENV`. The protection is shallower than it appears.

Inlining eliminates the intermediate name entirely:

```python
_CLEAN_ENV: Final[MappingProxyType[str, str]] = MappingProxyType(
    {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
    | {"GIT_CONFIG_NOSYSTEM": "1"}
)
```

(The `|=` dict-merge operator requires Python 3.9+, which this codebase already targets.)

tests/test_release_sh_gh_repo.py
```diff
@@ -103,7 +104,7 @@ def _run(self, fake_url: str) -> str:
                 f"#!/bin/sh\nprintf '%s\\n' {shlex.quote(fake_url)}\n"
             )
             stub.chmod(0o755)
-            env = {**os.environ, "PATH": f"{stub_dir}:{os.environ['PATH']}"}
+            env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV['PATH']}"}
```
lugassawan: **Low — `_CLEAN_ENV['PATH']` raises `KeyError` if `PATH` is absent**

`PATH` is present in virtually all POSIX environments, but is not guaranteed — a minimal container or a test fixture that strips it would turn this into an opaque `KeyError` rather than a meaningful test failure.

Safer options:
```python
# Option A — fallback empty string (subprocess inherits nothing for PATH)
env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV.get('PATH', '')}"}

# Option B — assert at test setup (loud failure with a clear message)
assert "PATH" in _CLEAN_ENV, "PATH missing from _CLEAN_ENV"
env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV['PATH']}"}
```

tests/conftest.py
lugassawan: Addressed in 74ac124089fe458021e16dae811d5b764a285fe7: inlined the dict literal directly into `MappingProxyType({...} | {"GIT_CONFIG_NOSYSTEM": "1"})` — `_clean` no longer exists as a module-level name, so there is nothing mutable to reach through the proxy.

tests/test_release_sh_gh_repo.py
lugassawan: Addressed in 74ac124089fe458021e16dae811d5b764a285fe7: switched to `_CLEAN_ENV.get('PATH', '')` to avoid `KeyError` in minimal environments.

PR #271, opened 2026-05-17
[ci] Route nightly skill-triggers failures to a tracking issue

### LINKED ISSUE #232: [chore] Add failure routing to nightly skill-triggers harness
## Type

## Surface

**Files to change:**
- `.github/workflows/skill-triggers.yml`

## Description

`.github/workflows/skill-triggers.yml` runs the `Skill Auto-Trigger Harness` daily at 07:00 UTC via `cron`. When `pytest tests/test_skill_triggers.py` fails (description drift, BM25 ranking change), the only signal is a red X on the Actions tab. There is no `if: failure()` step that opens an issue, posts to a discussion, or otherwise routes the failure.

The PR-time `pytest` job in `pr.yml` catches per-PR drift, but not **corpus-relative BM25 drift** that accumulates across many merged PRs — that's exactly what the daily cron is supposed to catch. Without failure routing, the catch is wasted: a maintainer not actively watching Actions will not notice description drift until they ship a release with a no-longer-triggering skill.

## Acceptance

- [ ] Add an `if: failure()` step to `.github/workflows/skill-triggers.yml` that uses `gh issue create` (or a maintained action) to file/append to a tracking issue with the failing parametrized test IDs.
- [ ] Issue body should include the failing test IDs and a link back to the workflow run.
- [ ] Choose between "file a new issue per failure" (noisier, easier to triage) or "append to a rolling tracking issue" (quieter, harder to forget).

---

## Audit reasoning (full 11-field schema)

| Field | Value |
|-------|-------|
| **title** | Nightly BM25 trigger harness fails silently — no issue, no notification |
| **severity** | Medium |
| **domain** | reliability |
| **file_line** | `.github/workflows/skill-triggers.yml:1-21` |
| **symptom** | Cron failures produce only a red X on the Actions tab — no issue, no chat hook, no email. |
| **root_cause** | Workflow wired for execution but not observability. The harness's purpose is to detect drift between daily runs and the next release — but without alerting, detection is lossy. |
| **reasoning_chain** | 1. Workflow runs only on cron + dispatch — not on PRs touching `skills/*/SKILL.md` (pr.yml's pytest covers those). 2. So this is a daily safety net, not a PR gate. 3. No `if: failure()` step, no issue-creating action, no chat hook. 4. GitHub's default Actions-failure notification is opt-in. 5. Description drift is exactly the slow-decay class a cron is meant to catch — wasted if no one is paged. |
| **counter_evidence_considered** | Is the maintainer subscribed to Actions failures? Only if opt-in. Does pr.yml suffice? It catches per-PR drift but not corpus-relative BM25 drift. Value in a daily run without alerting? Marginal — forensic timing only. |
| **confidence** | 0.70 |
| **effort** | low |
| **suggested_fix** | Add an `if: failure()` step using `gh issue create` (or a maintained action) to file/append to a tracking issue with the failing parametrized test IDs. |

lugassawan:
## Summary
- Adds `issues: write` permission to the `smoke` job so the failure step can create labels and issues.
- Pipes `pytest` output through `tee /tmp/harness.log` so the failure step can parse `FAILED` lines; pytest exit code is preserved because `pipefail` is on by default in GitHub Actions bash.
- Adds an `if: failure()` step that ensures the `skill-trigger-drift` label exists (check-then-create, idempotent), looks up the first open issue with that label, and comments on it or files a new one — implementing the rolling tracking-issue strategy so repeated nightly failures from the same drift episode accumulate in one thread rather than spawning duplicates.

## Test Plan
- [ ] YAML parses: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/skill-triggers.yml'))"` exits 0.
- [ ] Permissions diff confirms only `issues: write` added; no other scope widened.
- [ ] `GH_TOKEN` and `RUN_URL` are passed through `env:` (not inline `${{ }}` in `run:`) — safe against injection.
- [ ] End-to-end dry-run: trigger `workflow_dispatch` on this branch with a deliberate failure → confirm issue filed with label, body contains test IDs and run URL → trigger again → confirm comment appended to existing issue rather than new issue opened.

### Type-tailored checks (ci)
- [ ] Workflow YAML is syntactically valid and parses without warnings.
- [ ] `if: failure()` step is correctly gated — only fires on upstream step failure, not on success.
- [ ] `gh label create` is idempotent: running twice does not error on "already exists".
- [ ] `|| true` is applied only to the `grep` that may legitimately return empty; `set -euo pipefail` is present to fail loud on real errors.

Closes #232

### DIFF
--- a/.github/workflows/skill-triggers.yml
+++ b/.github/workflows/skill-triggers.yml
@@ -65,6 +65,7 @@ jobs:
           existing=$(gh issue list \
             --label skill-trigger-drift \
             --state open \
+            --limit 1 \
             --json number \
             --jq '.[0].number // empty')
 
--- a/.github/workflows/skill-triggers.yml
+++ b/.github/workflows/skill-triggers.yml
@@ -32,7 +32,7 @@ jobs:
           set -euo pipefail
 
           # Ensure the tracking label exists (idempotent; fails loud on real errors)
-          if ! gh label list --json name --jq '.[].name' | grep -qx skill-trigger-drift; then
+          if ! gh label list --limit 200 --json name --jq '.[].name' | grep -qx skill-trigger-drift; then
             gh label create skill-trigger-drift \
               --color fbca04 \
               --description 'Nightly BM25 harness detected description drift'
@@ -75,6 +75,6 @@ jobs:
             echo "Creating new tracking issue"
             gh issue create \
               --title "Nightly skill-trigger harness failing" \
-              --label skill-trigger-drift,ci \
+              --label skill-trigger-drift \
               --body "$body"
           fi
--- a/.github/workflows/skill-triggers.yml
+++ b/.github/workflows/skill-triggers.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      issues: write
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/skill-triggers.yml
```diff
@@ -11,11 +11,70 @@ jobs:
...
+            --jq '.[0].number // empty')
+
+          if [ -n "$existing" ]; then
+            echo "Appending to existing tracking issue #${existing}"
+            gh issue comment "$existing" --body "$body"
+          else
+            echo "Creating new tracking issue"
+            gh issue create \
+              --title "Nightly skill-trigger harness failing" \
+              --label skill-trigger-drift,ci \
```
lugassawan: **Medium — `ci` label hardcoded without existence guard**

`--label skill-trigger-drift,ci` assumes the `ci` label exists. If it is ever renamed or deleted, `gh issue create` returns a non-zero exit code and `set -euo pipefail` aborts the routing step — leaving the failure unreported.

**Options:**
1. Mirror the idempotent-create pattern already used for `skill-trigger-drift`, adding a guard for `ci` as well.
2. Drop `ci` from the create call and rely solely on `skill-trigger-drift` for programmatic filtering (simpler):

```suggestion
              --label skill-trigger-drift \
```

.github/workflows/skill-triggers.yml
```diff
@@ -11,11 +11,70 @@ jobs:
...
+      - name: Route failure to tracking issue
+        if: failure()
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+
+          # Ensure the tracking label exists (idempotent; fails loud on real errors)
+          if ! gh label list --json name --jq '.[].name' | grep -qx skill-trigger-drift; then
```
lugassawan: **Low — `gh label list` uses the default `--limit 30`**

If the repository's label count grows past 30, `skill-trigger-drift` will not appear in the output. The `grep -qx` check then fails to find it, causing `gh label create` to be called on a label that already exists — which returns a 422 error that aborts the step via `set -euo pipefail`.

Fix: add `--limit 200` (or any value comfortably above the realistic label ceiling):

```suggestion
          if ! gh label list --limit 200 --json name --jq '.[].name' | grep -qx skill-trigger-drift; then
```

.github/workflows/skill-triggers.yml
lugassawan: Addressed in e8f9f66: dropped `ci` from the `--label` flag; `skill-trigger-drift` alone is sufficient for programmatic filtering and avoids a silent abort if `ci` is ever renamed or deleted.

.github/workflows/skill-triggers.yml
lugassawan: Addressed in e8f9f66: added `--limit 200` to `gh label list` to prevent false 422s if the repository's label count grows past the default 30.

.github/workflows/skill-triggers.yml
```diff
@@ -11,11 +11,70 @@ jobs:
...
+          **Run:** ${RUN_URL}
+
+          ---
+
+          Usually means a skill's description has drifted (BM25 ranks a sibling skill above the target). Edit the relevant \`skills/<name>/SKILL.md\` description and re-run via \`workflow_dispatch\`.
+          EOF
+          )
+
+          # Look up open tracking issue
+          existing=$(gh issue list \
```
lugassawan, line 65: **Low — `gh issue list` fetches up to 30 results when only the first is needed**

No `--limit` flag is set here, so `gh` defaults to 30 results. Since only `.[0].number` is consumed by jq, 29 results are fetched and discarded on every failure run.

Adding `--limit 1` makes the intent explicit:

```suggestion
          existing=$(gh issue list \
            --label skill-trigger-drift \
            --state open \
            --limit 1 \
            --json number \
            --jq '.[0].number // empty')
```

lugassawan: Addressed in 8dc9c13: added `--limit 1` to `gh issue list` — only the first open tracking issue is needed.

PR #274, opened 2026-05-17
[fix] Fix two gh API reliability races in release.sh

### LINKED ISSUE #228: [bug] release.sh: gh pr checks error suppression + empty mergeCommit.oid race
## Surface affected

- [x] Build / CI / release

**Specific surface:** `scripts/release.sh` (lines 178-194 and 247-254)

## Claude Code version

N/A — defect is in repo automation, not Claude Code itself.

## Plugin version

0.1.11

## Reproduction steps

These are two related but distinct races, both surfaced by a cold-start audit and confirmed by depth-first `debugger` fan-out.

### Race A — `gh pr checks` errors silently coerced to "0 pending, 0 failed" (lines 178-183)

1. Run `scripts/release.sh` against an open bump PR.
2. During the CI-wait loop, induce a `gh` failure (revoke `GH_TOKEN`, drop network, or trigger a 5xx response from GitHub).
3. Observe: `PENDING=$(gh pr checks ... 2>/dev/null || echo "0")` swallows the gh error and substitutes literal `"0"`. Loop body at line 194 prints `All CI checks passed. Waiting for branch-protection mergeability...` — a **false claim**.
4. Inner mergeability poll at 197-224 then either (a) silently merges if gh recovers and CI was in fact green, or (b) aborts after 120s with a wrong diagnosis ("not mergeable" rather than "gh was unreliable").

Additional subtlety: `gh pr checks` documents exit code **8** for "at least one check failed (JSON still valid)". The current `|| echo "0"` collapses that legitimate failure exit with transport-layer errors — the script cannot distinguish "checks failed" from "API was flaky".

### Race B — Empty `mergeCommit.oid` aborts release with false mismatch (lines 247-254)

1. `gh pr merge --squash --delete-branch "$PR_NUM"` succeeds at line 227 — GitHub records the merge.
2. `git checkout main; git pull --ff-only origin main` at lines 244-245 fast-forwards local main to the new squash commit.
3. `gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid'` at line 247 fires immediately. GitHub's GraphQL `pullRequest.mergeCommit` field is **eventually consistent** for squash merges — under high-load periods or partial GitHub incidents it returns `null`, so `MERGE_SHA=""`.
4. Line 250's `[[ "$LOCAL_SHA" != "$MERGE_SHA" ]]` trivially evaluates true.
5. Script aborts at line 253 with `Something unexpected was pushed to main. Aborting tag to avoid tagging wrong commit.` — a **misleading tampering accusation**.
6. Operator-visible state: PR merged, version bumped, **no tag pushed**, tag-triggered `release.yml` never fires.

## Expected behavior

- gh transport failures should re-poll the CI-wait loop, not silently advance.
- gh's documented "checks failed" exit (8) should map to the existing FAILED-exit branch at lines 186-191, not to the "all green" path.
- Empty `mergeCommit.oid` should trigger a bounded poll (eventually consistent), not be treated as a SHA mismatch.

## Actual behavior

See reproduction. Under gh flakiness, the script either silently merges under a false "CI green" premise or aborts with a confusing tampering message.

## Environment

- macOS / Linux, bash 5.x
- `gh` CLI 2.x
- Triggered any time GitHub is partially degraded or `GH_TOKEN` expires mid-run

## Logs / screenshots

See audit detail in `swe-workbench:auditor` + `swe-workbench:debugger` reports below.

---

### Finding 1

| Field | Value |
|-------|-------|
| **title** | `gh pr checks` errors silently coerced to "0 pending, 0 failed" → merge attempt |
| **severity** | High |
| **domain** | reliability |
| **file_line** | `scripts/release.sh:178-183` |
| **symptom** | `PENDING=$(gh pr checks ... 2>/dev/null \|\| echo "0")` and parallel `FAILED` swallow gh failures (auth expiry, transient 5xx, network blip) and treat them as "all green". |
| **root_cause** | `\|\| echo "0"` conflates "API succeeded, zero pending" with "API failed, status unknown" — the loop cannot distinguish "ready to merge" from "I can't tell". Compounded by gh's documented `rc=8` for failed checks being collapsed into the same bucket. |
| **reasoning_chain** | 1. Line 178 sends gh stderr to `/dev/null` and falls back to `"0"`. 2. Line 181 `if [[ "$PENDING" -eq 0 ]]` evaluates true. 3. Line 194 prints "All CI checks passed". 4. Inner mergeability poll either silently proceeds (if gh recovers and CI is green) or aborts at 120s with wrong diagnosis. |
| **counter_evidence_considered** | Could `set -e` catch this? No — `\|\|` short-circuits. Does the mergeability gate at line 198 protect us? Only partially — limits damage when gh stays broken but allows silent-merge-under-false-premise when gh recovers. |
| **confidence** | 0.85 |
| **effort** | low |
| **suggested_fix** | Collapse the two gh calls into one, capture exit code, distinguish `rc=0` (clean) vs `rc=8` (failures present, JSON valid) vs everything else (re-poll). See debugger diff sketch below. |

### Finding 2

| Field | Value |
|-------|-------|
| **title** | Empty `mergeCommit.oid` after squash-merge aborts release with false mismatch |
| **severity** | High |
| **domain** | reliability |
| **file_line** | `scripts/release.sh:247-254` |
| **symptom** | Immediate `gh pr view --json mergeCommit -q '.mergeCommit.oid'` after `gh pr merge --squash` returns empty due to GitHub's eventually-consistent mergeCommit field. Mismatch check trivially succeeds → script aborts with misleading "Something unexpected was pushed to main". |
| **root_cause** | No retry/wait loop around `mergeCommit.oid` retrieval, no validation that the field is non-empty before comparing. Empty string treated as a valid SHA. |
| **reasoning_chain** | 1. Line 247 fetches `mergeCommit.oid` once. 2. GitHub's `merge_commit_sha` is documented as eventually consistent for squash. 3. Line 245 has just done `git pull --ff-only`, so `LOCAL_SHA` is correct. 4. If `MERGE_SHA=""`, `"$LOCAL_SHA" != ""` is always true → exit 1. 5. Misleading error; re-run picks up clean state but only after human reads false tampering alarm. |
| **counter_evidence_considered** | Does gh retry internally? No — `gh pr view` is single-shot GraphQL. Does the merge-state wait at 195-224 wait on `mergeCommit`? No — only on `mergeStateStatus`. Could a real divergence exist (someone landed a commit between merge and pull)? Yes — fix must distinguish "missing" from "differing", not just wait longer. |
| **confidence** | 0.75 |
| **effort** | low |
| **suggested_fix** | Bounded 60s poll (3s tick) for non-empty `mergeCommit.oid`; preserve original divergence check verbatim once a value exists. See debugger diff sketch below. |

---

### Fix for Finding 1 (lines 178-194)

```diff
-    PENDING=$(gh pr checks "$PR_NUM" --json state \
-      --jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length' 2>/dev/null || echo "0")
-
-    if [[ "$PENDING" -eq 0 ]]; then
-      FAILED=$(gh pr checks "$PR_NUM" --json conclusion \
-        --jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length' 2>/dev/null || echo "0")
+    CHECKS_JSON=$(gh pr checks "$PR_NUM" --json state,conclusion 2>/dev/null) || CHECKS_RC=$?
+    CHECKS_RC=${CHECKS_RC:-0}
+    # 0 = all pass, 8 = at least one failed (JSON still valid).
+    # Anything else is a transient — re-poll without advancing.
+    if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
+      echo "[$(date '+%H:%M:%S')] gh pr checks transient error (rc=${CHECKS_RC}); retrying..."
+      sleep 10; ELAPSED=$((ELAPSED + 10)); unset CHECKS_RC; continue
+    fi
+    unset CHECKS_RC
+    PENDING=$(echo "$CHECKS_JSON" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
+    if [[ "$PENDING" -eq 0 ]]; then
+      FAILED=$(echo "$CHECKS_JSON" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
```

### Fix for Finding 2 (lines 247-254)

```diff
-MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid')
-LOCAL_SHA=$(git rev-parse HEAD)
+# mergeCommit.oid is eventually consistent after `gh pr merge`; poll briefly.
+MERGE_SHA=""
+MERGE_VIEW_ELAPSED=0
+while [[ -z "$MERGE_SHA" && $MERGE_VIEW_ELAPSED -lt 60 ]]; do
+  MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
+  [[ -n "$MERGE_SHA" ]] && break
+  sleep 3; MERGE_VIEW_ELAPSED=$((MERGE_VIEW_ELAPSED + 3))
+done
+if [[ -z "$MERGE_SHA" ]]; then
+  echo "Error: gh did not surface mergeCommit.oid for PR #${PR_NUM} within 60s." >&2
+  echo "Re-run the script — it is resume-safe (PR is already merged)." >&2
+  exit 1
+fi
+LOCAL_SHA=$(git rev-parse HEAD)
```

## Test plan

`scripts/release.sh` has no test harness today. Suggested regression coverage:

- Add `tests/release/test_ci_wait.bats` with a PATH-shimmed `gh` stub returning empty/transient/failed responses.
- Symmetric tests: stub returns valid JSON containing a `FAILURE` conclusion → script hits FAILED-exit branch; stub returns mismatching SHA → original divergence message still fires.

lugassawan:
## Summary
- **Bug A**: Replace dual `gh pr checks` calls guarded by `|| echo "0"` with a single call that captures the raw exit code (`set +e`/`set -e` bracket), then branches: rc=0 or rc=8 (gh-documented "checks failed, JSON still valid") → parse JSON; any other rc → log heartbeat + sleep 10 + `continue` (transient retry). Prevents auth/network failures from silently masking as "0 pending" and causing false-green merges or misleading mergeability errors.
- **Bug B**: Replace single-shot `gh pr view … mergeCommit.oid` with a 60s bounded poll (3s tick). GitHub's GraphQL is eventually consistent after a squash merge; the original single-shot read returned `null`, making `MERGE_SHA=""` and triggering a false "Something unexpected was pushed to main" abort — leaving the PR merged, version bumped, and no tag pushed (release.yml never fires). The poll exits with an idempotent-rerun message on timeout.
- **Tests**: `tests/test_release_sh_ci_wait.py` — 9 regression tests covering static (no `|| echo "0"` fallback, single `gh pr checks` invocation, `mergeCommit` inside a poll loop) and dynamic (guard body routes rc=0/SUCCESS, rc=0/IN_PROGRESS, rc=8/FAILURE, rc=1/transient; poll exits 0 with SHA on call-3 success and exits 1 with idempotent-rerun message on persistent empty).

## Test Plan
- [x] `shellcheck --severity=warning scripts/release.sh` — clean
- [x] `bash -n scripts/release.sh` — parses cleanly
- [x] `pytest tests/test_release_sh_ci_wait.py -v` — all 9 scenarios pass
- [x] `pytest tests/test_release_sh_gh_repo.py -v` — existing tests still pass (no regression)
- [x] `pytest tests/ -x` — full suite 565 passed

- [x] Reproduce the original false-"tamper" abort path: confirmed `MERGE_SHA=""` triggers the timeout branch (not the divergence abort)
- [x] rc=8 with FAILURE JSON routes to the existing CI-failed exit block (not transient retry)
- [x] rc=1 (network error) logs a heartbeat and retries, does not abort or silently proceed

Closes #228

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -189,6 +189,8 @@ else
   LAST_HEARTBEAT=0
 
   while true; do
+    TOTAL=0; PENDING=0; FAILED=0
+
     if [[ $ELAPSED -ge $TIMEOUT ]]; then
       echo "Error: timed out waiting for CI on PR #${PR_NUM} after $((TIMEOUT / 60)) minutes." >&2
       echo "Check status at: ${PR_URL}" >&2
@@ -205,6 +207,13 @@ else
 
     if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
       echo "[$(date '+%H:%M:%S')] gh pr checks transient failure (rc=${CHECKS_RC}); retrying in 10s..."
+      sleep 10
+      ELAPSED=$((ELAPSED + 10))
+      continue
+    fi
+
+    if [[ $CHECKS_RC -eq 8 && -z "$CHECKS_JSON" ]]; then
+      echo "[$(date '+%H:%M:%S')] gh pr checks rc=8 but no output; retrying in 10s..."
       sleep 10
       ELAPSED=$((ELAPSED + 10))
       continue
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -198,13 +198,22 @@ else
       exit 1
     fi
 
-    PENDING=$(gh pr checks "$PR_NUM" --json state \
-      --jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length' 2>/dev/null || echo "0")
+    set +e
+    CHECKS_JSON=$(gh pr checks "$PR_NUM" --json state,conclusion 2>/dev/null)
+    CHECKS_RC=$?
+    set -e
+
+    if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
+      echo "[$(date '+%H:%M:%S')] gh pr checks transient failure (rc=${CHECKS_RC}); retrying in 10s..."
+      sleep 10
+      ELAPSED=$((ELAPSED + 10))
+      continue
+    fi
+
+    PENDING=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
+    FAILED=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
 
     if [[ "$PENDING" -eq 0 ]]; then
-      FAILED=$(gh pr checks "$PR_NUM" --json conclusion \
-        --jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length' 2>/dev/null || echo "0")
-
       if [[ "$FAILED" -gt 0 ]]; then
         echo "Error: ${FAILED} CI check(s) failed on PR #${PR_NUM}." >&2
         echo "Fix the failures at: ${PR_URL}" >&2
@@ -267,7 +276,22 @@ fi
 git checkout main
 git pull --ff-only origin main
 
-MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid')
+MERGE_SHA=""
+POLL_TIMEOUT=60
+POLL_ELAPSED=0
+while [[ $POLL_ELAPSED -lt $POLL_TIMEOUT ]]; do
+  MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
+  [[ -n "$MERGE_SHA" ]] && break
+  sleep 3
+  POLL_ELAPSED=$((POLL_ELAPSED + 3))
+done
+
+if [[ -z "$MERGE_SHA" ]]; then
+  echo "Error: GitHub did not return mergeCommit.oid for PR #${PR_NUM} within ${POLL_TIMEOUT}s." >&2
+  echo "PR is merged; tagging skipped. Re-run this script — it is safe and idempotent." >&2
+  exit 1
+fi
+
 LOCAL_SHA=$(git rev-parse HEAD)
 
 if [[ "$LOCAL_SHA" != "$MERGE_SHA" ]]; then
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -210,10 +210,13 @@ else
       continue
     fi
 
+    TOTAL=$(printf '%s' "${CHECKS_JSON:-[]}" | jq 'length')
     PENDING=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
     FAILED=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
 
-    if [[ "$PENDING" -eq 0 ]]; then
+    if [[ "$TOTAL" -eq 0 ]]; then
+      : # No checks registered yet — treat as pending and continue polling
+    elif [[ "$PENDING" -eq 0 ]]; then
       if [[ "$FAILED" -gt 0 ]]; then
         echo "Error: ${FAILED} CI check(s) failed on PR #${PR_NUM}." >&2
         echo "Fix the failures at: ${PR_URL}" >&2
@@ -282,6 +285,7 @@ POLL_ELAPSED=0
 while [[ $POLL_ELAPSED -lt $POLL_TIMEOUT ]]; do
   MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
   [[ -n "$MERGE_SHA" ]] && break
+  echo "[$(date '+%H:%M:%S')] mergeCommit.oid not yet available (${POLL_ELAPSED}s elapsed); retrying..."
   sleep 3
   POLL_ELAPSED=$((POLL_ELAPSED + 3))
 done
--- a/tests/test_release_sh_ci_wait.py
+++ b/tests/test_release_sh_ci_wait.py
@@ -40,6 +40,7 @@ def _is_comment(line: str) -> bool:
 #
 _GUARD_SNIPPET = textwrap.dedent("""\
     set -euo pipefail
+    TOTAL=0; PENDING=0; FAILED=0
     set +e
     CHECKS_JSON=$(gh pr checks "42" --json state,conclusion 2>/dev/null)
     CHECKS_RC=$?
@@ -47,6 +48,11 @@ _GUARD_SNIPPET = textwrap.dedent("""\
 
     if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
       echo "transient: rc=${CHECKS_RC}" >&2
+      exit 3
+    fi
+
+    if [[ $CHECKS_RC -eq 8 && -z "$CHECKS_JSON" ]]; then
+      echo "transient: rc=8 no output" >&2
       exit 3
     fi
 
@@ -185,6 +191,13 @@ class TestBugADynamic:
         """rc=0, empty array [] → TOTAL=0 → exit 2 (still polling, not a green merge)."""
         self._make_gh_stub(tmp_path, 'printf "[]"; exit 0')
         assert self._run(tmp_path).returncode == 2
+
+    def test_rc8_empty_output_treated_as_transient(self, tmp_path):
+        """rc=8 with no stdout → treated as transient (exit 3), not silent-pending."""
+        self._make_gh_stub(tmp_path, 'exit 8')
+        result = self._run(tmp_path)
+        assert result.returncode == 3
+        assert "transient" in result.stderr
 
 
 # ─── Bug B: static tests ──────────────────────────────────────────────────────
--- a/tests/test_release_sh_ci_wait.py
+++ b/tests/test_release_sh_ci_wait.py
@@ -1,0 +1,250 @@
+"""
+Regression tests for gh pr checks reliability in scripts/release.sh.
+
+Bug A: The dual gh pr checks calls with || echo "0" collapse three distinct signals
+       (rc=0 clean, rc=8 checks-failed-but-JSON-valid, transient auth/network) into
+       a single "0 pending" value — masking failures and causing false green merges.
+
+Bug B: GitHub's GraphQL mergeCommit.oid is eventually consistent after a squash merge.
+       The original single-shot read can return null, making MERGE_SHA="" and triggering
+       a false "tamper" abort — PR merged, version bumped, no tag pushed.
+"""
+import re
+import subprocess
+import tempfile
+import textwrap
+from pathlib import Path
+
+import pytest
+
+from conftest import _CLEAN_ENV
+
+RELEASE_SH = Path(__file__).parent.parent / "scripts" / "release.sh"
+
+
+def _script_lines() -> list[str]:
+    return RELEASE_SH.read_text().splitlines()
+
+
+def _is_comment(line: str) -> bool:
+    return line.lstrip().startswith("#")
+
+
+# ─── Bug A: guard-body snippet ───────────────────────────────────────────────
+#
+# Self-contained single-iteration extract of the CI-wait loop's check logic.
+# Maps routing outcomes to distinct exit codes so tests can assert the branch:
+#   0 = all green  (caller proceeds to merge)
+#   1 = CI failed  (caller aborts with error)
+#   2 = pending    (caller continues polling)
+#   3 = transient  (caller sleeps + retries — rc was not 0 or 8)
+#
+_GUARD_SNIPPET = textwrap.dedent("""\
+    set -euo pipefail
+    set +e
+    CHECKS_JSON=$(gh pr checks "42" --json state,conclusion 2>/dev/null)
+    CHECKS_RC=$?
+    set -e
+
+    if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
+      echo "transient: rc=${CHECKS_RC}" >&2
+      exit 3
+    fi
+
+    PENDING=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
+    FAILED=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
+
+    if [[ "$PENDING" -eq 0 ]]; then
+      if [[ "$FAILED" -gt 0 ]]; then
+        echo "CI check(s) failed" >&2
+        exit 1
+      fi
+      exit 0
+    fi
+    exit 2
+""")
+
+# ─── Bug B: poll-body snippet ─────────────────────────────────────────────────
+#
+# POLL_TIMEOUT injected via env so tests use a short window; sleep is stubbed.
+# Matches the bounded-poll structure introduced in the fix.
+#
+_POLL_SNIPPET = textwrap.dedent("""\
+    set -euo pipefail
+    MERGE_SHA=""
+    POLL_TIMEOUT=${POLL_TIMEOUT:-60}
+    POLL_ELAPSED=0
+    while [[ $POLL_ELAPSED -lt $POLL_TIMEOUT ]]; do
+      MERGE_SHA=$(gh pr view "42" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
+      [[ -n "$MERGE_SHA" ]] && break
+      sleep 3
+      POLL_ELAPSED=$((POLL_ELAPSED + 3))
+    done
+
+    if [[ -z "$MERGE_SHA" ]]; then
+      echo "Error: GitHub did not return mergeCommit.oid for PR #42 within ${POLL_TIMEOUT}s." >&2
+      echo "PR is merged; tagging skipped. Re-run this script — it is safe and idempotent." >&2
+      exit 1
+    fi
+
+    printf '%s\\n' "$MERGE_SHA"
+""")
+
+
+# ─── Bug A: static tests ──────────────────────────────────────────────────────
+
+
+class TestBugAStatic:
+    """Static: the dual-call + || echo "0" fallback pattern must be gone."""
+
+    def test_no_fallback_echo_zero(self):
+        for line in _script_lines():
+            if _is_comment(line):
+                continue
+            assert not re.search(r'gh pr checks.*\|\|.*echo.*"0"', line), (
+                f"Forbidden fallback still present: {line.strip()!r}"
+            )
+
+    def test_gh_pr_checks_called_exactly_once(self):
+        # Exclude echo/printf lines — those are log messages, not invocations.
+        calls = [
+            ln for ln in _script_lines()
+            if "gh pr checks" in ln
+            and not _is_comment(ln)
+            and not ln.lstrip().startswith(("echo ", "printf "))
+        ]
+        assert len(calls) == 1, (
+            f"Expected exactly 1 non-comment 'gh pr checks' invocation, found {len(calls)}: {calls}"
+        )
+
+
+# ─── Bug A: dynamic tests ─────────────────────────────────────────────────────
+
+
+class TestBugADynamic:
+    """Dynamic: guard body correctly routes all four rc/JSON combinations."""
+
+    def _run(self, stub_dir: Path) -> subprocess.CompletedProcess:
+        env = {**_CLEAN_ENV, "PATH": f"{stub_dir}:{_CLEAN_ENV.get('PATH', '')}"}
+        return subprocess.run(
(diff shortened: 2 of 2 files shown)


scripts/release.sh
```diff
@@ -198,13 +198,22 @@ else
...
+      continue
+    fi
 
-    if [[ "$PENDING" -eq 0 ]]; then
-      FAILED=$(gh pr checks "$PR_NUM" --json conclusion \
-        --jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length' 2>/dev/null || echo "0")
+    PENDING=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
+    FAILED=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
 
+    if [[ "$PENDING" -eq 0 ]]; then
```
lugassawan: **High** — rc=0 with an empty JSON array (`[]`) is treated as "all checks passed" and triggers an immediate merge.

When `gh pr checks` returns rc=0 but no checks have registered yet (freshly-created PR, CI workflow not yet triggered), `CHECKS_JSON=[]`. Both `PENDING` and `FAILED` resolve to `0`. The outer `if [[ "$PENDING" -eq 0 ]]` is true, `FAILED=0`, so the script proceeds to merge — with no CI actually run.

Suggested fix — add a zero-checks guard:

```bash
TOTAL=$(printf '%s' "${CHECKS_JSON:-[]}" | jq 'length')
if [[ "$TOTAL" -eq 0 ]]; then
  : # no checks registered yet — treat as pending, continue loop
elif [[ "$PENDING" -eq 0 ]]; then
  if [[ "$FAILED" -gt 0 ]]; then
    ...
  fi
  # proceed to merge
fi
```

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,250 @@
...
+        assert r2.returncode == 0, f"Expected 0 (recovered) on call 2, got {r2.returncode}"
+
+
+# ─── Bug B: static tests ──────────────────────────────────────────────────────
+
+
+class TestBugBStatic:
+    """Static: mergeCommit retrieval is inside a bounded poll loop (while + sleep)."""
+
+    def test_mergecommit_in_poll_loop(self):
```
lugassawan, line 209: **Medium** — The ±15-line proximity window can yield a false-pass if the `gh pr view mergeCommit` call is moved outside the bounded poll but lands near another loop.

The test searches for the first non-comment line matching `"mergeCommit"` and `"gh pr view"`, then checks `lines[max(0, i-5) : i+10]` for `"while"` and `"sleep"`. This passes as long as a `while` and `sleep` exist _anywhere nearby_ — including the mergeability-check inner loop above (which also has both). If the `mergeCommit` call were accidentally hoisted above the bounded poll, the test would still green.

A stronger guard: assert the matched line index falls between the indices of the `while [[ $POLL_ELAPSED` opening and the `done` that closes it (the first such pair after `git pull --ff-only`).

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def test_no_fallback_echo_zero(self):
+        for line in _script_lines():
+            if _is_comment(line):
+                continue
+            assert not re.search(r'gh pr checks.*\|\|.*echo.*"0"', line), (
+                f"Forbidden fallback still present: {line.strip()!r}"
+            )
+
+    def test_gh_pr_checks_called_exactly_once(self):
```
lugassawan, line 116: **Medium** — The `not ln.lstrip().startswith(("echo ", "printf "))` filter is a fragile heuristic.

It correctly excludes today's log line, but would silently miss a future double-invocation wrapped in a heredoc or assigned via `printf … | bash`. A regression that bypasses the filter would show `len(calls) == 1` and pass.

A more robust check: look for lines where `gh pr checks` appears as part of a command substitution or direct invocation (e.g., `$(gh pr checks` or `^gh pr checks`), rather than filtering on leading tokens.

scripts/release.sh
```diff
@@ -267,7 +276,22 @@ fi
...
 git pull --ff-only origin main
 
-MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid')
+MERGE_SHA=""
+POLL_TIMEOUT=60
+POLL_ELAPSED=0
+while [[ $POLL_ELAPSED -lt $POLL_TIMEOUT ]]; do
+  MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
+  [[ -n "$MERGE_SHA" ]] && break
+  sleep 3
```
lugassawan, line 298: **Medium** — The `mergeCommit.oid` poll can silently retry for up to 60 seconds with no output, making it look like a hang in CI logs.

The existing CI-wait loop has a heartbeat on line 265. Suggest adding a one-liner here:

```bash
  echo "[$(date '+%H:%M:%S')] mergeCommit.oid not yet available (${POLL_ELAPSED}s elapsed); retrying..."
```

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,250 @@
...
+    fi
+    exit 2
+""")
+
+# ─── Bug B: poll-body snippet ─────────────────────────────────────────────────
+#
+# POLL_TIMEOUT injected via env so tests use a short window; sleep is stubbed.
+# Matches the bounded-poll structure introduced in the fix.
+#
+_POLL_SNIPPET = textwrap.dedent("""\
```
lugassawan, line 80: **Low** — `_POLL_SNIPPET` duplicates the production shell logic verbatim ("mirror-implementation" antipattern).

If the poll interval, variable names, or timeout in `release.sh` change, the dynamic tests continue passing against the stale copy while the actual script diverges. The static test in `TestBugBStatic` already guards the structural invariant; for the dynamic tests it would be stronger to extract and `exec` the relevant lines from `release.sh` at runtime, similar to how `TestExtractionRoundTrip` in `test_release_sh_gh_repo.py` lifts the `_SNIPPET` from the actual script text.

scripts/release.sh
```diff
@@ -267,7 +276,22 @@ fi
 git checkout main
 git pull --ff-only origin main
 
-MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid')
+MERGE_SHA=""
+POLL_TIMEOUT=60
+POLL_ELAPSED=0
+while [[ $POLL_ELAPSED -lt $POLL_TIMEOUT ]]; do
+  MERGE_SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q '.mergeCommit.oid' 2>/dev/null || true)
```
lugassawan, line 295: **Low** — `2>/dev/null || true` silently discards actionable `gh` error output.

If `gh pr view` exits non-zero for a reason other than "no OID yet" (invalid PR number, expired auth mid-poll), the loop retries 60s then emits the generic "GitHub did not return mergeCommit.oid" message rather than the actionable `gh` error. Capturing stderr and re-emitting it on timeout would make auth failures immediately diagnosable.

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,250 @@
...
+       (rc=0 clean, rc=8 checks-failed-but-JSON-valid, transient auth/network) into
+       a single "0 pending" value — masking failures and causing false green merges.
+
+Bug B: GitHub's GraphQL mergeCommit.oid is eventually consistent after a squash merge.
+       The original single-shot read can return null, making MERGE_SHA="" and triggering
+       a false "tamper" abort — PR merged, version bumped, no tag pushed.
+"""
+import re
+import subprocess
+import tempfile
```
lugassawan: **Low** — `import tempfile` is unused. All temporary directory needs are satisfied by pytest's `tmp_path` fixture. Safe to remove.

scripts/release.sh
lugassawan: Addressed in c37461650183feb8e37e1a401717d06a2b2c1e4e: added `TOTAL=0` guard — when `CHECKS_JSON=[]` (no checks registered yet), the loop skips the merge gate and continues polling.

tests/test_release_sh_ci_wait.py
lugassawan: Addressed in c37461650183feb8e37e1a401717d06a2b2c1e4e: `test_mergecommit_in_poll_loop` now locates the `while [[ $POLL_ELAPSED` and `done` boundaries after `git pull --ff-only` and asserts the retrieval line falls strictly between them — no more proximity window.

tests/test_release_sh_ci_wait.py
lugassawan: Addressed in c37461650183feb8e37e1a401717d06a2b2c1e4e: replaced the `startswith(echo/printf)` filter with a regex that matches `$(gh pr checks` or line-leading `gh pr checks` — catches invocations regardless of how they're wrapped.

scripts/release.sh
lugassawan: Addressed in c37461650183feb8e37e1a401717d06a2b2c1e4e: added a timestamped heartbeat line (matching the CI-wait loop pattern) that prints on each retry.

tests/test_release_sh_ci_wait.py
lugassawan: Addressed in c37461650183feb8e37e1a401717d06a2b2c1e4e: removed unused `import tempfile`.

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,267 @@
...
+
+    if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
+      echo "transient: rc=${CHECKS_RC}" >&2
+      exit 3
+    fi
+
+    PENDING=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.state == "PENDING" or .state == "IN_PROGRESS" or .state == "QUEUED" or .state == "WAITING")] | length')
+    FAILED=$(printf '%s' "${CHECKS_JSON:-[]}" | jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length')
+
+    if [[ "$PENDING" -eq 0 ]]; then
```
lugassawan: **Medium** — `_GUARD_SNIPPET` does not include the `TOTAL` guard added in the follow-up commit, so the new "no checks registered yet → treat as pending" branch has no dynamic coverage.

If the `TOTAL=0` guard in `scripts/release.sh` were accidentally deleted or inverted, all 4 existing dynamic tests would still pass.

Suggested additions — update `_GUARD_SNIPPET` to include `TOTAL`:

```bash
TOTAL=$(printf '%s' "${CHECKS_JSON:-[]}" | jq 'length')
PENDING=$(...)
FAILED=$(...)

if [[ "$TOTAL" -eq 0 ]]; then
  exit 2
elif [[ "$PENDING" -eq 0 ]]; then
  ...
fi
exit 2
```

And add a test case:

```python
def test_empty_checks_array_treated_as_pending(self, tmp_path):
    """rc=0, empty array [] → TOTAL=0 → exit 2 (still polling)."""
    self._make_gh_stub(tmp_path, 'printf "[]"; exit 0')
    assert self._run(tmp_path).returncode == 2
```

tests/test_release_sh_ci_wait.py
```diff
@@ -0,0 +1,267 @@
...
+
+class TestBugBStatic:
+    """Static: mergeCommit retrieval is inside a bounded poll loop (while + sleep)."""
+
+    def test_mergecommit_in_poll_loop(self):
+        lines = _script_lines()
+
+        # Find the POLL_ELAPSED while-loop boundaries (the first one after git pull)
+        pull_idx = next(
+            (i for i, ln in enumerate(lines) if "git pull --ff-only" in ln), None
```
lugassawan: **Low** — `pull_idx` matches the first `"git pull --ff-only"` in `release.sh`, which is a rebase-sync pull near line 61, not the `git pull --ff-only origin main` at line 280 that precedes the POLL_ELAPSED loop.

Today this is harmless because the POLL_ELAPSED `while` is the only loop matching `r'while\s+\[\[.*POLL_ELAPSED'` anywhere in the file. But if a second bounded poll were added between lines 61–280, `while_idx` could anchor to the wrong loop.

Consider searching for `"git pull --ff-only origin main"` (the more specific string that appears only at the two canonical sync points) and taking the last match, or scanning backward from `while_idx` for the nearest preceding `git pull`.

tests/test_release_sh_ci_wait.py
lugassawan: Addressed in e241327: `_GUARD_SNIPPET` now includes `TOTAL` variable and the `if [[ "$TOTAL" -eq 0 ]]; then exit 2` branch, mirroring the production structure exactly. A new test `test_empty_checks_array_treated_as_pending` covers the empty-array → TOTAL=0 → pending-exit path.

tests/test_release_sh_ci_wait.py
lugassawan: Addressed in e241327: `pull_idx` now scans all occurrences and keeps the last one (post-merge sync at ~line 280), not the first (initial rebase-sync at ~line 61). Changed from `next()` to an explicit loop that overwrites `pull_idx` on each match.

scripts/release.sh
```diff
@@ -198,13 +198,25 @@ else
...
+    set +e
+    CHECKS_JSON=$(gh pr checks "$PR_NUM" --json state,conclusion 2>/dev/null)
+    CHECKS_RC=$?
+    set -e
+
+    if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
+      echo "[$(date '+%H:%M:%S')] gh pr checks transient failure (rc=${CHECKS_RC}); retrying in 10s..."
+      sleep 10
+      ELAPSED=$((ELAPSED + 10))
+      continue
```
lugassawan, line 212: **Critical** — `PENDING`, `TOTAL`, and `FAILED` are unbound variables under `set -u` when this `continue` fires.

This `continue` skips the assignments at lines 213–215. The heartbeat log below the main block expands `${PENDING}` — with `set -euo pipefail` active this produces `bash: PENDING: unbound variable` and kills the script on the first transient retry, precisely the case this fix is meant to handle gracefully.

**Fix:** Initialise all three to zero at the top of the `while true` loop, before any `continue`:
```bash
while true; do
  TOTAL=0; PENDING=0; FAILED=0  # reset each iteration; overwritten after a clean gh call
  if [[$ELAPSED -ge $TIMEOUT ]]; then
```

lugassawan, line 212: **Critical** — `PENDING`, `TOTAL`, and `FAILED` are unbound variables under `set -u` when this `continue` fires.

This `continue` skips the assignments at lines 213–215. The heartbeat log (after the loop body) expands `${PENDING}` — with `set -euo pipefail` active this raises `bash: PENDING: unbound variable` and kills the script on the first transient retry, precisely the case this fix is meant to handle gracefully.

**Fix:** Initialise all three at the top of the `while true` loop before any `continue`:
```bash
while true; do
  TOTAL=0; PENDING=0; FAILED=0  # reset each iteration; overwritten after a clean gh call
  if [[ $ELAPSED -ge $TIMEOUT ]]; then
```

scripts/release.sh
```diff
@@ -198,13 +198,25 @@ else
...
+    fi
 
-    if [[ "$PENDING" -eq 0 ]]; then
-      FAILED=$(gh pr checks "$PR_NUM" --json conclusion \
-        --jq '[.[] | select(.conclusion == "FAILURE" or .conclusion == "CANCELLED" or .conclusion == "TIMED_OUT")] | length' 2>/dev/null || echo "0")
+    TOTAL=$(printf '%s' "${CHECKS_JSON:-[]}" | jq 'length')
```
lugassawan, line 222: **High** — When `rc=8` and `gh pr checks` emits no stdout, `${CHECKS_JSON:-[]}` substitutes `[]`, `TOTAL=0`, and the loop silently retries as "no checks yet" — re-introducing a weaker version of Bug A.

`gh pr checks` documents `rc=8` as "at least one check failed, JSON still valid." If stdout is empty despite `rc=8` (partial write, signal, unusual `gh` version), the empty-array fallback masks the failure by treating it as "checks not registered yet" and polling forever.

**Fix:** Treat `rc=8` with empty `CHECKS_JSON` as a transient retry rather than falling through to the `TOTAL=0` branch:
```bash
if [[ $CHECKS_RC -ne 0 && $CHECKS_RC -ne 8 ]]; then
  # transient: retry
  ...
elif [[ $CHECKS_RC -eq 8 && -z "$CHECKS_JSON" ]]; then
  echo "[$(date '+%H:%M:%S')] gh pr checks rc=8 but no output; retrying in 10s..."
  sleep 10; ELAPSED=$((ELAPSED + 10)); continue
fi
```

scripts/release.sh
lugassawan: Addressed in e193328: `TOTAL=0; PENDING=0; FAILED=0` now initialized at the top of the `while true` loop before any `continue`, eliminating the unbound-variable crash under `set -u`.

lugassawan: Addressed in e193328: `TOTAL=0; PENDING=0; FAILED=0` now initialized at the top of the `while true` loop before any `continue`, eliminating the unbound-variable crash under `set -u`.

scripts/release.sh
lugassawan: Addressed in e193328: added explicit guard `if [[ $CHECKS_RC -eq 8 && -z "$CHECKS_JSON" ]]` that routes `rc=8` with empty stdout to the transient-retry branch (sleep 10 + continue) instead of falling through to `TOTAL=0` ("no checks yet"). New test `test_rc8_empty_output_treated_as_transient` covers this case.

PR #276, opened 2026-05-17
[chore] Harden workflows against supply-chain risk

### LINKED ISSUE #230: [chore] Harden workflows against supply-chain risk (hash-pin pip, scope token, SHA-pin actions)
## Type

- [x] Dependency / tooling bump
- [x] Repository config (templates, Dependabot, labels)

## Surface

**Files to change:**
- `.github/workflows/pr.yml`
- `.github/workflows/release.yml`
- `.github/workflows/skill-triggers.yml`
- `.github/dependabot.yml`
- `tests/requirements.txt` (replace with `tests/requirements.lock`)

## Description

A cold-start audit (`/swe-workbench:audit-codebase --depth=deep`) surfaced 5 related supply-chain issues. They all stem from the same root cause family: **third-party code installed at CI time without integrity verification**. Bundling them into one PR because they all touch the same 5 files and one mitigation strategy addresses all of them.

### Finding 6 — `pytest==8.3.4` installed without `--require-hashes` (Medium / security / 0.80)

`tests/requirements.txt:1` pins `pytest==8.3.4` by version only. `pip install -r tests/requirements.txt` (called by `.github/workflows/pr.yml:97` and `.github/workflows/skill-triggers.yml:19`) resolves the full transitive closure (`pluggy`, `iniconfig`, `packaging`, …) on every CI run with no integrity check. `.github/dependabot.yml` only watches `github-actions`, so unhashed transitives are not even monitored.

**Root cause:** one-line `requirements.txt` with no transitive lockfile (no `pip-compile --generate-hashes` output) and no Dependabot pip ecosystem entry.

**Counter-evidence considered:** runners are sandboxed and PR-from-fork tokens are read-only by default. That caps blast radius but does not eliminate compute-abuse / artifact-tampering vectors.

### Finding 7 — `jsonschema==4.23.0` installed on tag-trigger with `contents: write` (Medium / security / 0.70, raised from Low)

`.github/workflows/release.yml:29` runs `python3 -m pip install --quiet 'jsonschema==4.23.0'` inside a job that holds `permissions: contents: write` (line 8-9). A poisoned transitive (`attrs`, `jsonschema-specifications`, `referencing`, `rpds-py`) executes at `from jsonschema import Draft7Validator` (line 35) with the release-write `GITHUB_TOKEN` in env — classic supply-chain-to-release-channel pivot (precedent: event-stream, ua-parser-js, pytorch-nightly).

**Counter-evidence considered:** `rpds-py` ships prebuilt wheels for ubuntu+py3.12 (the runner combo), defusing the "Rust-build-on-install" angle — but the plain Python-import path remains live. `--quiet` actively suppresses installer warnings, reducing audit signal.

### Finding A — `release.yml` workflow-level `contents: write` is broader than needed (Medium / security)

Lines 8-9 grant `contents: write` to the whole `release` job, but only the final `gh release create` step needs it. The schema-validation step (the one that imports third-party Python) inherits the same elevated token.

**Why it matters:** Reduces the cost of any single-step compromise from "rewrite releases" to "fail the build."

### Finding B — Several jobs in `pr.yml` have no explicit `permissions:` block (Low / security)

The `pytest:` job in `pr.yml` (lines 90-98) does not declare `permissions:`. Unlike `skill-triggers.yml:12-13`, it inherits the workflow/repo default. Same gap for sibling jobs `validate-pr`, `validate-plugin-files`, `shellcheck`, `markdown-links`.

### Finding C — Floating major-version action tags (Medium / security)

`actions/checkout@v6`, `actions/setup-python@v6`, `lycheeverse/lychee-action@v2`, `ludeeus/action-shellcheck@2.0.0` are tag-pinned, not SHA-pinned. Tags are mutable; an attacker with push to the action repo or its release token can repoint `v6` to a malicious commit. `tj-actions/changed-files` (March 2025) is the canonical precedent.

**Counter-evidence considered:** First-party `actions/*` may be treated as trusted under some threat models. Third-party actions (`lycheeverse`, `ludeeus`) have a stronger argument for SHA pinning.

## Acceptance

One PR addresses all five findings:

- [ ] Add `pip` ecosystem to `.github/dependabot.yml` (single block, `directory: "/tests"`, weekly schedule).
- [ ] Replace `tests/requirements.txt` with a `pip-compile --generate-hashes`-produced `tests/requirements.lock`; switch `pr.yml:97` and `skill-triggers.yml:19` to `pip install --require-hashes -r tests/requirements.lock`.
- [ ] Hash-pin the release-time `jsonschema` install (either via a `release-requirements.lock` or by inlining a minimal Draft-07 validator — the latter eliminates the dep entirely).
- [ ] Drop `--quiet` from `release.yml:29` for auditability.
- [ ] Split `release.yml` into a `validate-schemas` job (`contents: read`) and a `publish-release` job (`contents: write`, `needs: validate-schemas`).
- [ ] Add `permissions: { contents: read }` to every job lacking one in `pr.yml`.
- [ ] SHA-pin third-party actions (`lycheeverse/lychee-action`, `ludeeus/action-shellcheck`) with `# vX.Y.Z` comments so Dependabot can auto-bump.

lugassawan:
## Summary
- Hash-pin `pytest==8.3.4` full transitive closure in `tests/requirements.lock` via `pip-compile --generate-hashes` (F6)
- Hash-pin `jsonschema==4.23.0` full transitive closure in `tests/release-requirements.lock` (F7)
- Switch both CI `pip install` calls to `--require-hashes -r <lockfile>`
- Split `release.yml` into `validate-schemas` (`contents: read`) and `publish-release` (`contents: write`) jobs — limits write privilege to the single step that needs it (FA)
- Add `permissions: contents: read` (+ `pull-requests: read` on `validate-pr`) to all 5 jobs in `pr.yml` (FB)
- SHA-pin `ludeeus/action-shellcheck@00cae500` (2.0.0) and `lycheeverse/lychee-action@8646ba30` (v2.8.0) — first-party `actions/*` left tag-pinned (FC)
- Add `pip` ecosystem entry to `.github/dependabot.yml` targeting `/tests`, weekly, `[chore]` prefix
- Add `scripts/lock-pip-requirements.sh` helper (`--check` mode for CI, `--help` flag, Python 3.12 venv)

## Test Plan
- [x] `pip install --require-hashes -r tests/requirements.lock` succeeds in clean Python 3.12 venv
- [x] `pip install --require-hashes -r tests/release-requirements.lock` succeeds in clean Python 3.12 venv
- [x] `pytest tests/ -v` — 556/556 passed
- [x] Schema validator (locked `jsonschema==4.23.0`) — both `plugin.json` and `marketplace.json` validate OK
- [x] `shellcheck scripts/lock-pip-requirements.sh` — clean
- [x] All 4 modified YAML files parse correctly
- [x] SHA pins verified: `00cae500` = shellcheck 2.0.0 commit; `8646ba30` = lychee-action v2.8.0 commit
- [x] Lockfile regeneration is deterministic (second run byte-identical)

- [x] Both lockfiles install cleanly with `pip install --require-hashes`
- [x] `scripts/lock-pip-requirements.sh` regenerates byte-identical lockfiles on second run
- [x] `release.yml` split: `validate-schemas` gates `publish-release` via `needs:`

Closes #230


scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
```
lugassawan, line 1: **High: `--check` mode is never invoked by any CI workflow**

The `--check` flag in this script provides drift detection (exits 1 if the lockfile is out of date), but no workflow currently calls it. A contributor can commit a stale `tests/requirements.lock` and CI won't catch it — `pip install --require-hashes` only fails if an existing hash is wrong, not if a newly-required transitive dependency is missing.

**Suggested fix:** add a `lockfile-check` job to `pr.yml`:
```yaml
  lockfile-check:
    runs-on: ubuntu-latest
    permissions:
      contents: read
    steps:
      - uses: actions/checkout@v6
      - uses: actions/setup-python@v6
        with:
          python-version: '3.12'
      - name: Verify lockfiles are up to date
        run: bash scripts/lock-pip-requirements.sh --check
```

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))_

.github/dependabot.yml
```diff
@@ -6,3 +6,10 @@ updates:
       interval: weekly
     commit-message:
       prefix: "[chore]"
+
+  - package-ecosystem: pip
+    directory: "/tests"
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: "[chore]"
```
lugassawan, line 15: **Medium: Dependabot pip PRs cannot regenerate `tests/requirements.lock`**

Dependabot will open PRs that bump `pytest` in `tests/requirements.txt` directly. Those PRs will immediately break CI because `pip install --require-hashes -r tests/requirements.lock` still references the old hashes — a human follow-up commit running `scripts/lock-pip-requirements.sh` is required before merge.

Additionally, `tests/release-requirements.in` uses the `.in` extension which Dependabot's pip ecosystem does not recognise, so `jsonschema` will not receive automated bumps at all.

**Suggested options:**
- Rename `tests/release-requirements.in` → `tests/release-requirements.txt` so Dependabot tracks it
- Document in `CONTRIBUTING.md` that Dependabot pip PRs require a manual `scripts/lock-pip-requirements.sh` run
- Add a GitHub Actions post-hook to auto-regenerate lockfiles on Dependabot pip PRs

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))_

.github/workflows/pr.yml
```diff
@@ -8,6 +8,9 @@ on:
 jobs:
   validate-pr:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
lugassawan, line 16: **Medium: No workflow-level `permissions:` floor**

All 5 jobs declare per-job `permissions:` blocks (good), but there is no top-level `permissions:` block. Any future job added without an explicit block will inherit the repo/org default, which may be broader than `contents: read`.

**Suggested fix:** add before `jobs:`:
```yaml
permissions:
  contents: read

jobs:
```

This provides a safe default floor without changing the behaviour of any existing job (per-job blocks override it).

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))_

tests/requirements.lock
```diff
@@ -0,0 +1,22 @@
+#
+# This file is autogenerated by pip-compile with Python 3.12
+# by the following command:
+#
+#    pip-compile --generate-hashes --output-file=tests/requirements.lock tests/requirements.txt
```
lugassawan, line 5: **Low: Lockfile header omits `--resolver=backtracking`**

The embedded `# by the following command:` header shows the invocation without `--resolver=backtracking`, yet the script always passes that flag. A developer following the header command verbatim will produce output without the flag, potentially causing a different resolution order and a false drift alarm from `--check` mode.

**Suggested fix:** this happens because pip-compile omits flags that match its defaults. Confirm that `backtracking` is the default in the installed pip-tools version — if so, the drift risk is cosmetic only. If not, the script should not suppress the flag from the header (pip-compile includes all non-default flags automatically).

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))_

scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,91 @@
...
+  if [[ "$CHECK" -eq 1 ]]; then
+    local tmp
+    tmp=$(mktemp)
+    "$VENV_DIR/bin/pip-compile" \
+      --generate-hashes \
+      --resolver=backtracking \
+      --quiet \
+      --no-header \
+      --output-file "$tmp" \
+      "$REPO_ROOT/$src"
```
lugassawan: **Important: Temp files leaked when `pip-compile` fails in `--check` mode**

With `set -euo pipefail`, if the `pip-compile` invocation (lines 57–63) exits non-zero (network error, resolution failure), bash immediately propagates the error before reaching either `rm -f` cleanup block (lines 73 or 76), leaking all three temp files on every failure.

Fix — add a `trap` at the function scope:

```bash
compile() {
  local src="$1" out="$2"
  local tmp existing_stripped new_stripped
  tmp=$(mktemp); existing_stripped=$(mktemp); new_stripped=$(mktemp)
  trap 'rm -f "$tmp" "$existing_stripped" "$new_stripped"' RETURN
  # ... rest unchanged
}
```

`trap ... RETURN` fires on any exit from the function including error paths.

scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,91 @@
...
+  PYTHON="python3"
+else
+  echo "Error: Python 3.12 is required (matching CI)." >&2
+  echo "Install it (e.g. 'mise use python@3.12') and ensure 'python3.12' is on PATH." >&2
+  exit 1
+fi
+
+# ── Venv + pip-tools ───────────────────────────────────────────
+
+VENV_DIR="/tmp/pip-lock-venv"
```
lugassawan: **Important: Hardcoded `/tmp/pip-lock-venv` creates silent stale-cache risk on reused runners**

GitHub-hosted runners are ephemeral so this is safe there, but self-hosted or locally-reused runners will find the directory already exists — `python -m venv` over an existing venv does **not** reinstall or upgrade `pip-tools`. The `--check` gate then compares the committed lockfile against output from a potentially older resolver version, silently masking drift.

Fix — scope the venv to a temp directory and clean it up:

```bash
VENV_DIR="$(mktemp -d)"
trap 'rm -rf "$VENV_DIR"' EXIT
```

If a persistent venv for local speed is preferred, document explicitly that this assumes an ephemeral CI environment.

scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,91 @@
...
+"$PYTHON" -m venv "$VENV_DIR"
+"$VENV_DIR/bin/pip" install --quiet pip-tools
```
lugassawan: **Important: `pip-tools` itself is installed without hash-pinning**

This is the only step in the supply-chain hardening effort that installs a package without `--require-hashes`. A compromised or yanked `pip-tools` release could produce tampered lockfiles that then pass the `--check` gate — undermining the entire hash-verification story.

Options:
1. **Minimum viable:** add a version pin and a comment accepting the risk:
   ```bash
   "$VENV_DIR/bin/pip" install --quiet "pip-tools==7.4.1"  # version-pinned; hash-pinning omitted (build tool, not runtime dep)
   ```
2. **Fully consistent:** create a `scripts/requirements-pip-tools.lock` with hashes and install via `--require-hashes -r`.

Option 1 is likely sufficient given `pip-tools` is a local/CI build tool, but the choice should be explicit.

scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,93 @@
...
+    # Run from REPO_ROOT so pip-compile records relative paths in the header.
+    (cd "$REPO_ROOT" && "$VENV_DIR/bin/pip-compile" \
+      --generate-hashes \
+      --resolver=backtracking \
+      --quiet \
+      --no-header \
+      --output-file "$tmp" \
+      "$src")
+    # Strip header from the existing lockfile for comparison
+    grep -v '^#' "$REPO_ROOT/$out" | grep -v '^$' > "$existing_stripped" || true
```
lugassawan, line 76: **Critical: Missing lockfile causes silent false-positive in `--check` mode**

When `--check` runs against a workspace where the lockfile does not yet exist (fresh clone before first generation, or after an accidental deletion), `grep ... "$REPO_ROOT/$out"` on this line fails. The `|| true` swallows the error, leaving `existing_stripped` empty. The subsequent `diff` then compares the freshly compiled lockfile against an **empty file** — it will always show a diff and exit 1. Wait, actually the opposite: if `existing_stripped` is empty and `new_stripped` has content, `diff -q` will report a difference and return 1, triggering the "out of date" error message. So the failure is noisy, not silent.

**Correction from reviewer**: on testing, diff between a non-empty `new_stripped` and an empty `existing_stripped` WILL report "Files differ", so `--check` will correctly exit 1 (false "out of date") rather than silently passing. The practical problem is therefore the **error message quality**: instead of a useful "lockfile does not exist — run the script to generate it", the user sees "out of date", which is confusing.

Add an explicit existence guard for a clearer error:

```bash
if [[ ! -f "$REPO_ROOT/$out" ]]; then
  echo "::error::$out does not exist. Run scripts/lock-pip-requirements.sh to generate it." >&2
  return 1
fi
```

scripts/lock-pip-requirements.sh
```diff
@@ -0,0 +1,93 @@
...
+"$VENV_DIR/bin/pip" install --quiet "pip-tools==7.5.3"
+
+# ── Compile ────────────────────────────────────────────────────
+
+compile() {
+  local src="$1" out="$2"
+  local tmp="" existing_stripped="" new_stripped=""
+  if [[ "$CHECK" -eq 1 ]]; then
+    # Set trap before mktemp so partial failures don't leak files.
+    trap 'rm -f ${tmp:+"$tmp"} ${existing_stripped:+"$existing_stripped"} ${new_stripped:+"$new_stripped"}' RETURN
```
lugassawan, line 59: **Important: `trap ... RETURN` set inside a function leaks into the enclosing shell if the script is `source`d**

In bash, a `trap` set inside a function body is not automatically scoped to that function — it overwrites the enclosing shell's RETURN trap and persists after the function returns. If someone runs `. scripts/lock-pip-requirements.sh` (source) instead of `bash scripts/lock-pip-requirements.sh` (subprocess), the second `compile` call's RETURN trap remains active in the caller's shell and will fire on the caller's next `return`.

In practice this script is always invoked as a subprocess (the `set -euo pipefail` header signals subprocess-only intent), so the risk is low. But the latent defect can be removed by explicitly clearing the trap on the clean path:

```bash
    echo "$out is up to date."
    trap - RETURN   # clear so the trap does not bleed into the outer shell
```

Or use a subshell for the entire check body so the trap is scoped to `EXIT` (which in a subshell is equivalent to RETURN, but is isolated):

```bash
compile() {
  (
    local src="$1" out="$2"
    trap 'rm -f ...' EXIT
    ...
  )
}
```

PR #279, opened 2026-05-17
[feat] Add /swe-workbench:doctor preflight diagnostic command

### LINKED ISSUE #238: [feat] Add /swe-workbench:doctor preflight diagnostic command
## Product framing

**Problem.** The plugin has 11+ runtime dependencies (`gh`, `rimba`, `jq`, `superpowers`, git, shell tools) but no way to check them before starting a workflow. Users only discover a missing tool mid-flow — after a worktree is created or a PR draft is in progress — forcing them to abort and diagnose manually.

**Value.** A single `/swe-workbench:doctor` command lets users verify their environment in seconds before starting any workflow, turning silent mid-flow failures into a clear preflight checklist.

**Impact / Effort.** Impact: L (directly prevents the most common new-user friction). Effort: S (read-only probes, no state changes).

---

## Goal

`/swe-workbench:doctor` runs a read-only preflight check of all runtime dependencies and auth status, printing a green/red table so users know exactly what's missing before they start a workflow.

## Motivation

Today, a user who hasn't installed `rimba` or hasn't run `gh auth login` discovers the gap only when `/implement` or `/review` fails mid-execution — sometimes after a worktree has been created. There is no "check your env" command. With 11+ external tools referenced across commands and skills, the blast radius of a missing dep grows with every new feature.

Current runtime deps referenced across the plugin (non-exhaustive):
- `gh` — required by almost every workflow (PR creation, issue filing, review)
- `rimba` — worktree management (`/implement`, `/review`, `/debug`)
- `jq` — JSON processing in hooks and scripts
- `git` — branching, worktree, log operations
- `claude` CLI — version check for compatibility

## Surface

- [x] Slash command

**New file:** `commands/doctor.md`

## Sketch

```
/swe-workbench:doctor
```

Output format (example):

```
swe-workbench preflight check
─────────────────────────────────────
✓  gh         2.45.0    (gh auth: logged in as lugassawan)
✓  git        2.44.0
✓  jq         1.7.1
✗  rimba      not found  — install: cargo install rimba
✓  claude     1.x.x
─────────────────────────────────────
1 dependency missing. Run the above install command, then re-run /doctor.
```

Implementation shape: read-only `which`/`--version` probes per tool; `gh auth status` for auth; print table; exit 0 (non-blocking — user decides whether to proceed).

## Acceptance (optional)

- [ ] `/swe-workbench:doctor` exists and is listed in `docs/catalog.md`
- [ ] Command probes `gh`, `git`, `jq`, `rimba`, `claude` at minimum
- [ ] `gh auth status` result is surfaced (logged-in account or "not logged in")
- [ ] Missing tools print an install hint
- [ ] Command is read-only — no files modified, no PRs created, no worktrees opened
- [ ] Output is human-readable without requiring any other tool (`jq`, etc.)

lugassawan:
## Summary

- Add `commands/doctor.md` — a thin command prompt that runs `scripts/doctor.sh` and prints output verbatim
- Add `scripts/doctor.sh` — read-only bash probe for `gh`, `git`, `jq`, `rimba`, `claude`; appends `gh auth status` to the `gh` row; prints ✓/✗ table; exits 0 always; writes no files
- Add `tests/test_doctor_command.py` (5 tests) and `tests/test_doctor_script.py` (6 tests) covering existence, output contract, missing-tool hints, auth-not-logged-in path, and no-file-write invariant
- Add `/swe-workbench:doctor` row to `docs/catalog.md` Commands table and to the README Commands bullet

## Test Plan

- [x] Examples in the diff were exercised manually — `bash scripts/doctor.sh` renders full ✓ table; `PATH=/usr/bin:/bin bash scripts/doctor.sh` shows ✗ for gh and claude with install hints, exit 0
- [x] `pytest tests/ -v` — 622/622 pass (10 new doctor tests + 1 new auth-not-logged-in test)
- [x] `shellcheck scripts/doctor.sh` — 0 issues
- [x] `bash scripts/validate.sh` — All checks passed

Closes #238


scripts/doctor.sh
```diff
@@ -0,0 +1,82 @@
...
+elif [[ -x "${HOME}/.local/bin/rimba" ]]; then
+  rimba_bin="${HOME}/.local/bin/rimba"
+elif [[ -x "${HOME}/go/bin/rimba" ]]; then
+  rimba_bin="${HOME}/go/bin/rimba"
+fi
+
+if [[ -n "${rimba_bin}" ]]; then
+  # Try --version; cobra CLIs may not support it, so suppress stderr and fall back.
+  rimba_ver=$("${rimba_bin}" --version 2>/dev/null | head -1) || true
+  [[ -z "${rimba_ver}" ]] && rimba_ver=$("${rimba_bin}" version 2>/dev/null | head -1) || true
```
lugassawan: **Medium** — fragile `A && B || C` operator-precedence pattern.

With bash, `[[ -z "${rimba_ver}" ]] && rimba_ver=$(cmd) || true` behaves differently from a proper `if/then`: if the condition is true but the subshell returns non-zero, `|| true` fires and `rimba_ver` is left empty. The next line then sets `rimba_ver="installed (version unknown)"` which is the intended fallback — so it works today. But the pattern is easy to misread and could mislead future maintainers.

Suggested fix: replace lines 61-62 with an explicit `if` block:
```bash
if [[ -z "${rimba_ver}" ]]; then
  rimba_ver=$("${rimba_bin}" version 2>/dev/null | head -1) || true
fi
[[ -z "${rimba_ver}" ]] && rimba_ver="installed (version unknown)"
```

tests/test_doctor_script.py
```diff
@@ -0,0 +1,144 @@
...
+import textwrap
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+DOCTOR_SH = ROOT / "scripts" / "doctor.sh"
+
+from conftest import _CLEAN_ENV
+
+
+def _make_mock_tools(tmp_path: Path, omit: set | None = None) -> Path:
```
lugassawan, line 16: **Medium** — `set | None` is Python 3.10+ union syntax.

If the CI or any user runs the test suite on Python 3.9, this annotation will raise `TypeError: unsupported operand type(s) for |: "type" and "NoneType"` at import time, silently failing the whole test file — not just this test.

The rest of the suite uses no type annotations on function signatures. Suggested fix: either add `from __future__ import annotations` at the top of the file (defers evaluation, works on 3.9), or change the annotation to `Optional[set]` with `from typing import Optional`.

scripts/doctor.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+# Read-only preflight probe for swe-workbench runtime dependencies.
+# Prints a green/red table to stdout. Writes no files. Exits 0 always.
+set -uo pipefail  # -e intentionally omitted: exit 0 is required even when probes fail
```
lugassawan: **Low** — `set -uo pipefail` comment only documents `-e` omission; `-u` is undocumented.

The current comment explains why `-e` is absent but does not mention that `-u` (treat unset variables as errors) is active. A future contributor adding a new probe call who references an uninitialized variable will see an unexplained abort under `-u`.

Suggested addition to the comment:
```bash
set -uo pipefail  # -e intentionally omitted: exit 0 is required even when probes fail
                  # -u: all variables must be initialized before use
```

tests/test_doctor_script.py
```diff
@@ -0,0 +1,144 @@
...
+    assert "missing" in result.stdout.lower(), "Summary must mention missing dependencies"
+
+
+def test_gh_auth_status_surfaced(tmp_path):
+    """Doctor output must include gh auth status annotation on the gh row."""
+    bin_dir = _make_mock_tools(tmp_path)
+    env = {**_CLEAN_ENV, "PATH": f"{bin_dir}:/usr/bin:/bin", "HOME": str(tmp_path)}
+    result = _run_doctor(env)
+    assert result.returncode == 0
+    assert "gh auth" in result.stdout, (
```
lugassawan, line 111: **High** — test does not verify account name is present in the logged-in output.

`assert "gh auth" in result.stdout` passes even if auth-user extraction is broken and shows `(gh auth: not logged in)` instead of `(gh auth: logged in as mockuser)`. The acceptance criterion requires surfacing the actual account name, not just the prefix.

Suggested fix: add a stronger assertion:
```python
assert "logged in as mockuser" in result.stdout, (
    "gh row must show the authenticated username"
)
```

lugassawan: Addressed in d46f7ec: added `assert "logged in as mockuser" in result.stdout` to verify the username is extracted and displayed, not just the `gh auth` prefix.

scripts/doctor.sh
lugassawan: Addressed in d46f7ec: extended the `set` comment to document `-u` semantics alongside the existing `-e` note.

tests/test_doctor_script.py
lugassawan: Addressed in d46f7ec: added `from __future__ import annotations` to defer annotation evaluation, making `set | None` valid on Python 3.9+.

scripts/doctor.sh
lugassawan: Addressed in d46f7ec: replaced fragile `A && B || C` with explicit `if/then` block in rimba version fallback.

tests/test_doctor_script.py
```diff
@@ -0,0 +1,149 @@
...
+    bin_dir.mkdir()
+    omit = omit or set()
+
+    tools = {
+        "gh": textwrap.dedent("""\
+            #!/usr/bin/env bash
+            if [[ "$1" == "--version" ]]; then
+              echo "gh version 2.45.0 (2024-01-01)"
+            elif [[ "$1" == "auth" && "$2" == "status" ]]; then
+              echo "Logged in to github.com account mockuser (keyring)" >&2
```
lugassawan, line 28: **Low** — mock `gh auth status` writes to stderr (`>&2`) but the assertion on `result.stdout` works through `printf` in the script, not through a direct stderr-capture test. If the `2>&1` on `doctor.sh:16` were accidentally changed to `2>/dev/null`, `test_gh_auth_status_surfaced` would still pass while auth parsing silently broke.

Consider adding a comment on the mock noting the `>&2` is intentional and coupled to the `2>&1` redirect in the script, or add a dedicated test that uses a gh mock that *only* writes to stderr and asserts the username still appears in output.

PR #280, opened 2026-05-17
[chore] Auto-regenerate pip lockfiles on Dependabot PRs

lugassawan:
## Summary
- Add `.github/workflows/dependabot-lockfile.yml` — a new CI workflow that automatically regenerates `tests/requirements.lock` and `tests/release-requirements.lock` when Dependabot bumps `tests/requirements.txt` or `tests/release-requirements.txt`
- Without this, Dependabot only updates the `.txt` input file; the hashed `.lock` outputs (managed by `scripts/lock-pip-requirements.sh` via `pip-compile --generate-hashes`) go stale, causing `lockfile-check` to fail on every Dependabot pip PR (root cause of #277 and #278)
- The workflow is scoped to `dependabot[bot]`-authored PRs via `github.event.pull_request.user.login` (not `github.actor`, which changes when a human pushes to the branch)

## Test Plan
- [ ] CI passes on this PR (`validate-pr`, `shellcheck`, `validate-plugin-files`, `lockfile-check`, `pytest`, `Markdown link check`)

### Type-tailored checks (ci)
- [ ] Workflow YAML parses cleanly (confirmed by `validate-plugin-files` / shellcheck)
- [ ] The `if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}` guard uses expression syntax (bare string would always be truthy)
- [ ] `git diff --quiet HEAD -- tests/requirements.lock tests/release-requirements.lock` correctly short-circuits on a no-op run
- [ ] `contents: write` is scoped to the single job; workflow default is `contents: read`
- [ ] No `pull_request_target` trigger (would give untrusted code write access)

Issue: N/A — follow-up to #277 and #278; makes future Dependabot pip bumps self-healing

### DIFF
--- a/.github/workflows/dependabot-lockfile.yml
+++ b/.github/workflows/dependabot-lockfile.yml
@@ -12,6 +12,12 @@ on:
 permissions:
   contents: read
 
+# Cancel in-progress runs when Dependabot pushes a new commit — only the
+# latest lockfile state matters, so stale runs are safe to discard.
+concurrency:
+  group: dependabot-lockfile-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   regen-lockfiles:
     runs-on: ubuntu-latest
--- a/.github/workflows/dependabot-lockfile.yml
+++ b/.github/workflows/dependabot-lockfile.yml
@@ -1,0 +1,50 @@
+name: Dependabot Lockfile Regen
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches: [main]
+    paths:
+      - tests/requirements.txt
+      - tests/release-requirements.txt
+
+# Default to least-privilege; the one job that writes elevates explicitly.
+permissions:
+  contents: read
+
+jobs:
+  regen-lockfiles:
+    runs-on: ubuntu-latest
+    # Only run for Dependabot PRs. Checking the PR author (user.login) rather than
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-lockfile.yml
```diff
@@ -0,0 +1,50 @@
...
+  pull_request:
+    types: [opened, synchronize]
+    branches: [main]
+    paths:
+      - tests/requirements.txt
+      - tests/release-requirements.txt
+
+# Default to least-privilege; the one job that writes elevates explicitly.
+permissions:
+  contents: read
```
lugassawan, line 13: **Missing `concurrency` group — parallel runs can cause a push collision**

If Dependabot pushes two commits in quick succession (e.g. it rebases its own branch), two concurrent `regen-lockfiles` runs can both reach the `git push` step simultaneously. The second push will fail non-fast-forward, leaving the PR red and requiring a manual re-run — the exact failure mode this workflow is meant to eliminate.

Suggested fix — add right after the `permissions:` block at the workflow level:

```yaml
concurrency:
  group: dependabot-lockfile-${{ github.ref }}
  cancel-in-progress: true
```

`cancel-in-progress: true` is safe here: only the latest commit's lockfile state matters, so cancelling a stale run is desirable.

PR #283, opened 2026-05-17
[fix] Suppress stderr in sync-and-verify.sh pull block

lugassawan:
## Summary

- `scripts/sync-and-verify.sh` Block B (the `git checkout && git pull --ff-only` step) redirected only stdout to `/dev/null`, leaving git's fetch progress messages on stderr. A caller using `eval "$(... 2>&1)"` merged that stderr into the captured stdout; bash then eval'd `* branch main -> FETCH_HEAD` as a shell redirect, creating an empty `FETCH_HEAD` file in the working tree.
- Fix: add `2>&1` after `>/dev/null` on the pull line so all git progress output is discarded on the happy path. The `|| echo "sync-main: best-effort failed …" >&2` fallback is preserved.
- Regression test (`TestSyncAndVerifyEvalSafety`): pushes a real commit to a fixture origin, runs the script with merged stderr, eval's the output from a clean temp directory, and asserts no stray `FETCH_HEAD` file is created.

## Test Plan

- [x] New regression test fails against unfixed script (red commit `b2cb6bd`)
- [x] New regression test passes after one-line fix (green commit `1bd36e5`)
- [x] `shellcheck scripts/sync-and-verify.sh` — clean
- [x] Full test suite: 627 tests pass (`pytest tests/`)
- [x] `ruff check` / `ruff format` — no issues on modified files
- [ ] Manual smoke: `eval "$(bash skills/workflow-cleanup-merged/scripts/sync-and-verify.sh main main 2>&1)"` from a temp dir with a decoy file — verify no `FETCH_HEAD` created

Issue: N/A — stray empty FETCH_HEAD file found on main after /cleanup-merged for PR #282; root cause is unsafe eval of git output containing git's fetch tracking line

### DIFF
--- a/skills/language-bash/SKILL.md
+++ b/skills/language-bash/SKILL.md
@@ -117,7 +117,7 @@ cd "$eval_cwd" && eval "$(bash script.sh arg 2>&1)"
 # Correct — capture first, then move, then eval:
 output="$(bash script.sh arg 2>&1)"; cd "$eval_cwd"; eval "$output"
 ```
-  Under `set -e`, check `$?` after the capture step before eval-ing — a non-zero exit from the subshell leaves `$output` empty or partial and `eval "$output"` silently no-ops.
+  Under `set -e`, use `output=$(…) || handle_error` — `$?` is unreachable because the parent script aborts at the failed assignment before the next line executes. The `||` forms a conditional context that suppresses `set -e` and runs the handler on non-zero exit.
 
 ## Avoid
 - Backtick substitution `` `cmd` `` — use `$(cmd)`.
--- a/skills/language-bash/SKILL.md
+++ b/skills/language-bash/SKILL.md
@@ -109,6 +109,14 @@ fi
 assert_eq() { [[ "$1" == "$2" ]] || { echo "FAIL: expected '$2', got '$1'" >&2; exit 1; }; }
 ```
 - Mock external commands by prepending a temp dir containing stub scripts to `PATH`.
+- **eval/cwd trap**: when testing `eval "$(script 2>&1)"` patterns, capture the script output FIRST from a valid cwd, THEN `cd` to the eval directory, THEN eval. `$(...)` launches a subshell that inherits the cwd at expansion time — `cd eval_cwd && eval "$(script)"` means the script runs FROM `eval_cwd`, not the original directory, and may exit early.
+```bash
+# Wrong — script inherits eval_cwd as cwd, may exit early if it's not a git repo:
+cd "$eval_cwd" && eval "$(bash script.sh arg 2>&1)"
+
+# Correct — capture first, then move, then eval:
+output="$(bash script.sh arg 2>&1)"; cd "$eval_cwd"; eval "$output"
+```
 
 ## Avoid
 - Backtick substitution `` `cmd` `` — use `$(cmd)`.
--- a/skills/language-bash/SKILL.md
+++ b/skills/language-bash/SKILL.md
@@ -117,6 +117,7 @@ cd "$eval_cwd" && eval "$(bash script.sh arg 2>&1)"
 # Correct — capture first, then move, then eval:
 output="$(bash script.sh arg 2>&1)"; cd "$eval_cwd"; eval "$output"
 ```
+  Under `set -e`, check `$?` after the capture step before eval-ing — a non-zero exit from the subshell leaves `$output` empty or partial and `eval "$output"` silently no-ops.
 
 ## Avoid
 - Backtick substitution `` `cmd` `` — use `$(cmd)`.
--- a/tests/test_sync_and_verify.py
+++ b/tests/test_sync_and_verify.py
@@ -183,7 +183,9 @@ class TestSyncAndVerifyEvalSafety:
         no_hooks.mkdir(exist_ok=True)  # don't rely on git_repo fixture side-effect
 
         subprocess.run(
-            ["git", "clone", str(tmp_path / "origin.git"), str(clone)],
+            # --branch main: force HEAD checkout regardless of bare repo's default branch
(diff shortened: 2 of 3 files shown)


skills/language-bash/SKILL.md
```diff
@@ -109,6 +109,15 @@ fi
...
 - Mock external commands by prepending a temp dir containing stub scripts to `PATH`.
+- **eval/cwd trap**: when testing `eval "$(script 2>&1)"` patterns, capture the script output FIRST from a valid cwd, THEN `cd` to the eval directory, THEN eval. `$(...)` launches a subshell that inherits the cwd at expansion time — `cd eval_cwd && eval "$(script)"` means the script runs FROM `eval_cwd`, not the original directory, and may exit early.
+```bash
+# Wrong — script inherits eval_cwd as cwd, may exit early if it's not a git repo:
+cd "$eval_cwd" && eval "$(bash script.sh arg 2>&1)"
+
+# Correct — capture first, then move, then eval:
+output="$(bash script.sh arg 2>&1)"; cd "$eval_cwd"; eval "$output"
+```
+  Under `set -e`, check `$?` after the capture step before eval-ing — a non-zero exit from the subshell leaves `$output` empty or partial and `eval "$output"` silently no-ops.
```
lugassawan: **Medium** — The `set -e` caveat is misleading and could produce silently-broken error handling.

Under `set -e`, if `$(bash script.sh arg 2>&1)` exits non-zero, the parent script aborts immediately at the assignment — `$?` is never reachable. A reader following this note might write a dead-code guard:

```bash
output="$(bash script.sh arg 2>&1)"
[ $? -ne 0 ] && handle_error  # dead code under set -e — parent already exited
eval "$output"
```

Correct patterns under `set -e`:

```bash
# Option A — explicit || branch (recommended):
output="$(bash script.sh arg 2>&1)" || { echo 'capture failed' >&2; exit 1; }
eval "$output"

# Option B — temporarily suspend set -e:
set +e; output="$(bash script.sh arg 2>&1)"; rc=$?; set -e
[ $rc -ne 0 ] && { echo 'capture failed' >&2; exit 1; }
eval "$output"
```

Suggest rewriting: _"Under `set -e`, use `output=$(…) || handle_error` — `$?` is unreachable because the parent script aborts at the failed assignment before execution reaches the next line."_

tests/test_sync_and_verify.py
```diff
@@ -167,3 +166,132 @@ def test_branch_absent_trunk_default(self, git_repo_trunk):
...
+        ).stdout.strip()
+        assert origin_sha != local_sha, (
+            f"Pre-condition: origin/main ({origin_sha}) must be ahead of "
+            f"local HEAD ({local_sha}). git pull won't fetch — the "
+            f"'-> FETCH_HEAD' tracking line won't appear and the test is vacuous."
+        )
+
+        # Guard: eval_cwd and SCRIPT paths must not contain double-quotes,
+        # which would break the inline bash string we build below.
+        assert '"' not in str(eval_cwd) and '"' not in str(SCRIPT), (
```
lugassawan: **Low** — Double-quote guard omits `BRANCH`, which is also interpolated into the bash f-string (line 281).

`BRANCH = "feature/197-fixture-branch"` is safe today, but if the constant ever changes (e.g. to test a branch name with spaces or special chars), this guard won't catch it.

```suggestion
        assert '"' not in str(eval_cwd) and '"' not in str(SCRIPT) and '"' not in BRANCH, (
            f"Path or branch contains double-quote — inline bash string will break.\n"
            f"eval_cwd={eval_cwd}, SCRIPT={SCRIPT}, BRANCH={BRANCH}"
        )
```

tests/test_sync_and_verify.py
```diff
@@ -167,3 +166,132 @@ def test_branch_absent_trunk_default(self, git_repo_trunk):
...
+        )
+
+        # Capture BEFORE cd-ing to eval_cwd: the script needs a valid git cwd
+        # to resolve MAIN_REPO. If we cd first, the script exits early.
+        runner = (
+            f'output="$(bash "{SCRIPT}" "{BRANCH}" main 2>&1)"; '
+            f'cd "{eval_cwd}"; '
+            f'eval "$output" 2>/dev/null || true'
+        )
+        subprocess.run(
```
lugassawan: **Low** — Runner `subprocess.run` result is discarded. The `|| true` inside the bash string means bash exits 0 even on eval errors, so `check=True` is safe here and catches real setup failures (wrong path, broken f-string) that would otherwise let the test pass vacuously.

```suggestion
        result = subprocess.run(
            ["bash", "-c", runner],
            cwd=str(git_repo),
            capture_output=True,
            text=True,
            env=_CLEAN_ENV,
        )
        assert result.returncode == 0, (
            f"bash runner failed (rc={result.returncode}):\n{result.stderr}"
        )
```

skills/language-bash/SKILL.md
lugassawan: Addressed in e0d2dcd1264b65e354d0d07840c193408e2aae17: rewrote the caveat to say "use `output=$(…) || handle_error` — `$?` is unreachable under `set -e` because the parent aborts at the failed assignment before the next line executes."

tests/test_sync_and_verify.py
lugassawan: Addressed in e0d2dcd1264b65e354d0d07840c193408e2aae17: added `and '"' not in BRANCH` to the assertion and included `BRANCH` in the error message.

tests/test_sync_and_verify.py
lugassawan: Addressed in e0d2dcd1264b65e354d0d07840c193408e2aae17: captured the result and added `assert result.returncode == 0` before the FETCH_HEAD assertion.

PR #284, opened 2026-05-19
[docs] Add gitchamber shared partial for external-repo reading

lugassawan:
## Summary
- Add `agents/shared/external-repo-reading.md` — a shared partial teaching agents to use https://gitchamber.com when reading files from external GitHub repos (preferred over `WebFetch raw.githubusercontent.com` or `git clone`)
- Documents URL patterns, line-based pagination (`?start`/`?end`), glob filtering (`?glob=<pattern>`, with the requirement that the same glob is used across list/read/search), and the one-time `curl -s https://gitchamber.com` discovery step
- Wire the partial into 4 agents via a `## Reading external repos > See @./shared/external-repo-reading.md.` section each: `architect`, `senior-engineer`, `dependency-auditor`, `contributor-auditor`
- `security-auditor` was intentionally excluded — its `## Read-only enforcement` section forbids outbound `curl`/`wget` calls and it lacks `WebFetch`

## Test Plan
- [ ] `python scripts/validate.py` passes

### Type-tailored checks ([docs])
- [ ] `python scripts/validate.py` confirmed passing (evidence: ran in worktree, "All checks passed")
- [ ] Partial wiring confirmed: `grep -rn 'external-repo-reading' agents/` shows one reference per wired agent
- [ ] URL conventions match live `curl -s https://gitchamber.com` output (verified during implementation)
- [ ] `?glob=` parameter syntax and consistency requirement confirmed against gitchamber docs

Issue: N/A — documentation-only addition of a shared partial; no upstream issue filed

### DIFF
--- a/agents/contributor-auditor.md
+++ b/agents/contributor-auditor.md
@@ -113,7 +113,7 @@ The final line of every report must be:
 
 `Bash` is available for read-only investigation only.
 
-**Allowed:** `gh api`, `gh pr view`, `gh pr diff`, `git log`, `git diff`, `git show`, `grep`, `find`, `ls`.
+**Allowed:** `gh api`, `gh pr view`, `gh pr diff`, `git log`, `git diff`, `git show`, `grep`, `find`, `ls`, outbound reads to `gitchamber.com` (see `## Reading external repos`).
 
 **Forbidden:** any redirect (`>`, `>>`), `rm`, `mv`, `cp`, `git commit`, `git push`, `gh pr comment`, `gh pr review`, `gh pr merge`, or any command that writes to disk, modifies repo state, or posts to the PR.
 
--- a/agents/contributor-auditor.md
+++ b/agents/contributor-auditor.md
@@ -135,6 +135,10 @@ Output is markdown only. Never comment on the PR, apply labels, request changes 
 - Do not block on hygiene notes (Low severity) alone.
 - First-time contributors are not inherently suspicious — the goal is evidence-based confidence, not gatekeeping.
 
+## Reading external repos
+
+> See @./shared/external-repo-reading.md.
+
 ## Principle consultation
 
 > See @./shared/principles.md for the skill catalog.
--- a/agents/shared/external-repo-reading.md
+++ b/agents/shared/external-repo-reading.md
@@ -5,13 +5,13 @@ working repo, prefer **https://gitchamber.com** over fetching raw GitHub
 URLs or shelling out to `git clone`.
 
 Gitchamber URLs are plain HTTPS — use whichever tool your agent has:
-- **`Bash` agents:** use the shell examples below.
+- **`Bash` agents:** pass the URLs to `curl -s` or any HTTP client.
 - **`WebFetch` agents:** pass the same URLs directly to `WebFetch`.
 
 ## URL patterns
 
 ```
-BASE: https://gitchamber.com/repos/{owner}/{repo}/{branch}/
+BASE: https://gitchamber.com/repos/{owner}/{repo}/{branch}
 
 List files:  GET {BASE}/files
 Read file:   GET {BASE}/files/{filepath}?start=N&end=M&showLineNumbers=true
--- a/agents/shared/external-repo-reading.md
+++ b/agents/shared/external-repo-reading.md
@@ -1,12 +1,12 @@
-# Shared external-repo-reading reference
+# Shared external repo reading reference
 
 When you need to read source files from a GitHub repository other than the
-working repo, prefer **https://gitchamber.com** over `WebFetch`-ing raw
-GitHub URLs or shelling out to `git clone`.
+working repo, prefer **https://gitchamber.com** over fetching raw GitHub
+URLs or shelling out to `git clone`.
 
-Discovery: run `curl -s https://gitchamber.com` once per session to see the
-current URL conventions. The endpoint is `curl`-friendly, so agents with
-`Bash` (but not `WebFetch`) can use it without a tool-list change.
+Gitchamber URLs are plain HTTPS — use whichever tool your agent has:
+- **`Bash` agents:** use the shell examples below.
+- **`WebFetch` agents:** pass the same URLs directly to `WebFetch`.
 
 ## URL patterns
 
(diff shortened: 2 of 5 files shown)


agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,46 @@
+# Shared external-repo-reading reference
+
+When you need to read source files from a GitHub repository other than the
+working repo, prefer **https://gitchamber.com** over `WebFetch`-ing raw
+GitHub URLs or shelling out to `git clone`.
+
+Discovery: run `curl -s https://gitchamber.com` once per session to see the
+current URL conventions. The endpoint is `curl`-friendly, so agents with
+`Bash` (but not `WebFetch`) can use it without a tool-list change.
```
lugassawan: **High** — Bash-based instructions are unreachable for `architect` and `senior-engineer`

Both agents declare `Read, Grep, Glob, WebFetch, Skill` in their frontmatter — no `Bash`. This partial's primary workflow requires a shell tool that neither agent has. An agent without `Bash` receiving these instructions will either hallucinate a tool call or silently fall back to its prior behaviour.

The opening paragraph says to prefer gitchamber over fetching raw GitHub URLs, but never provides a `WebFetch`-compatible URL for agents that only have `WebFetch`.

**Suggested fix:** Add a `WebFetch` path alongside the shell-tool path, e.g.:

```
# Agents with Bash: use the shell examples below
# Agents with WebFetch only:
WebFetch: https://gitchamber.com/repos/{owner}/{repo}/{branch}/files
```

Alternatively, remove the `@./shared/external-repo-reading.md` reference from `architect.md` and `senior-engineer.md` until they have a viable execution path.

agents/contributor-auditor.md
```diff
@@ -135,6 +135,10 @@ Output is markdown only. Never comment on the PR, apply labels, request changes
 - Do not block on hygiene notes (Low severity) alone.
 - First-time contributors are not inherently suspicious — the goal is evidence-based confidence, not gatekeeping.
```
lugassawan, line 137: **High** — Shell-based outbound access is now implicitly authorised in `contributor-auditor` without being listed in its `## Read-only enforcement` section

`contributor-auditor`'s enforcement section explicitly lists allowed commands (`gh api`, `gh pr view`, `git log`, …) and forbidden commands. The new partial instructs the agent to make outbound network calls to read external repo files, but neither the `**Allowed:**` nor `**Forbidden:**` list mentions this capability — leaving it in an ambiguous grey zone that expands the agent's effective blast radius beyond what the audit section ever sanctioned.

**Suggested fix:** Add the gitchamber URL pattern to the `**Allowed:**` list in `contributor-auditor.md` if outbound reads are intentional, or explicitly forbid it if not. The current silence is an instruction gap.

agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,46 @@
+# Shared external-repo-reading reference
+
+When you need to read source files from a GitHub repository other than the
+working repo, prefer **https://gitchamber.com** over `WebFetch`-ing raw
+GitHub URLs or shelling out to `git clone`.
+
+Discovery: run `curl -s https://gitchamber.com` once per session to see the
```
lugassawan: **Medium** — Discovery-step runtime fetch is a fragile bootstrap dependency

Asking agents to fetch the root URL once per session to learn URL conventions adds a single point of failure: if the endpoint is temporarily unavailable or changes its response format, every external-repo read breaks silently mid-session.

The URL patterns documented immediately below (lines 13–19 of this partial) are concrete and self-sufficient. Agents should use them directly.

**Suggested fix:** Remove the discovery step or demote it to an "if in doubt" note. If URL conventions are expected to drift, document a versioned URL or a drift policy rather than a per-session fetch.

agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,46 @@
...
+
+```bash
+curl -s "https://gitchamber.com/repos/facebook/react/main/files"
+curl -s "https://gitchamber.com/repos/facebook/react/main/files/README.md?start=1&end=50"
+curl -s "https://gitchamber.com/repos/facebook/react/main/search/useState"
+```
+
+By default gitchamber indexes markdown files and READMEs. To read source
+files (`.ts`, `.py`, etc.), add `?glob=<pattern>` — the same glob must be
+used consistently across all operations (list, read, search) for a given repo.
```
lugassawan, line 31: **Medium** — Glob consistency requirement is under-specified when combined with pagination parameters

The instruction says "the same glob must be used consistently across all operations" but provides no example combining `?glob=` with `?start=`/`?end=`. An agent constructing a paginated read on a glob-filtered set has no canonical form to follow and may produce inconsistent behaviour.

**Suggested fix:** Add a combined example:

```
GET .../files/src/index.ts?glob=**/*.ts&start=1&end=50&showLineNumbers=true
```

agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,46 @@
+# Shared external-repo-reading reference
```
lugassawan: **Low** — Heading uses a hyphenated slug where other partials use plain noun phrases

Other shared partials use: `# Shared agent boundary reference`, `# Shared severity output contract`. This partial's heading contains a hyphen (`external-repo-reading`) which diverges from the established style.

**Suggested fix:** Rename to `# Shared external repo reading reference`.

agents/shared/external-repo-reading.md
lugassawan: Addressed in 4ea6ebeb: added explicit note that gitchamber URLs work with both Bash and WebFetch; examples now shown as plain URLs usable by either tool.

agents/contributor-auditor.md
lugassawan: Addressed in 4ea6ebeb: added outbound gitchamber reads to the Allowed list in contributor-auditor's Read-only enforcement section.

agents/shared/external-repo-reading.md
lugassawan: Addressed in 4ea6ebeb: removed the per-session discovery fetch requirement; demoted to a blockquote footnote for the 'if conventions seem to have changed' case.

agents/shared/external-repo-reading.md
lugassawan: Addressed in 4ea6ebeb: added combined URL example showing ?glob= and ?start=/?end= together; changed examples to plain tool-neutral URLs.

agents/shared/external-repo-reading.md
lugassawan: Addressed in 4ea6ebeb: renamed heading to '# Shared external repo reading reference' (no kebab).

agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,51 @@
...
+URLs or shelling out to `git clone`.
+
+Gitchamber URLs are plain HTTPS — use whichever tool your agent has:
+- **`Bash` agents:** use the shell examples below.
+- **`WebFetch` agents:** pass the same URLs directly to `WebFetch`.
+
+## URL patterns
+
+```
+BASE: https://gitchamber.com/repos/{owner}/{repo}/{branch}/
```
lugassawan: **Low** — BASE URL has a trailing slash that produces double-slash paths when concatenated

The template reads `BASE: https://gitchamber.com/repos/{owner}/{repo}/{branch}/` (trailing slash). An agent following it literally constructs `…/main//files` (two slashes). Most HTTP servers normalize this, but it's visually confusing and could cause a 404 on a strict router.

**Suggested fix:**
```
BASE: https://gitchamber.com/repos/{owner}/{repo}/{branch}
```
Then `{BASE}/files`, `{BASE}/search/…` etc. produce single-slash paths.

agents/shared/external-repo-reading.md
```diff
@@ -0,0 +1,51 @@
+# Shared external repo reading reference
+
+When you need to read source files from a GitHub repository other than the
+working repo, prefer **https://gitchamber.com** over fetching raw GitHub
+URLs or shelling out to `git clone`.
+
+Gitchamber URLs are plain HTTPS — use whichever tool your agent has:
+- **`Bash` agents:** use the shell examples below.
```
lugassawan: **Low** — "use the shell examples below" pointer on line 8 is stale after the fix

Line 8 says `**Bash agents:** use the shell examples below.` but the examples block (lines 21–27) now contains plain URLs with no shell syntax — equally usable by WebFetch agents. The pointer creates a false asymmetry and could mislead a Bash agent into expecting a `curl …` form it won't find.

**Suggested fix:**
```
- **`Bash` agents:** pass the URLs to `curl -s` or any HTTP client.
```

PR #288, opened 2026-05-20
[chore] Worktree permission-grant hook

### LINKED ISSUE #262: [chore] One-time permission prompt on worktree entry (AskUserQuestion)
### Problem
When entering a worktree (EnterWorktree or rimba add), users are prompted repeatedly for read+write access within the same session — once to navigate, again to read files, again to edit. This friction breaks flow and trains users to dismiss prompts without reading them, reducing security awareness.

### Value
One up-front permission grant per session per worktree eliminates repeated interruptions and clarifies expectations (read+write in a feature branch requires elevated access). The system detects when `.claude/` with relevant settings has already been copied into the worktree and skips prompts in that case, recognizing that permissions are already established. Users know the cost once and decide once, improving both UX and security posture.

### Acceptance criteria
- AskUserQuestion (or new permission-prompt command) is shown exactly once when entering a worktree, before the first access
- If `.claude/` directory with relevant settings files (e.g., `settings.json` containing tool allowlists) already exists in the worktree, the prompt is skipped entirely
- Prompt clearly states the worktree purpose (branch name) and asks for read+write access
- Session state (or hook-local variable) tracks "permission asked for branch X" to avoid re-prompting in the same session
- If user declines, subsequent operations fail gracefully with a message pointing to the initial prompt
- Prompt is skipped if permissions are already granted

### Impact / Effort
- **Impact: S** — UX improvement; no new capability, but removes friction in a common workflow (branch entry → edit → test); detects existing permission inheritance via `.claude/` copy.
- **Effort: M** — Requires hooking worktree entry (EnterWorktree or rimba add), checking for `.claude/` presence, session-state tracking, and AskUserQuestion (or new prompt) wiring; testing on multiple permission states and `.claude/` scenarios.

---

## Type
- [ ] Documentation (README, catalog, SECURITY, etc.)
- [ ] CI / workflow (`.github/workflows/`, actions)
- [ ] Dependency / tooling bump
- [ ] Repository config (templates, Dependabot, labels)
- [x] Release / packaging (`.claude-plugin/`, `scripts/`) — Hook integration, session state management

## Surface
**Files to change:**
- `.claude/settings.json` — add hook or session-state logic (optional, if using hook approach)
- `src/commands/enter-worktree.ts` or equivalent (or rimba integration point)
- `src/prompts/ask-user-question.ts` or new permission-prompt command
- `src/session.ts` (if tracking permission state in session)

## Description

Currently, when entering a worktree, users face repeated permission prompts for the same worktree within a single session:
1. Enter worktree → "Read access?" → grant
2. Read first file → "Write access?" → grant
3. Edit a file → "Write access again?" → grant (or different scope)

This repetition is annoying and trains users to auto-grant without thinking — a security risk.

**Desired state:**
When a user enters a worktree (via EnterWorktree, rimba add, or equivalent):
1. Check if `.claude/` directory exists in the worktree root with relevant settings files (e.g., `settings.json` containing tool allowlists or permissions). If so, skip the permission prompt entirely — permissions are already inherited from the copied directory.
2. Otherwise, show a single AskUserQuestion prompt asking for read+write access in that worktree for the remainder of the session. The prompt clearly states:
   - Branch or worktree name
   - Expected operations (read code, modify files, run tests)
   - Scope (this session only, this worktree only)

Once granted, no further prompts for that worktree in that session. If declined, operations fail with a message like "Permission for feature/ABC-123 was not granted earlier. Re-enter the worktree to ask again."

## Acceptance
- [ ] Hook or session-aware entry point added so prompt fires on worktree enter
- [ ] `.claude/` detection logic written and tested (check for presence and relevant settings files in the worktree)
- [ ] AskUserQuestion (or new command) drafted with clear context
- [ ] Session state or hook variable tracks "permission asked for <branch>" and skips re-prompting
- [ ] Integration tested with EnterWorktree and rimba add workflows, including `.claude/` copy scenarios
- [ ] Documentation updated with the new permission model (WORKFLOWS.md or settings guide)

lugassawan:
## Summary
- Add `hooks/worktree_permission_grant.sh`: a `PreToolUse:Read|Edit|Write` hook that reads `.claude/settings.local.json` from the current linked worktree and emits `permissionDecision: allow` when the requested file matches an allow entry
- Sibling-worktree absolute grants (`//abs/other-wt/**`) are remapped to the current worktree, so rimba-created worktrees inherit their allow lists without re-prompting on every file operation
- Security boundaries: linked-worktree detection via `--absolute-git-dir`, out-of-tree containment (string-prefix + `python3 os.path.realpath` for symlink resolution), `..` traversal guard, `settings.local.json`-only reads (never committed `settings.json`)
- Wire into `hooks/hooks.json` as a `PreToolUse` matcher `Read|Edit|Write`
- 14 pytest tests covering all allow/deny paths, security boundaries, and hooks.json wiring
- `docs/worktree-permission-grant.md` documenting the permission model and security table

## Test Plan
- [x] `shellcheck hooks/worktree_permission_grant.sh` → 0 issues
- [x] `bash scripts/validate.sh` → All checks passed
- [x] `pytest tests/test_hooks.py tests/test_worktree_permission_grant.py` → 81 passed
- [x] Full suite `pytest tests/` → 659 passed (no regressions)
- [ ] Manual smoke: in a fresh rimba worktree, `Read` an in-tree file → no permission prompt; `Edit` an in-tree file → prompt still appears (no write grant to remap)

- [x] `bash scripts/validate.sh` confirms hooks.json schema is valid after the new entry
- [x] Hook is executable (`chmod +x`) and passes `shellcheck`
- [x] No version bump in `plugin.json`/`marketplace.json` (out of scope for feature branches)

Closes #262

### DIFF
--- a/hooks/worktree_permission_grant.sh
+++ b/hooks/worktree_permission_grant.sh
@@ -52,6 +52,7 @@ esac
 wt_root=$(git -C "$file_dir" rev-parse --show-toplevel 2>/dev/null) || exit 0
 
 # Security: never grant for files outside the worktree root.
+# Strict prefix: file_abs must be a child of wt_root, not wt_root itself.
 case "$file_abs" in
     "$wt_root"/*) ;;
     *) exit 0 ;;
@@ -64,6 +65,7 @@ file_rel="${file_abs#"$wt_root"/}"
 # not available on macOS without coreutils; readlink -f requires the path to
 # exist, which breaks Write on new files).
 file_real=$(python3 -c "import os,sys; print(os.path.realpath(sys.argv[1]))" "$file_abs" 2>/dev/null) || exit 0
+# Same strict-prefix requirement: resolved path must also be a child of wt_root.
 case "$file_real" in
     "$wt_root"/*) ;;
     *) exit 0 ;;
--- a/hooks/worktree_permission_grant.sh
+++ b/hooks/worktree_permission_grant.sh
@@ -1,0 +1,171 @@
+#!/usr/bin/env bash
+# PreToolUse:Read|Edit|Write handler — grants permissions for file ops inside
+# the current git linked worktree when the worktree's own
+# .claude/settings.local.json permits them.
+#
+# Never blocks: exit 0 always, never exit 2.  Any error → fail-open (no grant,
+# normal harness prompt).  Grants are limited to files under the worktree root.
+set -u
+
+input=$(cat)
+
+# Fail-open on any jq parse error
+tool_name=$(printf '%s' "$input" | jq -r '.tool_name // empty' 2>/dev/null) || exit 0
+file_path=$(printf '%s' "$input" | jq -r '.tool_input.file_path // empty' 2>/dev/null) || exit 0
+
+[ -z "$tool_name" ] && exit 0
+[ -z "$file_path" ] && exit 0
+
+# Resolve to an absolute path.  Read/Edit/Write always supply absolute paths in
+# Claude Code, but handle the relative case gracefully via the cwd field.
+case "$file_path" in
+    /*)
+        file_abs="$file_path"
+        ;;
+    *)
+        cwd=$(printf '%s' "$input" | jq -r '.cwd // empty' 2>/dev/null) || exit 0
+        [ -z "$cwd" ] && exit 0
+        file_abs="$cwd/$file_path"
+        ;;
+esac
+
+# Reject paths containing a .. segment (refuse to reason about traversal).
+case "$file_abs" in
+    */../*|*/..|../*|..)
+        exit 0
+        ;;
+esac
+
+# Determine git context from the file's own directory (no reliance on cwd in
+# the hook payload, which is absent for file-op tools per CC internals).
+file_dir=$(dirname "$file_abs")
+git_dir=$(git -C "$file_dir" rev-parse --absolute-git-dir 2>/dev/null) || exit 0
+
+# Only act inside a linked worktree — the main checkout is handled natively.
+# Linked worktrees always have a git-dir path of the form:
+#   /repo/.git/worktrees/<name>
+case "$git_dir" in
+    */.git/worktrees/*) ;;
+    *) exit 0 ;;
+esac
+
+wt_root=$(git -C "$file_dir" rev-parse --show-toplevel 2>/dev/null) || exit 0
+
+# Security: never grant for files outside the worktree root.
+case "$file_abs" in
+    "$wt_root"/*) ;;
+    *) exit 0 ;;
+esac
+
+file_rel="${file_abs#"$wt_root"/}"
+
+# Secondary containment: resolve symlinks and verify the real path is still
+# inside the worktree.  python3 is used for portability (GNU realpath -m is
+# not available on macOS without coreutils; readlink -f requires the path to
+# exist, which breaks Write on new files).
+file_real=$(python3 -c "import os,sys; print(os.path.realpath(sys.argv[1]))" "$file_abs" 2>/dev/null) || exit 0
+case "$file_real" in
+    "$wt_root"/*) ;;
+    *) exit 0 ;;
+esac
+
+# Honor only settings.local.json — user-authored and gitignored.
+# Never read a committed settings.json, which an untrusted repo could ship.
+settings="$wt_root/.claude/settings.local.json"
+[ -f "$settings" ] || exit 0
+
+# parent is the directory that contains sibling worktrees (rimba layout:
+# repo-worktrees/<name> are all siblings under the same parent).
+parent=$(dirname "$wt_root")
+
+# Iterate allow entries, looking for a match for the current tool.
+while IFS= read -r entry; do
+    # Filter to entries whose tool prefix matches.
+    case "$entry" in
+        "$tool_name("*) ;;
+        *) continue ;;
+    esac
+
+    # Strip  tool_name( … )  to get the raw glob string.
+    glob="${entry#"$tool_name"(}"
+    glob="${glob%)}"
+
+    # Derive a worktree-relative pattern P from the glob.
+    case "$glob" in
+        //*)
+            # Claude Code absolute-path marker: //abs/path → /abs/path
+            abs_glob="/${glob#//}"
+            case "$abs_glob" in
+                "$wt_root"/*)
+                    P="${abs_glob#"$wt_root"/}"
+                    [ -z "$P" ] && P="**"
+                    ;;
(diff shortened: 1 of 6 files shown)


tests/test_worktree_permission_grant.py
```diff
@@ -0,0 +1,316 @@
...
+    # TC5 — no .claude/settings.local.json → no grant
+    def test_missing_settings_is_noop(self, wt_env):
+        wt = wt_env["wt"]
+        # .claude/ exists but settings.local.json is absent
+        result = _run(
+            {
+                "tool_name": "Read",
+                "tool_input": {"file_path": str(wt / "README")},
+            }
+        )
```
lugassawan, line 169: **High** — No positive `Edit` grant test.

`TC6` confirms `Edit` is denied when only `Read` entries exist, but there is no counterpart confirming an `Edit(...)` entry actually emits `permissionDecision: allow`. If the real Claude Code `Edit` tool uses a key other than `file_path` in `tool_input`, the hook would silently never grant Edit operations.

Suggested addition:
```python
def test_edit_grant_allows(self, wt_env):
    wt, other = wt_env["wt"], wt_env["other"]
    _settings(wt, [f"Edit({_cc_abs(other)}/**)"])
    result = _run({"tool_name": "Edit", "tool_input": {"file_path": str(wt / "README")}})
    assert result.returncode == 0
    assert json.loads(result.stdout) == _ALLOW_JSON
```

tests/test_worktree_permission_grant.py
```diff
@@ -0,0 +1,316 @@
...
+    def test_malformed_stdin_is_failopen(self):
+        result = subprocess.run(
+            ["bash", str(HOOK_SH)],
+            input="not-valid-json{{{",
+            capture_output=True,
+            text=True,
+            env=_CLEAN_ENV,
+        )
+        assert result.returncode == 0
+        assert result.stdout == ""
```
lugassawan, line 257: **Medium** — The symlink-containment guard (`python3 realpath`) has no test.

The hook resolves symlinks and re-checks containment — the most security-critical boundary. The docs list "resolved path must also be under `wt_root`" as a security guarantee, yet no test creates a symlink inside the worktree pointing outside it and asserts the hook does NOT grant.

Suggested addition:
```python
def test_symlink_escape_denied(self, wt_env):
    wt, other = wt_env["wt"], wt_env["other"]
    secret = wt_env["wts"].parent / "secret.txt"
    secret.write_text("secret")
    link = wt / "escape_link"
    link.symlink_to(secret)
    _settings(wt, [f"Read({_cc_abs(other)}/**)"])
    result = _run({"tool_name": "Read", "tool_input": {"file_path": str(link)}})
    assert result.returncode == 0
    assert result.stdout == ""
```

tests/test_hooks.py
```diff
@@ -255,3 +255,29 @@ def test_pre_push_runs_pytest(self):
         assert "pytest" in content and "tests/" in content, (
             "pre-push hook must invoke pytest against the tests/ directory"
         )
+
+
+# ──────────────────────────────────────────────
+# worktree permission-grant hook — wiring
+# ──────────────────────────────────────────────
+
```
lugassawan, line 263: **Low** — `TestWorktreePermissionHookWiring` here duplicates `TestHooksJsonEntry` in `tests/test_worktree_permission_grant.py` — both assert the same `Read|Edit|Write` matcher + `worktree_permission_grant.sh` command. Consider keeping the wiring check in one place (here, with the other hook-wiring tests) and removing the `TestHooksJsonEntry` class from the other file.

tests/test_worktree_permission_grant.py
```diff
@@ -0,0 +1,316 @@
...
+    def test_single_slash_abs_glob_matches(self, wt_env):
+        wt = wt_env["wt"]
+        _settings(wt, [f"Read({str(wt)}/**)"])
+        result = _run(
+            {
+                "tool_name": "Read",
+                "tool_input": {"file_path": str(wt / "README")},
+            }
+        )
+        assert result.returncode == 0
```
lugassawan, line 296: **Low** — TC comment numbers are out of sequence (TC12 appears after TC14). Method names are already self-documenting so dropping the `# TC<N>` comments entirely would be cleaner than renumbering.

tests/test_worktree_permission_grant.py
lugassawan: Addressed in e18fcea: added `test_edit_grant_allows` as the positive Edit counterpart to the deny test.

tests/test_worktree_permission_grant.py
lugassawan: Addressed in e18fcea: added `test_symlink_escape_denied` covering the python3 realpath guard.

tests/test_hooks.py
lugassawan: Addressed in e18fcea: removed `TestHooksJsonEntry` class and `HOOKS_JSON` import from test_worktree_permission_grant.py; wiring assertions now live exclusively in test_hooks.py.

tests/test_worktree_permission_grant.py
lugassawan: Addressed in e18fcea: removed all TC-number inline comments; method names are self-describing.

hooks/worktree_permission_grant.sh
```diff
@@ -0,0 +1,171 @@
...
+#   /repo/.git/worktrees/<name>
+case "$git_dir" in
+    */.git/worktrees/*) ;;
+    *) exit 0 ;;
+esac
+
+wt_root=$(git -C "$file_dir" rev-parse --show-toplevel 2>/dev/null) || exit 0
+
+# Security: never grant for files outside the worktree root.
+case "$file_abs" in
```
lugassawan, line 56: **Medium** — `file_abs == wt_root` exactly is silently rejected by the path containment guard (and the equivalent `file_real` check), but this intent isn't stated anywhere.

The guard uses `"$wt_root"/*` which requires a child path — so a `file_path` equal to `wt_root` itself exits silently at line 57. In practice `Read`/`Edit`/`Write` never target a bare directory, so this is not a real-world failure, but the asymmetry could confuse a future reader.

Suggested clarifying comment (no code change required):
```bash
# Strict prefix: file_abs must be a child of wt_root, not wt_root itself.
case "$file_abs" in
    "$wt_root"/*) ;;
    *) exit 0 ;;
esac
```

lugassawan: Addressed in 421c8e9: added clarifying comments on both containment guards (`file_abs` and `file_real`).

PR #293, opened 2026-05-21
[fix] Fix stale rimba add invocations for v1.9.x

lugassawan:
## Summary
- Fix `rimba add` invocations in `workflow-address-feedback` skill and its command prose: `--task` is only valid in `pr:<num>` mode (rimba v1.9.x), so plain-branch form was replaced with `rimba add "pr:$PR" --task ...`
- Document `rimba add branch:<current-branch>` promote mode in `workflow-development` SKILL.md and its plan template — covers the "work already started on a branch" case that was undocumented
- Add `tests/test_rimba_references.py`: two regression tests (text-scan, no rimba binary required) that pin the `--task ⟹ pr:` invariant and verify all three `add` modes are documented

## Test Plan
- [x] `pytest tests/test_rimba_references.py` — both tests red before fix, green after
- [x] `pytest tests/ -q` — 747 passed, 0 failures
- [x] `scripts/validate.sh` — all checks passed

Issue: N/A


tests/test_rimba_references.py
```diff
@@ -0,0 +1,24 @@
+"""Structural tests: documented rimba invocations match the v1.9.x CLI."""
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+DOC_DIRS = [ROOT / "skills", ROOT / "commands"]
```
lugassawan, line 6: **Medium — DOC_DIRS excludes `tests/`**

`test_rimba_add_task_flag_requires_pr_mode` only scans `skills/` and `commands/`, so any future `.md` doc or fixture string in `tests/` that demos the old invocation pattern (`rimba add <branch> --task ...`) would go undetected.

Either extend the scan:
```python
DOC_DIRS = [ROOT / "skills", ROOT / "commands", ROOT / "tests"]
```
Or add an intentional-exclusion comment so the next reader knows the scope was deliberate:
```python
# tests/ intentionally excluded — fixture strings are not agent-readable docs
DOC_DIRS = [ROOT / "skills", ROOT / "commands"]
```

tests/test_rimba_references.py
```diff
@@ -0,0 +1,24 @@
...
+            for lineno, line in enumerate(path.read_text(encoding="utf-8").splitlines(), 1):
+                if "rimba add" in line and "--task" in line and "pr:" not in line:
+                    offenders.append(f"{path.relative_to(ROOT)}:{lineno}")
+    assert not offenders, "rimba add --task used outside pr: mode:\n" + "\n".join(offenders)
+
+
+def test_workflow_development_documents_all_add_modes():
+    """workflow-development must document task, pr:, and branch: add modes."""
+    text = WORKFLOW_DEV.read_text(encoding="utf-8")
+    for token in ("rimba add", "pr:", "branch:"):
```
lugassawan: **Low — `"task"` token matches the generic `<task>` placeholder, not `--task` flag**

The docstring says "task, pr:, and branch: add modes" but the token `"task"` matches line 110's `rimba add [<service>/]<task>` placeholder, not a `--task` flag reference. If `--task` documentation is ever removed, this assertion still passes vacuously.

Consider either:
```python
for token in ("rimba add", "pr:", "branch:", "--task"):
```
or update the docstring to clarify that `"rimba add"` is the proxy for the positional-task mode.

skills/workflow-development/SKILL.md
```diff
@@ -108,6 +108,7 @@ RIMBA=$(command -v rimba 2>/dev/null \
 
 - **rimba MCP server active:** invoke the `add` tool on it (`rimba mcp`) — no shell process needed. Use `add pr:<num>` when implementing from a PR number.
 - **`$RIMBA` non-empty (binary found):** run `$RIMBA add [<service>/]<task> [--flag]` (or `$RIMBA add pr:<num> --task "<label>"` for a PR). Rimba handles branch-prefix conventions (`feature/`, `bugfix/`, `hotfix/`, `docs/`, `test/`, `chore/`), `.env`/`.tool-versions`/`.vscode` copying, `post_create` hooks, and lockfile sharing.
+- **Promote work already started** — if you began editing on the current branch in the main checkout (not the default branch), `$RIMBA add branch:<current-branch>` moves that work into its own worktree, transferring dirty changes via `git stash`. `--source` is not valid in this mode.
```
lugassawan, line 111: **Low — `$RIMBA` variable form vs bare `rimba` in template**

This SKILL.md bullet uses `$RIMBA add branch:<current-branch>`, consistent with the preceding binary-found bullet at line 110. The parallel bullet in `templates/plan-workflow-section.md` correctly uses bare `rimba add branch:<current-branch>` (the template has no `$RIMBA` variable in scope).

Minor cross-file inconsistency — no change required unless you want uniform bare-name form in both files. The SKILL.md form is defensible since it is in execution context.

tests/test_rimba_references.py
lugassawan: Addressed in b7b4815: added intentional-exclusion comment to DOC_DIRS clarifying that tests/ is out of scope because fixture strings are not agent-readable documentation.

tests/test_rimba_references.py
lugassawan: Addressed in b7b4815: added "--task" to the token list so the assertion explicitly verifies --task flag documentation rather than relying on the generic <task> placeholder match.

tests/test_rimba_references.py
```diff
@@ -0,0 +1,25 @@
...
+WORKFLOW_DEV = ROOT / "skills" / "workflow-development" / "SKILL.md"
+
+
+def test_rimba_add_task_flag_requires_pr_mode():
+    """`rimba add --task` is only valid in pr:<num> mode (rimba v1.9.x)."""
+    offenders = []
+    for d in DOC_DIRS:
+        for path in d.rglob("*.md"):
+            for lineno, line in enumerate(path.read_text(encoding="utf-8").splitlines(), 1):
+                if "rimba add" in line and "--task" in line and "pr:" not in line:
```
lugassawan, line 16: **Low — substring guard `"pr:" not in line` may produce false negatives**

The condition `"rimba add" in line and "--task" in line and "pr:" not in line` allows any line that contains `pr:` *anywhere* (e.g. in a comment or unrelated word) to bypass the check while still using `--task` outside PR mode.

Tightening to require `pr:` as part of the subcommand argument is more precise and equally readable:

```suggestion
                if "rimba add" in line and "--task" in line and "rimba add pr:" not in line:
```

Low-probability in a docs-only repo, but cheap to fix.

lugassawan: Investigated the suggested tightening — `"rimba add pr:" not in line` is actually stricter in the wrong direction. `skills/workflow-address-feedback/SKILL.md:93` uses the quoted form `rimba add "pr:$PR"`, which contains `pr:` but does NOT contain the literal `rimba add pr:` (a quote character sits between `add` and `pr:`). The tighter guard would produce a false negative on that line — exactly the failure mode this test is meant to prevent.

The original guard `"pr:" not in line` correctly covers both the unquoted form (`rimba add pr:$PR`) and the quoted form (`rimba add "pr:$PR"`), since both contain the substring `pr:`. Leaving it as-is.

PR #298, opened 2026-05-21
[feat] Persist workflow state across auto-compaction

### LINKED ISSUE #286: [feat] Persist swe-workbench workflow state across auto-compaction
## Problem

When Claude Code auto-compacts a long conversation due to high context usage, any active
swe-workbench workflow state (the current mode, the step the agent is on, decisions made
earlier in the session) is silently dropped. The next turn starts with no memory that a
workflow was in progress, leaving the user mid-flight with no way to resume cleanly.

## Goal

Provide a mechanism so that a swe-workbench workflow in progress can survive an
auto-compaction event and be transparently resumed on the next turn — without requiring the
user to re-explain the task from scratch.

## Motivation

Auto-compaction happens precisely when a task is large and complex — exactly the scenarios
where swe-workbench workflows (`workflow-development`, `workflow-bug-triage`,
`workflow-pr-review`, etc.) are most valuable. Losing workflow state at the worst moment
undermines the core value proposition of the plugin.

Concrete scenario:
1. User kicks off `/implement` on a multi-file feature.
2. Workflow progresses through branch creation, planning, and several edit steps.
3. Context fills up; Claude Code auto-compacts.
4. The next response has no knowledge that a workflow was running — the workflow is lost.

## Affected surfaces

- `workflow-development` (and all `workflow-*` skills)
- `hook` layer (session-start or post-compact hooks could trigger resume)
- `documentation` (users need guidance on what survives compaction and what doesn't)

## Sketch

Several approaches exist; the right one needs design discussion:

**A — Hook-based checkpoint**: Write lightweight workflow state (mode name, current step,
key decisions) to a sidecar file (e.g., `.claude/workflow-state.json`) at the end of each
workflow step. A `PostToolUse` or session-start hook reads this file and injects a resume
preamble if it finds an incomplete workflow.

**B — context-mode persistence**: Route workflow state through the context-mode knowledge
base so it survives compaction natively (context-mode already persists across the
compaction boundary).

**C — Explicit `/resume` command**: Give users a `/resume` skill that reads
`.claude/workflow-state.json` and reconstructs the workflow mid-session, without requiring
automatic hook wiring.

## Acceptance criteria

- [ ] A workflow interrupted by auto-compaction can be resumed without the user re-stating
      the original task or re-running setup steps.
- [ ] Resuming a workflow picks up at the correct step, not from the beginning.
- [ ] If no workflow was in progress, the resume path is a no-op (no noise for clean
      sessions).
- [ ] The solution is documented so users understand what is and isn't preserved.
- [ ] Existing workflow skills (`workflow-development`, `workflow-bug-triage`,
      `workflow-pr-review`, etc.) are updated to checkpoint their state at step boundaries.

lugassawan:
## Summary

- Add `hooks/workflow_resume_hint.sh` — a `SessionStart(compact)` hook that detects a per-branch checkpoint file and injects a resume preamble so interrupted workflows continue at the correct phase after auto-compaction
- Wire the hook in `hooks/hooks.json` under `SessionStart` with `matcher: "compact"`; path-containment guard and fail-open (exit 0 always) by design
- Add checkpoint-write and delete-on-terminal instructions to `workflow-development` (Mode B), `workflow-bug-triage`, and `workflow-pr-review` skills, all referencing `docs/workflow-state.md` for the schema
- Add `docs/workflow-state.md` covering the state-file path, schema, lifecycle table, what survives compaction, smoke test, and troubleshooting

## Test Plan

- [x] `bash scripts/validate.sh` passes (0 issues, all skill line caps respected)
- [x] `python -m pytest tests/` passes (804 tests, 0 failures)
- [x] Smoke test: `echo '{"cwd":"<toplevel>"}' | bash hooks/workflow_resume_hint.sh` emits correct `hookSpecificOutput` JSON for a fresh state file and empty output for no file
- [x] `bash -n hooks/workflow_resume_hint.sh` passes (syntax OK)

- [ ] No-op path: session with no state file → hook emits nothing, exits 0
- [ ] Resume path: valid, fresh, branch-matching file → preamble injected with skill + phase + completed phases
- [ ] Staleness path: file >24h old → deleted, no injection
- [ ] Fail-open path: branch mismatch / version ≠ 1 / malformed JSON → no injection, exit 0
- [ ] Branch with `%` in name → valid JSON output (no printf format-string corruption)

Closes #286

### DIFF
--- a/docs/workflow-state.md
+++ b/docs/workflow-state.md
@@ -54,7 +54,7 @@ but different branches never share a checkpoint file.
 - `mode` — optional mode letter (A/B/C for workflow-development; absent for single-mode skills).
 - `phase` / `phase_label` — current phase/step number and human label; `completed_phases` lists all finished phases.
 - `context` — flat free-form bag; `context.branch` is the resume-validation key. Skills pack skill-specific values here (`pr`, `head_sha`, `decision` for pr-review; project-detection values for workflow-development).
-- `updated_at` — ISO-8601 timestamp; the hook uses filesystem mtime for staleness, not this field.
+- `updated_at` — ISO-8601 timestamp (informational — not used for staleness; the hook uses filesystem mtime instead).
 
 ### Lifecycle
 
--- a/docs/workflow-state.md
+++ b/docs/workflow-state.md
@@ -1,0 +1,160 @@
+# Workflow state persistence
+
+When Claude Code auto-compacts a long conversation, any in-progress swe-workbench
+workflow state lives only in the model's working context and is silently dropped.
+Auto-compaction happens precisely on large, complex tasks — exactly where these workflows
+matter most — leaving the user mid-flight with no clean way to resume.
+
+Workflow state persistence externalizes that state to a sidecar JSON file written by the
+model at each phase/step boundary, and re-injects a resume preamble after compaction so
+the workflow continues at the correct step.
+
+## How it works
+
+### State file
+
+| Property | Value |
+|---|---|
+| **Path** | `<git-toplevel>/.claude/cache/workflow-state/<branch-with-slashes-as-dashes>.json` |
+| **Example** | `.claude/cache/workflow-state/feature-286-workflow-state-persistence.json` |
+| **Directory** | `.claude/cache/` — the plugin's ephemeral-state home; gitignored |
+
+Both the model (writer) and the hook (reader) resolve the git toplevel via
+`git rev-parse --show-toplevel` from the session's cwd. This means the path agreement
+holds whether the session runs in the main checkout or a linked worktree.
+
+Per-branch filenames give worktree isolation for free: two worktrees on the same repo
+but different branches never share a checkpoint file.
+
+### Schema
+
+```json
+{
+  "version": 1,
+  "skill": "swe-workbench:workflow-development",
+  "mode": "B",
+  "phase": "3",
+  "phase_label": "Verify",
+  "completed_phases": ["1", "2"],
+  "context": {
+    "branch": "feature/286-workflow-state-persistence",
+    "worktree_root": "/abs/path/to/worktree",
+    "pr": null,
+    "base": null,
+    "head_sha": null,
+    "decision": null,
+    "notes": "free-form key decisions / identifiers"
+  },
+  "updated_at": "2026-05-21T10:30:00Z"
+}
+```
+
+- `version` — schema version; the hook ignores files with unknown versions (fail-soft).
+- `skill` — fully-qualified skill name driving the workflow.
+- `mode` — optional mode letter (A/B/C for workflow-development; absent for single-mode skills).
+- `phase` / `phase_label` — current phase/step number and human label; `completed_phases` lists all finished phases.
+- `context` — flat free-form bag; `context.branch` is the resume-validation key. Skills pack skill-specific values here (`pr`, `head_sha`, `decision` for pr-review; project-detection values for workflow-development).
+- `updated_at` — ISO-8601 timestamp; the hook uses filesystem mtime for staleness, not this field.
+
+### Lifecycle
+
+| Actor | Event | Action |
+|---|---|---|
+| **Model** | Phase/step transition | Write (overwrite) the state file with the new phase |
+| **Model** | Terminal phase success | Delete the state file |
+| **Hook** | SessionStart(compact) | Read the file; inject resume preamble if fresh + valid |
+| **Hook** | SessionStart(compact), file >24h old | Delete the file; no injection (staleness sweep) |
+| **Hook** | SessionStart(compact), branch mismatch or version ≠ 1 | No injection (fail-open, file untouched) |
+
+The state file is a **fail-soft cache, not a source of truth**. The model owns the semantic
+lifecycle (write/delete at phase boundaries). The hook owns the mechanical lifecycle
+(detect-and-inject, age-sweep). The hook never reconstructs state — any ambiguity degrades
+to "ask", because a wrong resume is worse than no resume.
+
+### Hook registration
+
+| Event | Matcher | Script |
+|---|---|---|
+| `SessionStart` | `"compact"` | `hooks/workflow_resume_hint.sh` |
+
+`SessionStart` requires a `matcher` string (not a lifecycle event in `validate.py`); the
+`"compact"` matcher targets only compaction-triggered session starts.
+
+The hook always exits 0 — it never blocks session startup.
+
+## What survives compaction and what doesn't
+
+| Information | Survives? | Why |
+|---|---|---|
+| Current phase and phase label | ✓ | Written to state file at each transition |
+| Completed phases list | ✓ | Written to state file at each transition |
+| Skill name and mode | ✓ | Written to state file on entry |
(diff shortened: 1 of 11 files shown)


docs/workflow-state.md
```diff
@@ -0,0 +1,160 @@
...
+
+## How it works
+
+### State file
+
+| Property | Value |
+|---|---|
+| **Path** | `<git-toplevel>/.claude/cache/workflow-state/<branch-with-slashes-as-dashes>.json` |
+| **Example** | `.claude/cache/workflow-state/feature-286-workflow-state-persistence.json` |
+| **Directory** | `.claude/cache/` — the plugin's ephemeral-state home; gitignored |
```
lugassawan, line 20: **High: `.claude/cache/` not in `.gitignore`**

The doc states `.claude/cache/` is the plugin's ephemeral-state home and gitignored, but no `.gitignore` entry exists. State files written under `.claude/cache/workflow-state/` will be tracked by git in consumer repos, leaking branch names, PR numbers, workflow phase, and free-form notes into commits. Users following the smoke-test example would inadvertently stage this file.

**Fix:** Add `.claude/cache/` to `.gitignore`:
```
.claude/cache/
```

skills/workflow-pr-review/SKILL.md
```diff
@@ -28,7 +28,7 @@ This skill orchestrates; analysis is delegated to:
 
 - `swe-workbench:reviewer` subagent — produces `Severity | File:Line | Issue | Why | Fix` findings + a Review Decision footer (when instructed by this skill — see Step 4).
 - `swe-workbench:ticket-context` skill — prepended to the reviewer prompt when the PR body or commit messages reference a ticket key, atlassian/Confluence URL, or `#NNN` GitHub ref.
-
+- **Checkpoint:** write the workflow state file (see `docs/workflow-state.md`) at each step boundary, carrying `$PR`/`$BASE`/`$HEAD_SHA`/`$DECISION` in `context`. Delete the state file after Step 7.
```
lugassawan, line 30: **Medium: Missing blank line before `## 7-step flow` heading**

The Checkpoint bullet is appended immediately before the section heading with no blank line. Some Markdown renderers absorb the preceding list item into the heading's content block rather than treating it as a standalone list item, which may affect how the model reads the compiled skill.

**Fix:** Add a blank line after the Checkpoint bullet and before `## 7-step flow`.

hooks/workflow_resume_hint.sh
```diff
@@ -0,0 +1,89 @@
...
+    real_dir=$(python3 -c "import os,sys; print(os.path.realpath(sys.argv[1]))" \
+        "$state_dir" 2>/dev/null) || return 0
+    case "$real_state" in "$real_dir/"*) ;; *) return 0 ;; esac
+
+    # No-op gate: no checkpoint for this branch
+    [ -f "$state_file" ] || return 0
+
+    # Staleness gate: file older than 24h is deleted and skipped.
+    # Only the current branch's file is removed; other branches manage their own.
+    stale=$(find "$state_file" -mmin +1440 2>/dev/null)
```
lugassawan, line 44: **Low: `find -mmin +1440` races with a concurrent state-file write**

Between the `-f` existence check and this `find` call, the model could overwrite the state file with a fresh timestamp, causing the staleness check to report a stale file as fresh. In practice the compaction window makes this race vanishingly unlikely, and the worst consequence is a spurious resume at the wrong phase — not data loss.

No change needed unless the staleness sweep is generalized to scan the whole directory.

tests/test_workflow_resume_hook.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+# ---------------------------------------------------------------------------
+# No-op cases — hook must emit nothing and exit 0
+# ---------------------------------------------------------------------------
+
+class TestNoOp:
+    """Hook is silent when there is no state to resume."""
+
+    def test_no_state_file_exits_clean(self, hook_script, git_repo):
```
lugassawan, line 133: **Low: Test isolation note — keep `git_repo` function-scoped**

`test_no_state_file_exits_clean` and `test_stale_file_swept_and_no_injection` share the function-scoped `git_repo` fixture correctly. Adding `scope="module"` to `git_repo` in a future refactor would break isolation here because the stale-file test mutates mtime and relies on cleanup by the hook itself.

No change needed; the current function-scoped `git_repo` is correct. This is a caution against future refactoring.

docs/workflow-state.md
```diff
@@ -0,0 +1,160 @@
...
+  "updated_at": "2026-05-21T10:30:00Z"
+}
+```
+
+- `version` — schema version; the hook ignores files with unknown versions (fail-soft).
+- `skill` — fully-qualified skill name driving the workflow.
+- `mode` — optional mode letter (A/B/C for workflow-development; absent for single-mode skills).
+- `phase` / `phase_label` — current phase/step number and human label; `completed_phases` lists all finished phases.
+- `context` — flat free-form bag; `context.branch` is the resume-validation key. Skills pack skill-specific values here (`pr`, `head_sha`, `decision` for pr-review; project-detection values for workflow-development).
+- `updated_at` — ISO-8601 timestamp; the hook uses filesystem mtime for staleness, not this field.
```
lugassawan: **Low: `updated_at` is informational only — not used for staleness**

The hook correctly uses filesystem mtime for its staleness gate, not this field. If a skill writes an incorrect `updated_at` timestamp, it has no effect on behavior. The field is useful for human debugging but could confuse readers who assume the hook uses it.

**Suggestion:** Note `(informational — not used for staleness)` in the schema table or troubleshooting section to prevent future confusion.

docs/workflow-state.md
lugassawan: Addressed in 75e4c7d: Added `.claude/cache/` to `.gitignore`. State files written under `.claude/cache/workflow-state/` are now excluded from tracking in consumer repos.

skills/workflow-pr-review/SKILL.md
lugassawan: Addressed in 75e4c7d: Added blank line before `## 7-step flow`. To stay within the 300-line cap, the blank line between the Composition intro sentence and the first bullet was removed (semantically neutral in CommonMark).

hooks/workflow_resume_hint.sh
lugassawan: Acknowledged. The race is benign: if the model writes a fresh state file between the `-f` check and the `find -mmin +1440` call, the worst outcome is that a fresh file passes the staleness gate as expected. The concern would only materialize if the model wrote a stale-timestamped file, which isn't a realistic scenario. No change made.

tests/test_workflow_resume_hook.py
lugassawan: Acknowledged — good caution for future refactoring. The `git_repo` fixture is deliberately function-scoped to allow the stale-file test to mutate mtime and rely on hook cleanup. A module-scoped refactor would break that isolation. No change needed; this note is captured.

docs/workflow-state.md
lugassawan: Addressed in 75e4c7d: Updated the `updated_at` description to `(informational — not used for staleness; the hook uses filesystem mtime instead)` to prevent reader confusion.

PR #299, opened 2026-05-21
[feat] address-feedback: reuse current worktree when on the PR branch

### LINKED ISSUE #295: [feat] workflow-address-feedback: detect existing worktree before rimba add
## Product framing

**Problem:** When the user runs `workflow-address-feedback` from within a worktree that already has the PR branch checked out, the skill unconditionally runs `rimba add "pr:$PR" --task "address-feedback-$PR"`, attempting to create a new worktree. This interrupts the user's session unnecessarily — they have to cancel and tell the agent "we're already in the worktree."

**Value:** Detecting the current environment state before spawning a new worktree eliminates friction when the user is already positioned on the PR branch. Users stay in their existing context instead of being redirected to a new worktree.

**Acceptance criteria:**
- Given the user is in a worktree whose branch matches the PR head branch, when they run `/swe-workbench:address-feedback <N>`, then the skill skips `rimba add` and continues in the current worktree.
- Given the user is NOT on the PR branch, when they run the command, then the skill creates the worktree as normal via `rimba add "pr:$PR" --task "address-feedback-$PR"`.
- The detection uses `git rev-parse --abbrev-ref HEAD` and compares against `headRefName` from the PR JSON.

**Impact:** Small — addresses a usability friction point; not critical to core functionality but improves session flow for common workflows.  
**Effort:** Small — requires one `git rev-parse` check and a conditional skip of the `rimba add` block.

## Reproduction

1. Open a PR with review feedback threads.
2. Manually check out or be positioned in a worktree for that PR's branch.
3. Run `/swe-workbench:address-feedback <PR_number>`.
4. Observe: skill tries to create a duplicate worktree, forcing the user to interrupt.

## Expected behavior

Skill detects `git rev-parse --abbrev-ref HEAD` matches `headRefName`, sets `WT=$(pwd)`, and proceeds directly to Phase 3 (triage).

---
_Reported via `/swe-workbench:report-issue` — plugin v0.1.13, Claude Code 2.1.145._

lugassawan:
## Summary

- Phase 2 of `workflow-address-feedback` now opens with a reuse-guard: if `git rev-parse --abbrev-ref HEAD` already equals `$PR_BRANCH` (and is not a detached HEAD), the skill sets `WT=$(pwd)` and `REUSED_WT=1`, skipping `rimba add` entirely.
- A dirty-tree check emits a non-blocking warning so the user can catch unrelated edits before Phase 4 commits them.
- Phase 6 now checks `REUSED_WT` before running cleanup: when `1`, it skips `rimba remove` and the `git worktree remove --force`/`rm -rf` fallback — preventing destruction of the user's live checkout.
- Two regression tests added (21 total in the address-feedback suite).

## Test Plan

- [ ] `pytest tests/test_workflow_address_feedback_skill.py -v` — all 21 pass
- [ ] `pytest tests/ -v` — all 767 pass (no regressions)
- [ ] `bash scripts/validate.sh` — all checks passed

- [ ] When the current branch matches `$PR_BRANCH`, the reuse-guard fires: `WT=$(pwd)` is set, `REUSED_WT=1` is set, and `rimba add` is not called.
- [ ] Phase 6 emits "Reused existing worktree … skipping cleanup" and does not run `rimba remove` or `git worktree remove --force` when `REUSED_WT=1`.
- [ ] When the current branch does NOT match `$PR_BRANCH`, the guard does not fire and both the rimba-path and fallback-path remain reachable.
- [ ] In detached HEAD state (`git rev-parse --abbrev-ref HEAD` returns `HEAD`), the guard does not fire.

Closes #295

### DIFF
--- a/skills/workflow-address-feedback/SKILL.md
+++ b/skills/workflow-address-feedback/SKILL.md
@@ -108,7 +108,21 @@ if [ "$CURRENT_BRANCH" = "$PR_BRANCH" ] && [ "$CURRENT_BRANCH" != "HEAD" ]; then
 fi
 ```
 
-If `$WT` is set by the check above, skip the rest of Phase 2 and proceed to Phase 3 — when the tree was dirty, a non-blocking warning was already emitted; the user may stash before Phase 4 commits. Otherwise create a new durable worktree:
+If `$WT` is not yet set, check whether a worktree for `$PR_BRANCH` already exists elsewhere on disk (e.g. the session is on `main` but the branch was checked out previously):
+
+```bash
+if [ -z "$WT" ]; then
+  EXISTING_WT=$(git worktree list --porcelain \
+    | awk 'BEGIN{wt=""} /^worktree /{wt=$2} /^branch refs\/heads\/'"$PR_BRANCH"'/{print wt; exit}')
+  if [ -n "$EXISTING_WT" ] && [ -d "$EXISTING_WT" ]; then
+    WT="$EXISTING_WT"
+    REUSED_WT=1
+    echo "Found existing worktree for '$PR_BRANCH' at $WT (skipping rimba add)."
+  fi
+fi
+```
+
+If `$WT` is set by either check above, skip the rest of Phase 2 and proceed to Phase 3 — when the tree was dirty (first guard only), a non-blocking warning was already emitted; the user may stash before Phase 4 commits. Otherwise create a new durable worktree:
 
 **When rimba is available** (preferred):
 
@@ -238,7 +252,7 @@ Cleanup is **failure-tolerant**: if both rimba and the git fallback fail, log a 
 
 | Mistake | Fix |
 |---|---|
-| Create a new worktree when already on the PR branch | Phase 2 first compares `git rev-parse --abbrev-ref HEAD` against `$PR_BRANCH`; on a match it reuses `$(pwd)` and skips `rimba add`. Only fall through to creation when the branches differ. |
+| Create a new worktree when already on the PR branch or when one already exists | Phase 2 runs two guards before `rimba add`: (1) compares `git rev-parse --abbrev-ref HEAD` against `$PR_BRANCH` — match reuses `$(pwd)`; (2) scans `git worktree list --porcelain` for a registered worktree on `$PR_BRANCH` — match reuses that path. Only fall through to creation when both checks find nothing. |
 | Omit `--skip-deps --skip-hooks` on the rimba call | Always pass both flags — same as other worktree-creating skills. Deps can be installed manually in the worktree if needed. |
 | Leave the worktree behind after skill exits | Phase 6 always removes it — skip Phase 6 only when exiting before Phase 2 (no worktree created yet) or when the reuse-guard fired (`REUSED_WT=1`, nothing was created). |
 | Deleting `$PR_BRANCH` directly in Phase 6 fallback cleanup | Only remove the worktree directory — `$PR_BRANCH` is the real PR head branch; deleting it via `git branch -D` would destroy the owner's PR. |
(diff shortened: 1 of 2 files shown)


tests/test_workflow_address_feedback_skill.py
```diff
@@ -207,3 +207,36 @@ def test_address_feedback_skill_cleanup_uses_existing_wt():
         "Phase 6 must not re-assign $WT — use the value set in Phase 2 so the fallback "
         "targets the correct worktree directory regardless of which Phase 2 branch (rimba vs. git) ran"
     )
+
+
+def test_address_feedback_skill_reuses_worktree_when_on_pr_branch():
+    """Phase 2 must reuse the current worktree when the branch matches the PR head (closes #295)."""
+    text = SKILL_MD.read_text()
+    # Detection: compare current branch against $PR_BRANCH on one guard line.
+    guard_lines = [
```
lugassawan: **Medium — test guard matches prose row, not the code block**

`guard_lines` searches for lines containing both `rev-parse --abbrev-ref HEAD` and `PR_BRANCH`. In the skill those tokens never co-appear in the code block: `CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)` has no `PR_BRANCH`, and the `if [ "$CURRENT_BRANCH" = "$PR_BRANCH" ]` line has no `rev-parse`. The only line matching both is the Common mistakes table row — so removing the Phase 2 code while keeping the table would leave this test green.

**Suggested fix:**
```python
assert "CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)" in text, (
    "SKILL.md Phase 2 must assign CURRENT_BRANCH via git rev-parse --abbrev-ref HEAD"
)
assert re.search(r\\s*=\s*\, text), (
    "SKILL.md Phase 2 must compare $CURRENT_BRANCH against $PR_BRANCH in the if-condition"
)
```

skills/workflow-address-feedback/SKILL.md
```diff
@@ -95,6 +95,21 @@ If a prior triage save exists at `/tmp/swe-workbench-address-feedback/${PR}-tria
...
+First, check whether the current branch already matches the PR head — if so, reuse the current worktree instead of creating a new one:
+
+```bash
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [ "$CURRENT_BRANCH" = "$PR_BRANCH" ] && [ "$CURRENT_BRANCH" != "HEAD" ]; then
+  WT=$(pwd)
+  REUSED_WT=1
+  echo "Already on PR branch '$PR_BRANCH' — reusing the current worktree at $WT (skipping rimba add)."
+  DIRTY=$(git status --porcelain)
+  [ -n "$DIRTY" ] && echo "Note: working tree has uncommitted changes; confirm before fixes are committed so unrelated edits aren't swept into the feedback commit."
```
lugassawan: **Low — dirty-tree prose implies a checkpoint that does not exist**

"the user can intervene before Phase 4 commits" implies a pause-and-confirm step, but the code just emits a `Note:` echo and continues immediately into Phase 3. There is no mechanism to halt. If an LLM follows the prose literally, it may still proceed without giving the user an opportunity to stash.

**Suggested softer wording:** "the user may stash before Phase 4 commits" — or add an explicit `read -p "Uncommitted changes detected. Continue? [y/N] "` step after the warning.

tests/test_workflow_address_feedback_skill.py
lugassawan: Addressed in d6a05d25b25efbd50da7aec534940bdf0903a839: replaced the vacuous `guard_lines` comprehension with two concrete assertions — one for `CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)` and one regex-matching the if-condition `"$CURRENT_BRANCH" = "$PR_BRANCH"`.

skills/workflow-address-feedback/SKILL.md
lugassawan: Addressed in d6a05d25b25efbd50da7aec534940bdf0903a839: softened the echo message and the follow-on prose to say 'the user may stash before Phase 4 commits' — removing the implication of a built-in blocking checkpoint.

PR #316, opened 2026-05-27
[fix] skill_usage_flush: anchor buffer glob to 8-digit date prefix

### LINKED ISSUE #305: [bug] skill_usage_flush: buffer glob '*-<agent_id>.txt' can flush another agent's buffer
## Product framing

- **Problem.** A maintainer relying on per-subagent skill-usage telemetry can see one subagent's skills silently attributed to — or consumed by — a different subagent, so the orchestrator transcript misreports which agent invoked which skills.
- **Value.** Anyone using the subagent skill-usage telemetry (#227) gets accurate, non-cross-contaminated attribution.
- **Acceptance criteria.**
  - Given two subagents with `agent_id`s `foo` and `bar-foo`, when `foo`'s SubagentStop flush runs, only `*-foo.txt` buffers (not `*-bar-foo.txt`) are read and deleted.
  - The flush glob matches only `<8-digit-date>-<sanitized agent_id>.txt`.
- **Impact / Effort.** Impact: M — silent telemetry mis-attribution/loss. Effort: S — anchor the date prefix in one glob.

## Surface affected
- [x] Hook

**Specific surface:** `hooks/skill_usage_flush.sh` (SubagentStop handler)

## Claude Code version
2.1.146

## Plugin version
0.1.14

## Reproduction steps
1. Two swe-workbench-owned subagents run in the same project on the same day with `agent_id`s where one is a hyphen-suffix of the other (e.g. `foo` and `bar-foo`; `agent_id` permits `-` per the `^[A-Za-z0-9_-]+$` sanitizer at `skill_usage_flush.sh:16`).
2. Each writes `.claude/cache/skill-usage/<YYYYMMDD>-<agent_id>.txt` (`skill_usage_record.sh:39`).
3. When `foo`'s SubagentStop fires, the loop `for f in "$cache_dir"/*-"$agent_id".txt` (`skill_usage_flush.sh:32`) expands `*` to `<YYYYMMDD>-bar`, so `<YYYYMMDD>-bar-foo.txt` also matches.

## Expected behavior
Flushing agent `foo` reads/deletes only `foo`'s own buffer(s).

## Actual behavior
`bar-foo`'s buffer is also read (its skills mis-attributed to `foo`'s flush line) and deleted, so `bar-foo`'s own later flush finds nothing.

## Logs / screenshots
Fix sketch: anchor the date — `for f in "$cache_dir"/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]-"$agent_id".txt`. Distinct from #303 item #4 (`skill_usage_record.sh` excluding `.errors.log` from cleanup).

lugassawan:
## Summary
- **Bug:** `skill_usage_flush.sh` glob `*-<id>.txt` matched sibling agent buffers whose id is a hyphen-suffix of another (e.g. `bar-foo`'s buffer `<date>-bar-foo.txt` was picked up when flushing agent `foo`), causing skill mis-attribution and silent buffer deletion.
- **Fix:** anchored the glob with `[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]-<id>.txt`; the record hook always writes `<YYYYMMDD>-<id>.txt`, so every real buffer still matches.
- **Tests:** two new tests added to `TestFlushHook` — sibling isolation (AC#1) and non-date-prefix rejection (AC#2) — both red before the fix, green after.

## Test Plan
- [x] `pytest tests/test_skill_usage_telemetry.py -v` — 23/23 pass (2 new)
- [x] `pytest tests/ -v` — 876/876 pass, no regressions
- [x] `python scripts/validate.py` — all checks passed

- [x] Reproduced the bug with two agents (`foo` and `bar-foo`) before the fix
- [x] Verified `bar-foo`'s buffer survives after flushing `foo` post-fix
- [x] Midnight-straddle (`test_midnight_straddle`) still passes — both date-stamped buffers for the same agent are still matched

Closes #305

### DIFF
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -29,7 +29,9 @@ cache_dir="${CLAUDE_PROJECT_DIR:-$PWD}/.claude/cache/skill-usage"
 # Buffer may span today's and yesterday's date-stamped file (straddle midnight).
 # The 8-digit bracket prefix is intentional: without it, "*-foo.txt" greedily
 # matches a sibling agent "bar-foo"'s "<date>-bar-foo.txt" (id allows hyphens).
-# Compatible with bash 3.2+ (POSIX bracket classes, no extglob required).
+# The suffix -"$agent_id".txt is an exact match, so a longer id ("bar-foo") never
+# matches a shorter one's glob ("foo") either. Compatible with bash 3.2+ (POSIX
+# bracket classes, no extglob required).
 buffers=()
 for f in "$cache_dir"/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]-"$agent_id".txt; do
   [ -f "$f" ] && buffers+=("$f")
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -27,9 +27,11 @@ fi
 cache_dir="${CLAUDE_PROJECT_DIR:-$PWD}/.claude/cache/skill-usage"
 
 # Buffer may span today's and yesterday's date-stamped file (straddle midnight).
-# Use a glob loop so paths with spaces are handled correctly (compatible with bash 3.2+).
+# The 8-digit bracket prefix is intentional: without it, "*-foo.txt" greedily
+# matches a sibling agent "bar-foo"'s "<date>-bar-foo.txt" (id allows hyphens).
+# Compatible with bash 3.2+ (POSIX bracket classes, no extglob required).
 buffers=()
-for f in "$cache_dir"/*-"$agent_id".txt; do
+for f in "$cache_dir"/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]-"$agent_id".txt; do
   [ -f "$f" ] && buffers+=("$f")
 done
 [ "${#buffers[@]}" -eq 0 ] && { printf '{}'; exit 0; }
--- a/tests/test_skill_usage_telemetry.py
+++ b/tests/test_skill_usage_telemetry.py
@@ -4,6 +4,7 @@ Subprocess-driven so they exercise real POSIX sh logic.  Each test gets an
 isolated cache dir via tmp_path + CLAUDE_PROJECT_DIR, and a minimal agents/
 tree via CLAUDE_PLUGIN_ROOT.
 """
+import datetime
 import json
 import os
 import subprocess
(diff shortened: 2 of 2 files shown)


hooks/skill_usage_flush.sh
```diff
@@ -27,9 +27,11 @@ fi
 cache_dir="${CLAUDE_PROJECT_DIR:-$PWD}/.claude/cache/skill-usage"
 
 # Buffer may span today's and yesterday's date-stamped file (straddle midnight).
-# Use a glob loop so paths with spaces are handled correctly (compatible with bash 3.2+).
+# The 8-digit bracket prefix is intentional: without it, "*-foo.txt" greedily
+# matches a sibling agent "bar-foo"'s "<date>-bar-foo.txt" (id allows hyphens).
+# Compatible with bash 3.2+ (POSIX bracket classes, no extglob required).
 buffers=()
-for f in "$cache_dir"/*-"$agent_id".txt; do
+for f in "$cache_dir"/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]-"$agent_id".txt; do
```
lugassawan, line 36: **Medium:** The comment explains the `bar-foo` → `foo` direction but leaves the inverse implicit. Adding one sentence clarifying that `-"$agent_id".txt` is a suffix anchor (exact id match) would make both directions unambiguous without touching code.

Suggested addition to the comment block:
```
# The suffix -"$agent_id".txt also ensures a longer id (e.g. "other-foo") never
# matches a shorter one's glob — the expansion is exact, not prefix-matched.
```

tests/test_skill_usage_telemetry.py
```diff
@@ -386,6 +386,56 @@ def test_midnight_straddle(self, plugin_root, cache_dir):
         assert not buf_today.exists()
         assert not buf_yesterday.exists()
 
+    def test_sibling_agent_buffer_not_flushed(self, plugin_root, cache_dir):
+        """agent 'foo' flush must not touch sibling agent 'bar-foo' whose id ends in '-foo'."""
+        import datetime
```
lugassawan: **Low:** `import datetime` is inside each test body rather than at the module top. The existing module-level imports (`json`, `os`, `subprocess`, etc.) are all stdlib — move this one up for consistency.

```suggestion
import datetime
```
(at the module-level import block, e.g. after `import os`). Tests are correct either way.

hooks/skill_usage_flush.sh
lugassawan: Addressed in 4cc49ed: extended the comment to clarify that the suffix `-"$agent_id".txt` is an exact match, so a longer id (e.g. `other-foo`) also cannot match a shorter one's glob.

tests/test_skill_usage_telemetry.py
lugassawan: Addressed in 4cc49ed: moved `import datetime` to the module-level import block alongside the other stdlib imports; removed the duplicated local imports from all three test bodies (`test_midnight_straddle`, `test_sibling_agent_buffer_not_flushed`, `test_date_anchor_rejects_non_date_prefix`).

PR #320, opened 2026-05-27
[ci] Add PR concurrency guard and per-job timeouts

### LINKED ISSUE #307: [chore] CI: add concurrency guard to pr.yml and per-job timeout-minutes to all workflows
## Product framing

- **Problem.** Every push to a PR (`synchronize`) starts a fresh 6-job `pr.yml` run with no cancellation of the stale run, and no workflow caps job runtime, so a hung step (lychee network stall, pip resolver, gh API) can run to GitHub's 6-hour default before being killed — wasting Actions minutes and queue time.
- **Value.** Faster feedback, lower runner cost: stale runs auto-cancel on the next push; hung jobs fail fast.
- **Acceptance criteria.**
  - `pr.yml` has a top-level `concurrency` block keyed by ref with `cancel-in-progress: true`.
  - Every job across `pr.yml`, `release.yml`, `skill-triggers.yml`, `dependabot-lockfile.yml` sets a sensible `timeout-minutes`.
  - A second push to an open PR cancels the prior in-progress run.
- **Impact / Effort.** Impact: M — wasted minutes + blocked queue. Effort: S — YAML-only.

## Type

## Surface
**Files to change:** `.github/workflows/pr.yml` (add `concurrency`); `timeout-minutes` on jobs in `pr.yml`, `release.yml`, `skill-triggers.yml`, `dependabot-lockfile.yml`.

## Description
Current: `pr.yml` triggers on `[opened, edited, synchronize]` (line 5) with no `concurrency:` key (only `dependabot-lockfile.yml` has one); no workflow sets `timeout-minutes` (grep returns none across all four). Desired: add `concurrency: { group: pr-${{ github.workflow }}-${{ github.ref }}, cancel-in-progress: true }` to `pr.yml`, and per-job `timeout-minutes` (e.g. 10) everywhere. This is run-management hardening — distinct from #303#3 (`.python-version` pin) and #230 (dependency pinning).

## Acceptance
- [ ] `pr.yml` concurrency guard cancels stale runs on re-push
- [ ] All jobs in all four workflows have `timeout-minutes`

lugassawan:
## Summary
- Add top-level `concurrency` block to `pr.yml` keyed on `github.ref` with `cancel-in-progress: true`, so a second push to an open PR auto-cancels the stale in-progress run.
- Add `timeout-minutes` to every job across all four workflow files (`pr.yml`, `release.yml`, `skill-triggers.yml`, `dependabot-lockfile.yml`) — 10 jobs total — so hung steps (e.g. `lychee` network requests in `markdown-links`) fail fast instead of running to GitHub's 6-hour default.
- Add `tests/test_workflow_hardening.py` — dependency-free TDD regression guard (stdlib only: `re`, `pathlib`, `pytest`) with two functions: `test_every_job_has_timeout_minutes` (parametrized over all four workflow files) and `test_pr_yml_has_concurrency_guard`.

## Test Plan
- [ ] `pytest tests/test_workflow_hardening.py -v` — 5 tests pass (was red before YAML edits, green after)
- [ ] `pytest tests/ -v` — 891 tests pass, no regressions

### Type-tailored checks ([ci])
- [ ] Workflow YAML parses without errors on push (GitHub Actions runner validates YAML structure on run)
- [ ] `pr.yml` concurrency block is at top-level (0-indent), `cancel-in-progress: true` confirmed
- [ ] All 10 jobs have `timeout-minutes` at 4-space indent (job-level property)
- [ ] A second push to this PR cancels the prior in-progress run (visible in Actions tab as "Cancelled")

Closes #307

### DIFF
--- a/tests/test_workflow_hardening.py
+++ b/tests/test_workflow_hardening.py
@@ -1,0 +1,72 @@
+"""Guard: every CI job must declare timeout-minutes; pr.yml must have a concurrency guard.
+
+Prevents hung jobs (especially lychee's live network requests in markdown-links) from
+consuming GitHub Actions quota up to the 6-hour default, and ensures stale in-progress
+runs are cancelled when a new push arrives on an open PR.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+ROOT = Path(__file__).parent.parent
+
+WORKFLOW_FILES = [
+    "pr.yml",
+    "release.yml",
+    "skill-triggers.yml",
+    "dependabot-lockfile.yml",
+]
+
+# Anchors to the start of the jobs: section; everything else (on:, permissions:) is ignored.
+JOBS_SECTION_RE = re.compile(r"^jobs:\s*$", re.MULTILINE)
+
+# Matches job-name keys at exactly 2-space indent with no inline value, e.g. "  smoke:"
+JOB_KEY_RE = re.compile(r"^  (\S[^:]*):\s*$", re.MULTILINE)
+
+# Matches timeout-minutes at exactly 4-space indent (direct job property), e.g. "    timeout-minutes: 15"
+TIMEOUT_RE = re.compile(r"^    timeout-minutes:\s*\d+\s*$", re.MULTILINE)
+
+
+@pytest.mark.parametrize("filename", WORKFLOW_FILES)
+def test_every_job_has_timeout_minutes(filename):
+    text = (ROOT / ".github" / "workflows" / filename).read_text()
+
+    jobs_match = JOBS_SECTION_RE.search(text)
+    assert jobs_match, f"{filename}: no 'jobs:' section found"
+    # Scope all subsequent scanning to the jobs: block only, so trigger keys in
+    # the on: block (pull_request, schedule, etc.) are not mistaken for job names.
+    jobs_section = text[jobs_match.end():]
+
+    job_matches = list(JOB_KEY_RE.finditer(jobs_section))
+    assert job_matches, f"{filename}: no job keys found under 'jobs:' — check indentation regex"
+
+    missing = []
+    for i, match in enumerate(job_matches):
+        job_name = match.group(1)
+        start = match.end()
+        end = job_matches[i + 1].start() if i + 1 < len(job_matches) else len(jobs_section)
+        segment = jobs_section[start:end]
+        if not TIMEOUT_RE.search(segment):
+            missing.append(job_name)
+
+    assert not missing, (
+        f"{filename}: job(s) missing 'timeout-minutes': {missing}"
+    )
+
+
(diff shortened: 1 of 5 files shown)


tests/test_workflow_hardening.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+ROOT = Path(__file__).parent.parent
+
+WORKFLOW_FILES = [
```
lugassawan, line 16: **Medium: Static `WORKFLOW_FILES` list silently misses future workflow additions**

A developer adding a fifth workflow (e.g. `nightly.yml`) without `timeout-minutes` would not be caught here because the file is absent from the parametrize list.

Replace with a glob so new files are covered automatically:

```python
import glob

WORKFLOW_FILES = [
    Path(p).name
    for p in sorted(glob.glob(str(ROOT / ".github" / "workflows" / "*.yml")))
]
```

If some files should be exempt, maintain an opt-out set rather than an opt-in list.

tests/test_workflow_hardening.py
```diff
@@ -0,0 +1,72 @@
...
+]
+
+# Anchors to the start of the jobs: section; everything else (on:, permissions:) is ignored.
+JOBS_SECTION_RE = re.compile(r"^jobs:\s*$", re.MULTILINE)
+
+# Matches job-name keys at exactly 2-space indent with no inline value, e.g. "  smoke:"
+JOB_KEY_RE = re.compile(r"^  (\S[^:]*):\s*$", re.MULTILINE)
+
+# Matches timeout-minutes at exactly 4-space indent (direct job property), e.g. "    timeout-minutes: 15"
+TIMEOUT_RE = re.compile(r"^    timeout-minutes:\s*\d+\s*$", re.MULTILINE)
```
lugassawan: **Low: `TIMEOUT_RE` accepts `timeout-minutes: 0`**

`0` is a valid digit sequence but disables the timeout entirely on GitHub Actions (zero means no limit). A typo or merge-conflict residue could pass this guard silently.

Tighten to require a positive integer:

```python
TIMEOUT_RE = re.compile(r"^    timeout-minutes:\s*[1-9]\d*\s*$", re.MULTILINE)
```

tests/test_workflow_hardening.py
lugassawan: Addressed in 9e51ce7: replaced static list with `glob.glob('.github/workflows/*.yml')` so future workflow additions are automatically covered.

tests/test_workflow_hardening.py
lugassawan: Addressed in 9e51ce7: tightened `TIMEOUT_RE` to `[1-9]\d*` — rejects `0` which disables the timeout on GitHub Actions.

PR #321, opened 2026-05-27
[chore] Add tests for lock-pip-requirements.sh --check gate

### LINKED ISSUE #308: [chore] Add tests for lock-pip-requirements.sh lockfile-drift gate
## Product framing

- **Problem.** The lockfile-drift gate is untrustworthy: `scripts/lock-pip-requirements.sh` is invoked by `pr.yml`'s `lockfile-check` job and by `dependabot-lockfile.yml`, but no test exercises its `--check` diff logic, so a regression in its header-stripping comparison could silently pass a stale or tampered lockfile.
- **Value.** The supply-chain lockfile gate gets a regression net; future edits to the comparison logic can't quietly break drift detection.
- **Acceptance criteria.**
  - A new `tests/test_lock_pip_requirements.py` asserts `--check` exits non-zero when the committed lockfile drifts from a freshly-resolved one.
  - It asserts `--check` exits 0 when in sync.
  - It covers the `--help` / unknown-arg path.
- **Impact / Effort.** Impact: M — a silently-broken supply-chain gate. Effort: M.

## Type

## Surface
**Files to change:** add `tests/test_lock_pip_requirements.py`; subject under test `scripts/lock-pip-requirements.sh`.

## Description
Current: `grep -rl lock-pip tests/` returns nothing — zero coverage — though `pr.yml` `lockfile-check` and `dependabot-lockfile.yml` both depend on the script's `--check` and regeneration paths; the header-stripping `grep -v '^#'` comparison is non-trivial and easy to regress. Desired: a focused pytest module (mirroring `test_release_sh_*` / `test_setup_sh` style) that drives the script in a temp dir and asserts drift / in-sync / `--help` behavior. Distinct from #303#7 (which adds `--require-hashes` to the pip-tools *bootstrap* — a security change, not test coverage).

## Acceptance
- [ ] `--check` exits non-zero on drift
- [ ] `--check` exits 0 when in sync
- [ ] `--help` / unknown-arg path covered

lugassawan:
## Summary

- Extended `tests/test_lock_pip_requirements.py` with two behavioral test classes (`TestArgParsing`, `TestCheckDriftGate`); existing static guard classes are untouched
- `TestArgParsing`: drives the real script via subprocess for `--help` / `-h` / unknown-arg — no stubs needed (arg loop exits before Python 3.12 preflight)
- `TestCheckDriftGate`: scaffolds a hermetic temp repo with a stubbed `python3.12` that creates a fake venv; exercises the `compile()` function's `grep-strip + diff` comparison (script lines 77–82) with three scenarios: in-sync → exit 0, drift (extra dep line) → exit non-zero, and header/blank-only diff → exit 0 (false-positive guard)

## Test Plan

- [x] `python3 -m pytest tests/test_lock_pip_requirements.py -v` — all 14 tests pass (8 pre-existing + 6 new)
- [x] `python3 -m pytest tests/ -v` — full 892-test suite green, zero regressions
- [x] `bash scripts/validate.sh` — plugin validation passes, production files unmodified

- [x] No production code changed; `scripts/lock-pip-requirements.sh` is untouched
- [x] Stubs are generated as runtime strings (not committed `.sh`), so `shellcheck`/`validate` are unaffected
- [x] `STUB_DIR` passed as env var (not embedded as literal) to avoid single-quote fragility in stub path

Closes #308


tests/test_lock_pip_requirements.py
```diff
@@ -63,3 +162,113 @@ def test_lock_script_compiles_build_requirements(self):
...
+        """Fresh compile output matches committed lockfiles → exit 0."""
+        paths = _make_repo(tmp_path)
+        result = subprocess.run(
+            ["bash", str(paths["script"]), "--check"],
+            capture_output=True,
+            text=True,
+            env=self._env(paths),
+        )
+        assert result.returncode == 0, result.stderr
+        assert "up to date" in result.stdout
```
lugassawan, line 235: **Medium** — `assert "up to date" in result.stdout` only checks that the phrase appears at least once. The script emits this message once per compiled file (three total). If one of the three compile calls fails silently or the stdout format changes, the assertion still passes.

Consider also asserting the count:
```python
assert result.stdout.count("up to date") == 3
```

tests/test_lock_pip_requirements.py
```diff
@@ -1,17 +1,116 @@
...
+        "#!/usr/bin/env bash",
+        "output='' src=''",
+        "while [[ $# -gt 0 ]]; do",
+        "    case \"$1\" in",
+        "        --output-file) output=\"$2\"; shift 2 ;;",
+        "        --*) shift ;;",
+        "        *) src=\"$1\"; shift ;;",
+        "    esac",
+        "done",
+        'lock="${src%.txt}.lock"',
```
lugassawan, line 83: **Medium** — The pip-compile stub resolves the lock path as `${src%.txt}.lock` where `$src` is a relative path (`tests/requirements.txt`). This only resolves correctly when CWD is the repo root — which the real script guarantees via `(cd "$REPO_ROOT" && ...)`. But if the venv's mktemp directory were ever set as CWD (e.g. by a future script change), the stub would silently produce an empty output, causing the in-sync test to pass for the wrong reason (empty == empty).

Currently the `cd "$REPO_ROOT"` call in the script guarantees correctness, so this is a latent risk rather than an active bug. Worth a comment noting the CWD dependency so a future script change doesn't break the stub silently.

tests/test_lock_pip_requirements.py
```diff
@@ -63,3 +162,113 @@ def test_lock_script_compiles_build_requirements(self):
...
+
+class TestCheckDriftGate:
+    """Behavioral tests for the --check lockfile-drift gate.
+
+    Each test scaffolds a hermetic temp repo (no network, no Python 3.12).
+    The stub python3.12 creates a fake venv so the script's compile() function
+    runs fully, exercising the grep-strip + diff comparison at lines 77–82.
+    """
+
+    def _env(self, paths: dict, *, drift: bool = False) -> dict:
```
lugassawan, line 217: **Low** — `_env(paths, drift=False)` omits `DRIFT_TEST` from the returned dict, relying on the inherited environment to not have it set. If a developer runs tests with `DRIFT_TEST=1` in their shell, the non-drift tests will fail confusingly.

Explicit opt-out is safer:
```python
"DRIFT_TEST": "1" if drift else "0",
```

tests/test_lock_pip_requirements.py
```diff
@@ -63,3 +162,113 @@ def test_lock_script_compiles_build_requirements(self):
...
+    def test_check_exits_nonzero_on_drift(self, tmp_path):
+        """Fresh compile has an extra dependency line → exit non-zero."""
+        paths = _make_repo(tmp_path)
+        result = subprocess.run(
+            ["bash", str(paths["script"]), "--check"],
+            capture_output=True,
+            text=True,
+            env=self._env(paths, drift=True),
+        )
+        assert result.returncode != 0
```
lugassawan: **Low** — `assert result.returncode != 0` accepts any non-zero exit code, including 127 (command not found) or 2 (bash syntax error), which would mask a scaffold failure as a passing drift test.

The script's `compile()` returns 1 on drift, and `set -euo pipefail` propagates that as exit code 1. Consider tightening to:
```python
assert result.returncode == 1
```

tests/test_lock_pip_requirements.py
lugassawan: Addressed in c8a80648dcbdacf3c97eb71153731e10c16a58be: added `assert result.stdout.count("up to date") == 3` to verify all three compile calls report success.

tests/test_lock_pip_requirements.py
lugassawan: Addressed in c8a80648dcbdacf3c97eb71153731e10c16a58be: added an inline comment documenting that the stub's lock path resolution requires CWD to be the repo root (guaranteed by the script's `cd "$REPO_ROOT"` call).

tests/test_lock_pip_requirements.py
lugassawan: Addressed in c8a80648dcbdacf3c97eb71153731e10c16a58be: `_env()` now always emits `"DRIFT_TEST": "1" if drift else "0"` — explicit opt-out prevents ambient env var from contaminating non-drift tests.

tests/test_lock_pip_requirements.py
lugassawan: Addressed in c8a80648dcbdacf3c97eb71153731e10c16a58be: tightened to `assert result.returncode == 1` — rejects scaffold errors (exit 127, exit 2) that would otherwise pass silently.

PR #325, opened 2026-05-29
[fix] Unblock ephemeral worktree cleanup from the bash guard

### LINKED ISSUE #324: [bug] Ephemeral worktree cleanup blocked by bash guard when rimba places worktrees under home
## Summary

When `/review` (PR review) and `/address-feedback` run their cleanup after a session, the fallback `rm -rf "$WT"` is blocked by the `PreToolUse:Bash` guard with exit 2 + `BLOCKED: destructive rm against root or home`.

## Root cause

`hooks/bash_guard.sh` blocks any `rm -r*f*` whose target matches `/Users/<user>/…` (or `/home/…`). rimba places worktrees under the home tree (`/Users/<user>/dir-worktrees/`), so the fallback cleanup command is killed.

## Repro

1. Configure rimba to place worktrees under `~/dir-worktrees/` (default).
2. Run `/review <PR>` or `/address-feedback <PR>`.
3. On cleanup, when `rimba remove` fails, the skill runs `rm -rf "$WT"` where `$WT = /Users/<user>/dir-worktrees/pr-review-N`.
4. The guard blocks it. The worktree directory is left behind.

## Expected

The ephemeral worktree is cleaned up silently. Raw `rm -rf /Users/...` (non-worktree targets) remains blocked.

## Fix

Add `scripts/clean-ephemeral.sh` — a self-validating helper that proves a path is a registered git linked worktree with an ephemeral basename before removing it. Route the three cleanup fallbacks through it instead of a bare `rm -rf`.

lugassawan:
## Summary

- Add `scripts/clean-ephemeral.sh`: a self-validating helper that removes an ephemeral worktree path only after proving it is (a) a registered git linked worktree with an ephemeral-task basename, OR (b) under a known safe tmp/share directory. The guard (`hooks/bash_guard.sh`) is left unchanged.
- Route the three review/feedback cleanup fallbacks through the helper instead of bare `rm -rf "$WT"`, which the bash guard blocks when rimba places worktrees under the home tree (`~/dir-worktrees/`).
- Add `tests/test_clean_ephemeral.py` (16 cases covering happy path, rejection of dangerous/non-ephemeral paths, submodule `.git` files, traversal attempts).
- Extend `tests/test_workflow_pr_review_skill.py` and `tests/test_workflow_address_feedback_skill.py` with call-site assertions that enforce the new invariant.

## Test Plan

- [x] `shellcheck scripts/clean-ephemeral.sh` — 0 issues
- [x] `python3 -m pytest tests/ -q` — 927/927 passed (including 16 new cases in `test_clean_ephemeral.py`)
- [x] `bash scripts/validate.sh` — All checks passed
- [x] Pre-push hook ran full test suite — 927 passed
- [x] Manual verify: `bash scripts/clean-ephemeral.sh "$HOME/Documents"` exits 1, deletes nothing; raw `rm -rf $HOME/Documents` still blocked by guard (`test_hooks.py` unchanged)
- [x] Manual verify: in a temp git repo, `git worktree add /tmp/clean-test/pr-review-9 HEAD && bash scripts/clean-ephemeral.sh /tmp/clean-test/pr-review-9` exits 0 and removes the worktree

- [ ] Reproduce original bug: rimba worktree under `~/dir-worktrees/pr-review-N`; verify cleanup no longer exits 2
- [ ] Confirm raw destructive `rm -rf /Users/...` still blocked (run `pytest tests/test_hooks.py -q`)

Closes #324


scripts/clean-ephemeral.sh
```diff
@@ -0,0 +1,125 @@
...
+# clean-ephemeral.sh <path>
+#
+# Remove an ephemeral git worktree directory, but only after proving it is a
+# sanctioned artifact.  Exits 1 (never silently) when any safety check fails.
+#
+# Safety checks (ALL must pass):
+#   1. Argument is non-empty, absolute, and contains no ".." segment.
+#   2. Path is not "/", not "$HOME", and is at least two levels deep.
+#   3. Basename matches the ephemeral-task prefix regex used by the three flows:
+#        ^(pr-review|pr-review-followup|address-feedback)-[A-Za-z0-9._-]+$
```
lugassawan: **Medium** — Header comment says `pr-review-followup` but `basename_is_ephemeral()` uses `pr-followup`. The rimba task label in `workflow-pr-review-followup/SKILL.md` is `pr-followup-$PR`, so the comment is misleading. A path named `pr-review-followup-42` would be rejected while the header promises it passes.

```suggestion
#        ^(pr-review|pr-followup|address-feedback)-[A-Za-z0-9._-]+$
```

skills/workflow-address-feedback/SKILL.md
```diff
@@ -241,7 +241,7 @@ else
     echo "Cleaned up worktree address-feedback-$PR."
   else
     # $WT is set in Phase 2 (both rimba and fallback paths); do not re-assign here
-    git worktree remove --force "$WT" 2>/dev/null && rm -rf "$WT" 2>/dev/null
+    git worktree remove --force "$WT" 2>/dev/null && bash "${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}/scripts/clean-ephemeral.sh" "$WT" 2>/dev/null
```
lugassawan: **Medium** — `&&` gates `clean-ephemeral.sh` on `git worktree remove` succeeding. If `git worktree remove --force` fails (busy, permission, unregistered), the directory is left behind silently — exactly the failure scenario this PR is fixing. The original `rm -rf` also had this behaviour, but since we're touching the line, fix it:

```suggestion
    git worktree remove --force "$WT" 2>/dev/null; bash "${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}/scripts/clean-ephemeral.sh" "$WT" 2>/dev/null
```

scripts/clean-ephemeral.sh
```diff
@@ -0,0 +1,125 @@
...
+[ -n "$TARGET" ] || reject "path argument is required"
+
+# Must be absolute.
+case "$TARGET" in
+  /*) ;;
+  *) reject "path must be absolute: $TARGET" ;;
+esac
+
+# Must not contain ".." segments (e.g. /foo/../bar, or a bare ../ prefix).
+if echo "$TARGET" | grep -qE '(^|/)\.\.(\/|$)'; then
```
lugassawan: **Low** — `echo` with a value starting with `-` can be misinterpreted as a flag on some shells. `printf '%s'` is already used at line 57 for consistency.

```suggestion
if printf '%s' "$TARGET" | grep -qE '(^|/)\.\.(\\/|$)'; then
```

PR #326, opened 2026-05-29
[chore] Machine-enforce _CLEAN_ENV rule for test subprocess calls

lugassawan:
## Summary

- Adds `_TEST_ENV_LEAK_RE` regex + `check_test_subprocess_env()` to `scripts/validate.py` — statically flags any `tests/*.py` file (excluding `conftest.py` and `test_validate.py`) that passes `env=os.environ` or `env={**os.environ` to `subprocess.run`. Wired into `main()` after `check_hook_scripts()`.
- Adds `_GIT_DIR_GUARD_SENTINEL` + session-scoped autouse fixture `_git_dir_leak_guard` to `tests/conftest.py` — patches `subprocess.run` to raise `AssertionError` if `GIT_DIR` appears in the effective subprocess env; injects a sentinel `GIT_DIR` so the guard fires standalone (not just under the pre-push hook). Covers `subprocess.check_output` automatically; `call`/`check_call` noted as out-of-scope in the docstring.
- Fixes pre-existing bare `subprocess.run()` in `test_skill_usage_telemetry.py:TestHooksJson` (caught by the new runtime guard — adds `env=dict(_CLEAN_ENV)`).
- Adds 10 new tests: `TestCheckTestSubprocessEnv` (6 cases) in `tests/test_validate.py` and `TestGitDirLeakGuard` (4 cases) in new `tests/test_conftest_guard.py`.

## Test Plan

- [x] `python3 -m pytest tests/ -q` — 937/937 pass (includes 10 new cases)
- [x] Pre-push hook ran and passed on push (937 tests, validate passed)
- [x] Writing a temp file with `env=os.environ` in `tests/` → `pytest` raises `AssertionError` from the runtime guard immediately; `validate.sh` also fails with the leak message
- [x] `conftest.py` and `test_validate.py` are correctly exempt from the static check
- [x] `test_conftest_guard.py` env patterns (`env={"GIT_DIR": "/x"}`, `env=dict(_CLEAN_ENV)`) do not trigger the static checker

Issue: N/A — machine-enforces the `_CLEAN_ENV` invariant already documented at `tests/README.md:45-47` (belt-and-suspenders against a recurrence of the `{**os.environ}` env-leak confirmed in PR #325)

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -518,7 +518,9 @@ def check_unwired_principle_skills(cache=None):
 # Python hook syntax check
 # ──────────────────────────────────────────────
 
-_TEST_ENV_LEAK_RE = re.compile(r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|os\.environ)\b')
+_TEST_ENV_LEAK_RE = re.compile(
+    r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|dict\s*\(\s*os\.environ\s*\)|os\.environ\b)'
+)
 
 
 _TEST_ENV_EXEMPT = frozenset({"conftest.py", "test_validate.py"})
@@ -530,7 +532,7 @@ def check_test_subprocess_env():
     conftest.py is exempt (defines _CLEAN_ENV and the runtime guard).
     test_validate.py is exempt (contains bad-pattern strings as test fixture data)."""
     tests_dir = ROOT / "tests"
-    for py_file in sorted(tests_dir.glob("*.py")):
+    for py_file in sorted(tests_dir.glob("*.py")):  # intentionally flat — subdirs not scanned
         if py_file.name in _TEST_ENV_EXEMPT:
             continue
         try:
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -521,13 +521,17 @@ def check_unwired_principle_skills(cache=None):
 _TEST_ENV_LEAK_RE = re.compile(r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|os\.environ)\b')
 
 
+_TEST_ENV_EXEMPT = frozenset({"conftest.py", "test_validate.py"})
+
+
 def check_test_subprocess_env():
     """tests/*.py subprocess sites must not pass the raw parent env.
     Use env=dict(_CLEAN_ENV) or env={**_CLEAN_ENV, ...}. See tests/README.md.
-    conftest.py is exempt (defines _CLEAN_ENV and the runtime guard)."""
+    conftest.py is exempt (defines _CLEAN_ENV and the runtime guard).
+    test_validate.py is exempt (contains bad-pattern strings as test fixture data)."""
     tests_dir = ROOT / "tests"
     for py_file in sorted(tests_dir.glob("*.py")):
-        if py_file.name == "conftest.py":
+        if py_file.name in _TEST_ENV_EXEMPT:
             continue
         try:
             text = py_file.read_text(encoding="utf-8")
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -518,6 +518,33 @@ def check_unwired_principle_skills(cache=None):
 # Python hook syntax check
 # ──────────────────────────────────────────────
 
+_TEST_ENV_LEAK_RE = re.compile(r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|os\.environ)\b')
+
+
+def check_test_subprocess_env():
+    """tests/*.py subprocess sites must not pass the raw parent env.
+    Use env=dict(_CLEAN_ENV) or env={**_CLEAN_ENV, ...}. See tests/README.md.
+    conftest.py is exempt (defines _CLEAN_ENV and the runtime guard)."""
+    tests_dir = ROOT / "tests"
+    for py_file in sorted(tests_dir.glob("*.py")):
+        if py_file.name == "conftest.py":
+            continue
+        try:
+            text = py_file.read_text(encoding="utf-8")
+        except OSError as e:
+            fail(py_file.relative_to(ROOT), f"could not read file: {e}")
+            continue
+        for i, line in enumerate(text.splitlines(), 1):
+            if _TEST_ENV_LEAK_RE.search(line):
+                fail(
+                    py_file.relative_to(ROOT),
+                    f"line {i}: subprocess env= leaks the parent environment "
+                    f"(GIT_DIR leaks into git children under the pre-push hook). "
+                    f"Use env=dict(_CLEAN_ENV) or env={{**_CLEAN_ENV, ...}}. "
+                    f"See tests/README.md.",
+                )
+
+
 def check_hook_scripts():
     hooks_dir = ROOT / "hooks"
     for py_file in sorted(hooks_dir.glob("*.py")):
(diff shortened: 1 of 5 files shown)


scripts/validate.py
```diff
@@ -518,6 +518,37 @@ def check_unwired_principle_skills(cache=None):
 # Python hook syntax check
 # ──────────────────────────────────────────────
 
+_TEST_ENV_LEAK_RE = re.compile(r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|os\.environ)\b')
```
lugassawan: **High:** `_TEST_ENV_LEAK_RE` does not match `env=dict(os.environ)`.

`dict(os.environ)` produces a mutable dict containing all env vars including `GIT_DIR` — semantically equivalent to `os.environ` itself. The static checker silently passes it; only the runtime guard catches it at test time.

Extend the alternation:
```python
_TEST_ENV_LEAK_RE = re.compile(
    r"\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|dict\s*\(\s*os\.environ\s*\)|os\.environ)\b"
)
```
Also add `test_env_dict_os_environ_flagged` to `TestCheckTestSubprocessEnv`.

tests/conftest.py
```diff
@@ -29,6 +30,39 @@
...
+def _git_dir_leak_guard():
+    """Fail loudly if any subprocess.run receives GIT_DIR in its env.
+    Injects a sentinel GIT_DIR so the guard fires standalone (not just under the hook).
+    See tests/README.md for the _CLEAN_ENV pattern.
+    Coverage note: patching subprocess.run also covers subprocess.check_output (which
+    delegates to run internally). subprocess.call / check_call go directly to Popen and
+    are NOT covered — avoid them in tests."""
+    _orig_run = subprocess.run
+
+    def _guarded_run(*args, **kwargs):
```
lugassawan, line 46: **Medium:** The coverage note says patching `subprocess.run` "covers `subprocess.check_output`" — this is a CPython implementation detail (CPython's `check_output` calls `run` via the module global, so the monkeypatch intercepts it). Mark it explicitly to avoid confusion on non-CPython runtimes:

```suggestion
    Coverage note: patching subprocess.run also covers subprocess.check_output (which
    delegates to run on CPython — implementation detail, not guaranteed by the stdlib spec).
    subprocess.call / check_call go directly to Popen and are NOT covered — avoid them in tests.
```

tests/conftest.py
```diff
@@ -29,6 +30,39 @@
     }
 )
 
+_GIT_DIR_GUARD_SENTINEL: Final[str] = "/tmp/fake-git-dir-guard"
```
lugassawan: **Low:** `"/tmp/fake-git-dir-guard"` looks like a real path. Since the sentinel is compared as a string (never stat'd or opened), a non-path value makes the intent unambiguous:

```suggestion
_GIT_DIR_GUARD_SENTINEL: Final[str] = "__pytest_git_dir_sentinel__"
```

scripts/validate.py
```diff
@@ -518,6 +518,37 @@ def check_unwired_principle_skills(cache=None):
...
+_TEST_ENV_EXEMPT = frozenset({"conftest.py", "test_validate.py"})
+
+
+def check_test_subprocess_env():
+    """tests/*.py subprocess sites must not pass the raw parent env.
+    Use env=dict(_CLEAN_ENV) or env={**_CLEAN_ENV, ...}. See tests/README.md.
+    conftest.py is exempt (defines _CLEAN_ENV and the runtime guard).
+    test_validate.py is exempt (contains bad-pattern strings as test fixture data)."""
+    tests_dir = ROOT / "tests"
+    for py_file in sorted(tests_dir.glob("*.py")):
```
lugassawan: **Low:** `glob("*.py")` silently skips any `tests/subdir/` helpers added in the future. Add a brief comment to make the depth-limit intentional:

```suggestion
        for py_file in sorted(tests_dir.glob("*.py")):  # intentionally flat — subdirs not scanned
```

tests/test_conftest_guard.py
```diff
@@ -0,0 +1,29 @@
...
+        with pytest.raises(AssertionError, match="GIT_DIR"):
+            subprocess.run(["true"], env={"GIT_DIR": "/x"})
+
+    def test_guard_blocks_git_dir_inherited(self):
+        """subprocess.run with no env= inherits process env; sentinel makes guard fire."""
+        with pytest.raises(AssertionError, match="GIT_DIR"):
+            subprocess.run(["true"])
+
+    def test_guard_allows_clean_env(self):
+        """subprocess.run with env=dict(_CLEAN_ENV) succeeds (no GIT_DIR in _CLEAN_ENV)."""
```
lugassawan, line 23: **Low:** The guard's allowance path is only exercised via `test_guard_allows_clean_env` using `_CLEAN_ENV`. Consider adding a `test_guard_allows_arbitrary_clean_env` that passes a minimal dict without `GIT_DIR` (e.g. `env={"PATH": "/usr/bin"}`) to confirm the guard does not over-block non-`_CLEAN_ENV` envs.

scripts/validate.py
lugassawan: Addressed in f639cce: extended regex to cover `env=dict(os.environ)`, moved `\b` inside the `os\.environ` alternation only (dict/splat end with non-word chars). Added `test_env_dict_os_environ_flagged`.

tests/conftest.py
lugassawan: Addressed in f639cce: docstring now reads 'delegates to run on CPython — implementation detail, not guaranteed by the stdlib spec'.

tests/conftest.py
lugassawan: Addressed in f639cce: renamed to `__pytest_git_dir_sentinel__`.

scripts/validate.py
lugassawan: Addressed in f639cce: added `# intentionally flat — subdirs not scanned` comment.

tests/test_conftest_guard.py
lugassawan: Addressed in f639cce: added `test_guard_allows_arbitrary_clean_env` (passes `env={"PATH": "/usr/bin"}`).

scripts/validate.py
```diff
@@ -518,6 +518,39 @@ def check_unwired_principle_skills(cache=None):
 # Python hook syntax check
 # ──────────────────────────────────────────────
 
+_TEST_ENV_LEAK_RE = re.compile(
+    r'\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|dict\s*\(\s*os\.environ\s*\)|os\.environ\b)'
```
lugassawan: **Low (informational):** `os\.environ\b` in the last alternation also matches `os.environ.copy()` and `os.environ.items()`. `os.environ.copy()` is arguably still worth catching (it copies all vars including `GIT_DIR`), but `os.environ.items()` is read-only iteration and would be a false positive if it appeared as `env=os.environ.items()`. To exclude method-call forms:

```suggestion
_TEST_ENV_LEAK_RE = re.compile(
    r"\benv\s*=\s*(?:\{\s*\*\*\s*os\.environ|dict\s*\(\s*os\.environ\s*\)|os\.environ(?!\s*\.))"
)
```

Not blocking — no current occurrences in the repo, and the runtime guard still catches `os.environ.copy()` at test time.

lugassawan: Addressed in 2e6f7a7: replaced trailing \b with negative lookahead `(?!\s*\.)` so `os.environ.copy()` and `os.environ.items()` are no longer matched. Added `test_env_method_call_not_flagged` to cover both cases.

PR #379, opened 2026-05-29
[fix] Drive language/principle skill selection deterministically

lugassawan:
## Summary
- **Root cause:** Only 2 of 17 subagents included `@./shared/languages.md`; the rest never loaded language-specific idioms. The docs advertised auto-loading as automatic, but no mechanism ever built that capability.
- **Subagents (deterministic):** Added `@./shared/languages.md` to all 16 code-touching agents and a mandatory `language-*` gate paragraph that requires agents to identify the language in scope and invoke the matching skill — not "when relevant" but required.
- **Main session (non-blocking hint):** New `PostToolUse:Read|Edit|Write` hook emits a hint when a code file is opened, mapping the extension to the correct `language-*` skill. De-duped per session/skill via a date-scoped sentinel dir. Always exits 0.
- **Structural gate:** `validate.py` now requires code-touching agents (those with `@./shared/principles.md`) to also include `@./shared/languages.md`. The `product-manager` agent is whitelisted. Error message names `_NON_CODE_AGENTS` for discoverability.
- **Docs:** `auto-load` → `auto-hint` in README.md and docs/catalog.md to stop advertising a capability that wasn't built.
- **Bonus:** Added `Auto-load when` clause to `principle-communication` and `principle-data-modeling` SKILL.md descriptions.

## Test Plan
- [x] Changed skills/commands/agents load without errors (`python scripts/validate.py` — 0 issues)
- [x] 1029/1029 pytest tests pass after rebase on main
- [x] T1 (languages.md in all 16 agents), T2 (language-* gate in all 16), T3 (hook wired, non-blocking, de-duplication, adversarial inputs), T4 (Auto-load when in all principle-*/language-* SKILL.md) all green
- [x] Hook tested manually: emits JSON hint on first call, silent on second (same session+skill)

- [x] Reproduce: before patch, `python -m pytest tests/test_agent_language_catalog.py --tb=no -q` shows 35 failures; after patch, 0 failures
- [x] Verify: `hooks/skill_autoload_hint.sh` exits 0 for all adversarial payloads (empty stdin, malformed JSON, missing file_path, no extension)

Issue: N/A — addresses user-reported language/principle skill under-selection; no filed issue exists yet

### DIFF
--- a/hooks/skill_autoload_hint.sh
+++ b/hooks/skill_autoload_hint.sh
@@ -62,7 +62,7 @@ main() {
 
     # Already hinted this session for this skill → silent no-op
     [ -f "$sentinel" ] && exit 0
-    touch "$sentinel" 2>/dev/null || true
+    touch "$sentinel" 2>/dev/null || exit 0
 
     # Emit hint via hookSpecificOutput
     local hint="Consider \`swe-workbench:${skill}\` for .${ext} work — invoke it via the Skill tool to apply language-specific idioms."
--- a/hooks/skill_autoload_hint.sh
+++ b/hooks/skill_autoload_hint.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# PostToolUse:Read|Edit|Write handler — emits a non-blocking hint when the
+# touched file's extension maps to a known language-* skill.
+#
+# Behaviour:
+#   • De-duplicates: emits at most once per (session, skill) pair.
+#   • Never blocks: exit 0 always.  A broken hook must never gate a tool call.
+set -u
+
+# ── extension → skill map ────────────────────────────────────────────────────
+ext_to_skill() {
+    local ext="$1"
+    case "$ext" in
+        py)                   echo "language-python" ;;
+        ts|tsx|js|jsx|mjs|cjs) echo "language-typescript" ;;
+        go)                   echo "language-go" ;;
+        rs)                   echo "language-rust" ;;
+        java)                 echo "language-java" ;;
+        kt|kts)               echo "language-kotlin" ;;
+        rb)                   echo "language-ruby" ;;
+        swift)                echo "language-swift" ;;
+        sh|bash)              echo "language-bash" ;;
+        sql)                  echo "language-sql" ;;
+        cs)                   echo "language-csharp" ;;
+        *)                    echo "" ;;
+    esac
+}
+
+main() {
+    local input file_path session_id ext skill safe_session safe_skill sentinel dir today
+
+    input=$(cat)
+
+    # Extract file path (Read/Edit supply file_path; Write also supplies it)
+    file_path=$(printf '%s' "$input" | jq -r '.tool_input.file_path // empty' 2>/dev/null) || exit 0
+    [ -n "$file_path" ] || exit 0
+
+    # Derive extension (lowercase, POSIX tr for Bash 3 compat).
+    # Guard: if ext equals the bare filename there was no dot → not an extension.
+    ext="${file_path##*.}"
+    ext=$(printf '%s' "$ext" | tr '[:upper:]' '[:lower:]')
+    [ "$ext" = "${file_path##*/}" ] && exit 0   # no dot in filename → no extension
+
+    skill=$(ext_to_skill "$ext")
+    [ -n "$skill" ] || exit 0             # unmapped extension → no hint
+
+    # Session-scoped de-dup sentinel.
+    # Use || true so a jq-absent environment still falls through to the PID fallback.
+    session_id=$(printf '%s' "$input" | jq -r '.session_id // empty' 2>/dev/null) || true
+    [ -n "$session_id" ] || session_id="$$"   # fallback: PID (coarse but safe)
+
+    # Sanitize both components: only alphanumeric + dash + underscore
+    safe_session="${session_id//[^A-Za-z0-9_-]/_}"
+    safe_skill="${skill//[^A-Za-z0-9_-]/_}"
+
+    # Date-scoped subdir provides a coarse TTL: yesterday's sentinels are invisible
+    # today, preventing cross-session suppression on long-running machines.
+    today=$(date +%Y%m%d 2>/dev/null) || today="default"
+    dir="${TMPDIR:-/tmp}/swe_wb_skill_hints/${today}"
+    mkdir -p "$dir" 2>/dev/null || exit 0
+    sentinel="${dir}/${safe_session}_${safe_skill}"
+
+    # Already hinted this session for this skill → silent no-op
+    [ -f "$sentinel" ] && exit 0
+    touch "$sentinel" 2>/dev/null || true
+
+    # Emit hint via hookSpecificOutput
+    local hint="Consider \`swe-workbench:${skill}\` for .${ext} work — invoke it via the Skill tool to apply language-specific idioms."
+    jq -cn --arg ctx "$hint" \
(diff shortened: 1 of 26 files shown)


tests/test_agent_language_catalog.py
```diff
@@ -0,0 +1,110 @@
...
+def _skill_dirs_with_prefix(prefix: str):
+    return [p for p in SKILLS_DIR.iterdir() if p.is_dir() and p.name.startswith(prefix)]
+
+
+@pytest.mark.parametrize(
+    "skill_dir",
+    _skill_dirs_with_prefix("principle-") + _skill_dirs_with_prefix("language-"),
+    ids=lambda p: p.name,
+)
+def test_skill_description_has_autoload_clause(skill_dir):
```
lugassawan, line 101: **Medium** — T4's full-text scan passes even if `Auto-load when` appears only in the skill body (not the `description:` frontmatter field the harness reads for triggering). Narrow to the frontmatter block:

```python
parts = text.split('---', 2)
frontmatter = parts[1] if len(parts) >= 3 else ''
assert re.search(r'auto-load when', frontmatter, re.IGNORECASE), ...
```

hooks/skill_autoload_hint.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    # Date-scoped subdir provides a coarse TTL: yesterday's sentinels are invisible
+    # today, preventing cross-session suppression on long-running machines.
+    today=$(date +%Y%m%d 2>/dev/null) || today="default"
+    dir="${TMPDIR:-/tmp}/swe_wb_skill_hints/${today}"
+    mkdir -p "$dir" 2>/dev/null || exit 0
+    sentinel="${dir}/${safe_session}_${safe_skill}"
+
+    # Already hinted this session for this skill → silent no-op
+    [ -f "$sentinel" ] && exit 0
```
lugassawan, line 64: **Low** — if `touch "\$sentinel"` fails (inode quota or read-only TMPDIR), the sentinel is never written and the hint fires on every tool call for that session+skill. Silently exit instead:

```suggestion
    touch "$sentinel" 2>/dev/null || exit 0
```

scripts/validate.py
```diff
@@ -406,7 +406,16 @@ def _expected_slice(sid):
                 fail(slice_path.relative_to(ROOT),
                      f"entry 'swe-workbench:{sid}' belongs in {_expected_slice(sid)}, not {slice_file}")
 
-    # Every agent must reference at least one slice catalog
+    # Agents that reference @./shared/principles.md but are not code-touching
+    # (e.g. product-manager files GitHub issues and never reads source).
+    # These are explicitly exempt from the languages.md co-requirement below.
+    _NON_CODE_AGENTS = frozenset({
```
lugassawan: **Low** — `_NON_CODE_AGENTS` is a function-local frozenset inside check_catalog_completeness(). A future contributor adding a non-code agent would need to know to look inside that function. Consider promoting it to module scope with a comment citing the exemption criteria.

tests/test_hooks.py
```diff
@@ -281,3 +281,124 @@ def test_hook_script_exists(self):
...
+            "skill_autoload_hint.sh" in e["hooks"][0]["command"] for e in entries
+        ), f"No entry references skill_autoload_hint.sh; entries: {entries}"
+
+    def test_hook_script_exists_and_executable(self):
+        assert self.SCRIPT.exists(), f"Missing hook script: {self.SCRIPT}"
+        assert os.access(self.SCRIPT, os.X_OK), (
+            f"Hook script not executable: {self.SCRIPT}"
+        )
+
+    def test_hook_is_nonblocking(self):
```
lugassawan: **Low** — test_hook_is_nonblocking runs without TMPDIR isolation (unlike the de-dup tests). The sentinel for test-session-abc-123 + language-python is written to the real $TMPDIR and left on CI runners. Add `env["TMPDIR"] = str(tmp_path)` (add tmp_path parameter) to isolate this test.

tests/test_agent_language_catalog.py
lugassawan: Addressed in 5d49a697388be253ada7eede1876095b9708c788: T4 frontmatter precision: narrow Auto-load when check to YAML frontmatter only.

hooks/skill_autoload_hint.sh
lugassawan: Addressed in 5d49a697388be253ada7eede1876095b9708c788: hook sentinel: touch sentinel || exit 0 prevents per-call hint spam on failed touch.

scripts/validate.py
lugassawan: Addressed in 5d49a697388be253ada7eede1876095b9708c788: _NON_CODE_AGENTS promoted to module scope with explanatory comment.

tests/test_hooks.py
lugassawan: Addressed in 5d49a697388be253ada7eede1876095b9708c788: test_hook_is_nonblocking: added tmp_path TMPDIR isolation.

scripts/validate.py
```diff
@@ -424,6 +434,16 @@ def _expected_slice(sid):
...
                  " '@./shared/languages.md', '@./shared/workflows.md'")
+            continue
+        # Code-touching agents must include both catalogs.
+        if (agent_md.stem not in _NON_CODE_AGENTS
+                and "@./shared/principles.md" in agent_text
+                and "@./shared/languages.md" not in agent_text):
+            fail(agent_md.relative_to(ROOT),
+                 "code-touching agent includes '@./shared/principles.md' but is missing"
+                 " '@./shared/languages.md' — add it so language-* skills are in scope."
+                 " If this agent never touches source code, add its stem to"
```
lugassawan, line 445: **Low** — stale error message still references the old function-local location. After promoting `_NON_CODE_AGENTS` to module scope, the hint `add its stem to _NON_CODE_AGENTS in check_catalog_completeness()` points to the wrong place. Update to:

```suggestion
                 " _NON_CODE_AGENTS at the top of validate.py.")
```

tests/test_agent_language_catalog.py
```diff
@@ -0,0 +1,115 @@
...
+        "the agent to detect the language and invoke the matching language-* skill."
+    )
+
+
+# ──────────────────────────────────────────────────────────────
+# T4 — 'Auto-load when' present in every principle-*/language-* SKILL.md
+# ──────────────────────────────────────────────────────────────
+
+
+def _skill_dirs_with_prefix(prefix: str):
```
lugassawan, line 90: **Low** — `_skill_dirs_with_prefix` is called at pytest collection time with no guard for a missing SKILLS_DIR. If the directory doesn't exist (partial clone, CI subset), `SKILLS_DIR.iterdir()` raises `FileNotFoundError` during collection, producing a cryptic error. Degrade gracefully:

```python
def _skill_dirs_with_prefix(prefix: str):
    if not SKILLS_DIR.is_dir():
        return []
    return [p for p in SKILLS_DIR.iterdir() if p.is_dir() and p.name.startswith(prefix)]
```

scripts/validate.py
lugassawan: Addressed in 027a071e78ea5c5a980544d9c0f4110138130f60: error message now says "at the top of validate.py" instead of the removed function-local reference.

tests/test_agent_language_catalog.py
lugassawan: Addressed in 027a071e78ea5c5a980544d9c0f4110138130f60: added `if not SKILLS_DIR.is_dir(): return []` guard so collection degrades cleanly on partial clones.

PR #382, opened 2026-05-30
[fix] Block force-push to release/* branches

### LINKED ISSUE #341: [bug] bash_guard.sh force-push guard omits release/* branches (and has no tests)
## Product framing
- **Problem:** A contributor (or agent) can `git push --force` to a `release/*` branch without the guard intervening, even though the same branches are protected against `git reset --hard`.
- **Value:** Release branches carry tagged history; closing this aligns the two guards so the most destructive remote op is as protected as the local one.
- **Acceptance criteria:**
  - [ ] Force-push to a `release/*` branch is blocked by `bash_guard.sh`.
  - [ ] `tests/test_bash_guard.py` exists and covers force-push + reset-hard across `main`/`master`/`release/*`.
- **Impact / Effort:** Impact: S (defense-in-depth, accident-prevention). Effort: S (one regex + a test file).

## Surface affected
- [x] Hook
- [x] Build / CI / release
**Specific surface:** `hooks/bash_guard.sh`

## Plugin version
0.1.15

## Reproduction steps
1. Construct a force-push to a release branch, e.g. `git push --force origin release/1.2`.
2. `bash_guard.sh` evaluates the force-push block (`hooks/bash_guard.sh:41-45`).
3. The branch regex on line 42 matches only `(main|master)` — `release/*` is not matched, so the command is allowed.

## Expected behavior
Force-push to `release/*` is blocked, consistent with the sibling `git reset --hard` block (`:47-55`) which already protects `main|master|release/*`.

## Actual behavior
Force-push to `release/*` passes unblocked; the protected-branch set diverges between the two guards. Separately, **no `test_bash_guard*` file exists** — every branch-protection rule in the hook is untested.

## Additional context
Cooperative-agent guardrail (not an adversarial control). Suggested fix: extend the alternation at `:42` to `(main|master|release/[^[:space:]:]*)`, and add `tests/test_bash_guard.py`. Surfaced by a deep cold-start audit.

lugassawan:
## Summary
- Extends the force-push branch regex in `hooks/bash_guard.sh` (line 42) from `(main|master)` to `(main|master|release/[^[:space:]:]*)` — matching the protected-branch set already used by the reset-hard guard
- Closes the guard gap where pushing with the force flag to a release branch slipped through unblocked
- Updates the header docstring and BLOCKED message to mention `release/*`
- Adds four `test_blocked` and two `test_allowed` parametrize cases to `TestForcePushBlocker` in `tests/test_hooks.py` — no new test file; extends the canonical suite

## Test Plan
- [x] TDD cycle completed: 4 new cases failed before fix, all 22 `TestForcePushBlocker` pass after
- [x] `prerelease/x` and `force-with-lease` to release branch confirmed allowed (no false positives)
- [x] Full suite: 1057 tests pass (`python3 -m pytest tests/ -q`)
- [x] Pre-push hook ran and passed on push

Closes #341

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -122,6 +122,7 @@ class TestForcePushBlocker:
         "git push -f origin release/2025-01",
         "git push --force origin HEAD:release/1.2",
         "git push --force origin release/1.2:release/1.2",
+        "git push --force origin release/x:main",
     ])
     def test_blocked(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -117,6 +117,11 @@ class TestForcePushBlocker:
         "git push --force origin feature:main",
         "git push --force origin HEAD:master",
         "git push origin main --force",
+        # release/* cases (issue #341)
+        "git push --force origin release/1.2",
+        "git push -f origin release/2025-01",
+        "git push --force origin HEAD:release/1.2",
+        "git push --force origin release/1.2:release/1.2",
     ])
     def test_blocked(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
@@ -135,6 +140,9 @@ class TestForcePushBlocker:
         "git push --force-with-lease origin master",
         "git push --force-if-includes origin main",
         "git push --force-with-lease=origin/main origin main",
+        # no false positives for similar-looking branch names (issue #341)
+        "git push --force origin prerelease/x",
+        "git push --force-with-lease origin release/1.2",
     ])
     def test_allowed(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
--- a/hooks/bash_guard.sh
+++ b/hooks/bash_guard.sh
@@ -3,7 +3,7 @@
 #
 # Blocks:
 #   - rm -rf against /, /*, ~, $HOME, /Users[/<user>], /home[/<user>]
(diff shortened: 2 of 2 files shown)


tests/test_hooks.py
```diff
@@ -117,6 +117,11 @@ class TestForcePushBlocker:
         "git push --force origin feature:main",
         "git push --force origin HEAD:master",
         "git push origin main --force",
+        # release/* cases (issue #341)
+        "git push --force origin release/1.2",
+        "git push -f origin release/2025-01",
+        "git push --force origin HEAD:release/1.2",
+        "git push --force origin release/1.2:release/1.2",
     ])
```
lugassawan, line 126: **Low:** Missing blocked test case for the cross-protected refspec form where destination is a protected branch (e.g., pushing from a release branch to main, or vice versa).

The current parametrize list covers `release/1.2:release/1.2` (same-branch refspec), `HEAD:release/1.2` (detached HEAD source), and bare `release/1.2`. An explicit case with a cross-protected destination would lock in that the regex blocks pushes regardless of the source branch name, preventing a silent regression if the regex is ever narrowed.

**Suggested fix:** Add to `test_blocked`:
```python
"git push --force origin release/x:main",    # cross-protected: release src → main dst
```

lugassawan: Addressed in 3884deff: added cross-protected destination test case (release-src to main-dst) to TestForcePushBlocker.

PR #383, opened 2026-05-30
[fix] skill_usage_flush: encode before rm, guard against jq failure

### LINKED ISSUE #343: [bug] skill_usage_flush.sh deletes buffers before the telemetry emit is verified
## Product framing
- **Problem:** If the telemetry `jq` encode fails, the flush hook emits malformed JSON and has already deleted the source buffer — no recovery.
- **Value:** Makes the SubagentStop telemetry emit robust to a transient `jq`/encode failure.
- **Acceptance criteria:**
  - [ ] The hook always emits valid JSON (a clean `{}` on encode failure).
  - [ ] The buffer is retained when the emit fails.
- **Impact / Effort:** Impact: S (telemetry only, no correctness/data impact). Effort: S (reorder + one guard).

## Surface affected
- [x] Hook
**Specific surface:** `hooks/skill_usage_flush.sh`

## Plugin version
0.1.15

## Reproduction steps
1. On SubagentStop, the hook reads buffers into `$skills` (`:42`), then `rm -f "${buffers[@]}"` (`:46`).
2. It then builds `msg` via `jq -Rs` (`:49`) and emits (`:50`). The script runs under `set -u` only (no `pipefail`), so a `jq` failure is swallowed and `msg` is empty.
3. Result: emits malformed `{"systemMessage": , "suppressOutput": true}` while the buffer is already deleted.

## Expected behavior
Valid JSON is always emitted; on encode failure the hook emits `{}` and leaves the buffer in place.

## Actual behavior
A `jq` failure yields malformed JSON and an irrecoverable buffer deletion (single-shot per subagent, no retry).

## Additional context
Payload survives in-memory (`$skills`) so there is no loss *within* an invocation — the real defect is the swallowed `jq` failure + premature `rm`. Low severity. Suggested fix: `msg=$(... | jq -Rs .) || { printf '{}'; exit 0; }`, emit, then `rm` last. Surfaced by a deep cold-start audit.

lugassawan:
## Summary
- `hooks/skill_usage_flush.sh`: reorder encode → emit → rm ("commit last" discipline)
- Guard the `jq -Rs` command substitution with `|| { printf '{}'; exit 0; }` so a failed encode emits valid `{}` and leaves the buffer intact for retry
- Update stale comment that incorrectly said "regardless of whether the emit succeeds"

## Test Plan
- [x] `pytest tests/test_skill_usage_telemetry.py -v` — all TestFlushHook cases green, including new `test_flush_retains_buffer_and_emits_empty_json_on_encode_failure`
- [x] `pytest tests/ -v` — 1059 tests pass, no regressions
- [x] `shellcheck --severity=warning hooks/skill_usage_flush.sh` — clean

- [ ] Reproduce: pipe `{"agent_id":"abc123","agent_type":"reviewer"}` into the hook with a failing-`jq` shim on PATH → stdout is `{}`, buffer file retained
- [ ] Success path unchanged: real `jq` path still emits `{"systemMessage": "Skills used by reviewer: …", "suppressOutput": true}` and deletes buffer

Closes #343

### DIFF
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -42,10 +42,11 @@ done
 skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++ { out = (out ? out ", " : "") $0 } END { print out }')
 [ -z "$skills" ] && { printf '{}'; exit 0; }
 
-# Clean up all matching buffers regardless of whether the emit succeeds below.
-rm -f "${buffers[@]}" 2>/dev/null || true
-
-# Safely JSON-encode the message.
-msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .)
+# Safely JSON-encode the message. On encode failure emit a clean {} and leave the
+# buffers in place so a later flush can retry — never emit malformed JSON.
+msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}'; exit 0; }
 printf '{"systemMessage": %s, "suppressOutput": true}\n' "$msg"
+
+# Clean up the buffers only after a successful emit.
+rm -f "${buffers[@]}" 2>/dev/null || true
 exit 0
--- a/hooks/skill_usage_flush.sh
+++ b/hooks/skill_usage_flush.sh
@@ -44,7 +44,7 @@ skills=$(cat "${buffers[@]}" 2>/dev/null | awk '!seen[$0]++ { out = (out ? out "
 
 # Safely JSON-encode the message. On encode failure emit a clean {} and leave the
 # buffers in place so a later flush can retry — never emit malformed JSON.
-msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}'; exit 0; }
+msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}\n'; exit 0; }
 printf '{"systemMessage": %s, "suppressOutput": true}\n' "$msg"
 
 # Clean up the buffers only after a successful emit.
--- a/tests/test_skill_usage_telemetry.py
+++ b/tests/test_skill_usage_telemetry.py
@@ -7,6 +7,7 @@ tree via CLAUDE_PLUGIN_ROOT.
 import datetime
 import json
 import os
+import shutil
 import subprocess
 import time
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


tests/test_skill_usage_telemetry.py
```diff
@@ -431,6 +432,45 @@ def test_date_anchor_rejects_non_date_prefix(self, plugin_root, cache_dir):
...
+        skill_cache = _skill_cache(cache_dir)
+        skill_cache.mkdir(parents=True)
+        today = datetime.date.today().strftime("%Y%m%d")
+        buf = skill_cache / f"{today}-abc123.txt"
+        buf.write_text("skill-a\n")
+
+        # Inject a jq shim that exits 1 when called with -Rs (the encode flag).
+        fake_bin = tmp_path / "fake-bin"
+        fake_bin.mkdir()
+        real_jq = shutil.which("jq") or "jq"
```
lugassawan: **Medium** — Shim silently degrades when `jq` is absent from PATH.

The shim uses `shutil.which("jq") or "jq"` as fallback. If `jq` isn't installed at test-collection time, `real_jq` becomes the bare string `"jq"`, and all passthrough shim calls (`jq -r` for parsing `agent_id`/`agent_type`) fail with "command not found". The hook exits 0 with `{}` before reaching the encode step, so the buffer assertion still passes — but for the wrong reason, masking the shim breakage.

**Fix:** replace the fallback with an explicit assertion:
```python
real_jq = shutil.which("jq")
assert real_jq is not None, "jq must be installed for this test"
```

hooks/skill_usage_flush.sh
```diff
@@ -42,10 +42,11 @@ done
...
 [ -z "$skills" ] && { printf '{}'; exit 0; }
 
-# Clean up all matching buffers regardless of whether the emit succeeds below.
-rm -f "${buffers[@]}" 2>/dev/null || true
-
-# Safely JSON-encode the message.
-msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .)
+# Safely JSON-encode the message. On encode failure emit a clean {} and leave the
+# buffers in place so a later flush can retry — never emit malformed JSON.
+msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}'; exit 0; }
```
lugassawan: **Low** — The `|| { printf '{}'; exit 0; }` guard omits a trailing newline, while the success path uses `printf '...\n'`.

All other early-exit `printf '{}'` calls in the file also omit the newline (consistent with the existing pattern), but if a consumer reads the hook's stdout line-by-line, the missing newline could cause silent concatenation with any subsequent output in a pipeline.

**Suggestion:**
```diff
-msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}'; exit 0; }
+msg=$(printf 'Skills used by %s: %s' "$agent_type" "$skills" | jq -Rs .) || { printf '{}\n'; exit 0; }
```

tests/test_skill_usage_telemetry.py
lugassawan: Addressed in a9b4520a49238e45efca8a2169c6b64ae03d6fb8: replaced `shutil.which("jq") or "jq"` with an explicit `assert real_jq is not None` so a missing `jq` fails loudly at test-collection time rather than silently passing for the wrong reason.

hooks/skill_usage_flush.sh
lugassawan: Addressed in a9b4520a49238e45efca8a2169c6b64ae03d6fb8: changed `printf '{}'` to `printf '{}\n'` in the encode-failure guard, consistent with the success-path output.

PR #384, opened 2026-05-30
[feat] Document dependency flow and enforce no-cycles check

### LINKED ISSUE #371: [feat] Document the agent ↔ command ↔ skill dependency flow to prevent cycles
## Product framing

- **Problem.** `docs/extending.md` and `CONTRIBUTING.md` do not document the allowed dependency direction between agents, commands, and skills. This allows cycles (a skill activating a command that activates the same skill) that break the loading model. The dependency direction is inferable from the codebase but is nowhere stated as a rule.
- **Value.** Contributors understand the layering DAG before writing a new skill or agent, preventing cycles that cause infinite activation loops or context corruption.
- **Acceptance criteria.** `docs/extending.md` gains a "Dependency flow" section with the layering DAG as ASCII art + prose rules; `CONTRIBUTING.md` links it. Optional: `scripts/validate.py` gains a cycle-check that builds the `@`-include/invocation graph and fails on back-edges (machine-enforcement, following the `a2bdf6e` precedent for machine-enforced rules).
- **Impact / Effort.** Impact: M — cycles are a silent correctness bug; documenting the rule prevents them for all future contributors. Effort: S (doc only) / M (doc + validate.py cycle check).

> **Split option:** The doc-only guideline and the `validate.py` cycle-check are independent deliverables — file as one issue (combined) or split into two (doc + enforcement) as preferred.
>
> **Context:** `validate.py` already validates `@`-include format (line 433) but builds no graph and cannot detect cycles. The layering rule exists implicitly in the codebase but is nowhere written down.

## Goal
Document the agent ↔ command ↔ skill dependency flow in `docs/extending.md` and optionally enforce it with a `validate.py` cycle check.

## Motivation
A contributor adding a back-edge (skill → command → same skill) would not know they are creating a cycle until activation breaks at runtime. Writing the rule down costs almost nothing; the optional machine enforcement eliminates the class entirely.

## Surface
- [x] Documentation
- [x] Build / CI (if `validate.py` cycle check is included)

## Sketch (optional)
**Layering DAG (to document):**
```
commands → skills → agents/shared/*
commands → subagents (delegate)
subagents → agents/shared/* (via @-include)
```
Rules: no skill activates a command; orchestrator skills compose only lower-level skills; no back-edges.

**Optional `validate.py` addition:** parse each skill file for `Skill tool` invocations and `@`-include directives, build a directed graph, run DFS cycle detection, fail with the cycle path if found.

## Acceptance (optional)
- [ ] `docs/extending.md` gains a "Dependency flow" section with ASCII DAG + prose rules
- [ ] `CONTRIBUTING.md` links the new section
- [ ] (Optional) `scripts/validate.py` gains cycle detection, fails CI on back-edges
- [ ] `scripts/validate.py` passes (existing checks unaffected)

lugassawan:
## Summary
- Add `## Dependency flow` section to `docs/extending.md` documenting the allowed command→skill/agent→skill layering, the async-handoff exemption, and the no-back-edges rule
- Add a `check_no_cycles` bullet to `CONTRIBUTING.md` → `## Validator` cross-referencing the new section
- Implement `check_no_cycles(cache)` in `scripts/validate.py`: DFS white/gray/black cycle detection on action-cued `` `swe-workbench:<id>` `` activations; excludes slash-command handoffs, `@`-include lines, and pointer-cue prose; registered in `main()`
- Add `TestCheckNoCycles` to `tests/test_validate.py`: 6 tests (green live-repo baseline + positive cycle + 4 negative exclusion cases)

## Test Plan
- [x] `python3 scripts/validate.py` → `All checks passed.` on the live repo (new check is green)
- [x] `pytest tests/test_validate.py::TestCheckNoCycles` → 6/6 passed
- [x] Full `pytest tests/test_validate.py` → 169/169 passed
- [x] Pre-push hook: full test suite (1065 tests) passed on push

- [x] New behaviour: `check_no_cycles` detects a synthetic skill A→skill B→skill A cycle and reports both ids in the failure message
- [x] New behaviour: prose `See \`swe-workbench:X\`` cross-refs produce no edge (pointer-cue exclusion works)
- [x] New behaviour: `` `/review` `` slash handoffs produce no edge (slash exclusion works, backticks required)
- [x] New behaviour: action-cued self-reference (`invoke \`swe-workbench:a\`` inside skill `a`) produces no self-edge (dst==src guard works)
- [x] New behaviour: skill→agent→skill mediated cycle is detected correctly
- [x] Green-baseline test: live repo has zero activation cycles

Closes #371

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -22,7 +22,7 @@ Orchestrator skills that compose many sub-skills (see the `development` workflow
 Three artifact kinds — commands, skills, and agents — form a strict layering:
 
 ```
-command ──► skill
+command ──► skill ──► skill
    │         │
    │         ▼
    └───────► agent ──► skill
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -17,6 +17,29 @@ Skills are intentionally small — each under 150 lines. A sharp, well-triggered
 
 Orchestrator skills that compose many sub-skills (see the `development` workflow) may exceed 150 lines. When they do, extract conditional content (mode templates, rarely-loaded sub-flows) into companion files inside the skill's directory rather than padding the always-loaded `SKILL.md`.
 
+## Dependency flow
+
+Three artifact kinds — commands, skills, and agents — form a strict layering:
+
+```
+command ──► skill
+   │         │
+   │         ▼
+   └───────► agent ──► skill
+```
+
+**Commands** are orchestrators (top layer). A command pulls in whatever skills and subagents a workflow needs. Commands may activate skills and agents.
+
+**Skills** may activate other skills. Composing a workflow from smaller skills is encouraged; it keeps any single skill under the 150-line cap (see `## Philosophy`). Skills must not activate commands.
+
+**Agents** may activate skills. An agent is a leaf worker dispatched *by* a command, never the reverse. Naming the entry command in a `**Reachable via:**` breadcrumb is documentation, not a dependency, and does not count as an activation.
+
+**Async handoffs:** a skill may suggest running a command as a human-driven next step (e.g. "run `/swe-workbench:review` next"). This is an asynchronous handoff — the user decides whether to follow up — not a synchronous activation, and does not form a loading cycle.
+
+**No back-edges:** nothing that a command activates may synchronously activate that command (or its skills) in turn. Cycles in the activation graph break Claude Code's loading behavior and are forbidden at all layers.
+
+`scripts/validate.py` (`check_no_cycles`) machine-enforces the no-cycle rule by scanning action-cued `` `swe-workbench:<id>` `` activations. Slash-command handoffs (`/swe-workbench:cmd`) and prose cross-references (`` See `swe-workbench:X` ``) are intentionally excluded from the graph — they are pointers, not activations.
+
 ## Adding worked examples to a skill
 
 For skills that describe patterns (e.g. `principle-*`), add language-specific implementations as companion files in an `examples/` subdirectory inside the skill's directory:
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -627,14 +627,16 @@ def _build_dep_graph(cache):
             # Skip @-include lines — pure file composition, not activations.
             if line.lstrip().startswith('@'):
                 continue
-            # Skip lines with slash-command refs — async handoffs, not activations.
-            if _SLASH_CMD_RE.search(line):
+            # Strip slash-command tokens (async handoffs) before checking for
+            # action cues and swe-workbench refs. Stripping rather than skipping
+            # the whole line preserves real activation edges that co-occur with
+            # a slash hint (e.g. "invoke `swe-workbench:x` (then run `/review`)").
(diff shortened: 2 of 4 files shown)


scripts/validate.py
```diff
@@ -580,6 +580,136 @@ def check_hook_scripts():
...
+
+    graph = {}
+
+    def _scan(src_node, text):
+        for line in text.splitlines():
+            # Skip @-include lines — pure file composition, not activations.
+            if line.lstrip().startswith('@'):
+                continue
+            # Skip lines with slash-command refs — async handoffs, not activations.
+            if _SLASH_CMD_RE.search(line):
```
lugassawan: **Critical:** `_SLASH_CMD_RE` match skips the **entire line** rather than just stripping the slash-command token. A line combining an action-cued activation with a parenthetical slash hint — e.g. `invoke \`swe-workbench:skill-a\` (then run \`/review\` next)` — is silently dropped from the graph, causing the back-edge to evade cycle detection. The pattern already exists in the live codebase (`skills/workflow-commit-and-pr/SKILL.md:218`).

**Fix:** Strip slash-command tokens before the action-cue check rather than skipping the whole line:
```python
clean = _SLASH_CMD_RE.sub('', line)
if not _ACTION_RE.search(clean):
    continue
if _POINTER_RE.search(clean):
    continue
for mid in _CYCLE_MENTION_RE.findall(clean):
```
This preserves the async-handoff exclusion rationale (the slash token is removed) while keeping the rest of the line in scope for edge emission.

docs/extending.md
```diff
@@ -17,6 +17,29 @@ Skills are intentionally small — each under 150 lines. A sharp, well-triggered
...
+## Dependency flow
+
+Three artifact kinds — commands, skills, and agents — form a strict layering:
+
+```
+command ──► skill
+   │         │
+   │         ▼
+   └───────► agent ──► skill
+```
```
lugassawan, line 29: **Important:** The ASCII diagram omits the `skill → skill` edge. The prose (line 33) explicitly states "Skills may activate other skills", but the diagram only shows `command → skill`, `command → agent`, and `agent → skill`. A contributor reading the diagram first will not see skill-to-skill composition as a permitted pattern, contradicting the prose immediately below.

**Fix:** Add `skill ──► skill` to the diagram, e.g.:
```
command ──► skill ──► skill
   │         │
   │         ▼
   └───────► agent ──► skill
```
The chained top row makes the skill-composition pattern visually obvious.

scripts/validate.py
lugassawan: Addressed in 3b3be7b: replaced whole-line skip with `_SLASH_CMD_RE.sub('', line)` so real activation edges that co-occur with a slash hint are no longer dropped from the graph.

docs/extending.md
lugassawan: Addressed in 3b3be7b: added `skill ──► skill` to the diagram's top row so the diagram matches the prose rule that skills may activate other skills.

PR #388, opened 2026-05-31
[fix] Guard both git checkout main sites in release.sh

### LINKED ISSUE #344: [bug] release.sh aborts with a raw git error when 'git checkout main' fails on resume (two sites)
## Product framing
- **Problem:** When resuming a release in a multi-worktree (rimba) layout, `git checkout main` fails with git's raw `fatal:` instead of the script's guided `Error:` UX — a dead-end in exactly the resume path it exists to serve.
- **Value:** Preserves the script's consistent, actionable error contract under worktree contention.
- **Acceptance criteria:**
  - [ ] A failed `git checkout main` at the resume arm prints a guided `Error:` + remediation and exits 1.
  - [ ] The identical site at `:310` (post-merge sync) is fixed the same way.
- **Impact / Effort:** Impact: M (breaks the guided-error contract; deterministic in worktree layouts). Effort: S (wrap two checkouts).

## Surface affected
- [x] Build / CI / release
**Specific surface:** `scripts/release.sh`

## Plugin version
0.1.15

## Reproduction steps
1. Use a rimba worktree layout where `main` is checked out in the primary worktree.
2. Run `scripts/release.sh` while on a `chore/bump-v*` branch (the resume arm).
3. Line 52 `git checkout main` fails (`fatal: 'main' is already used by worktree at ...`); under `set -euo pipefail` the script aborts with the raw git error.

## Expected behavior
A guided `Error: could not switch to main (checked out in another worktree?)` message with remediation, matching the script's other preflight checks (`:13`, `:22`, `:31`, `:55`).

## Actual behavior
Raw git `fatal:` abort, breaking the otherwise-consistent guided-error UX. The same unguarded shape exists at `:310`.

## Additional context
Aborts loudly (non-zero exit) — not silent, no corruption. Fix both sites:
`if ! git checkout main; then echo "Error: could not switch to main (checked out in another worktree?)." >&2; exit 1; fi`. Surfaced by a deep cold-start audit; honors the "fix all variants together" convention.

lugassawan:
## Summary
- Wraps the bare `git checkout main` at the resume arm (`:52`) in an `if ! git checkout main; then ... exit 1; fi` block — prints a guided `Error: could not switch to main (checked out in another worktree?).` to stderr
- Applies the identical guard to the post-merge sync site (`:310`), both producing the same `Error:` wording to match the script's preflight error contract (`:13`, `:22`, `:31`, `:55`)
- Adds `tests/test_release_sh_checkout_main.py` with 3 static-guard assertions (no bare checkouts, ≥2 guards present, each guard has the `Error:` echo + `exit 1`)

## Test Plan

- [x] `pytest tests/test_release_sh_checkout_main.py -v` → 3 tests pass (TDD green)
- [x] `pytest tests/test_release_sh_gh_repo.py tests/test_release_sh_ci_wait.py` → 25 tests pass (no regression)
- [x] `bash -n scripts/release.sh` → syntax OK
- [x] Full suite `pytest -q` → 1073 tests pass

- [ ] Simulate `git checkout main` failure (worktree contention) and confirm guided `Error:` message is printed to stderr with `exit 1` at both sites
- [ ] Confirm `echo "  git checkout main ..."` help-text lines inside double-quoted strings are not incorrectly flagged as bare checkouts

Closes #344

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -51,6 +51,7 @@ case "$CURRENT_BRANCH" in
     echo "Detected in-progress release branch '${CURRENT_BRANCH}'. Switching to main."
     if ! git checkout main; then
       echo "Error: could not switch to main (checked out in another worktree?)." >&2
+      echo "  Remedy: cd to the main worktree, switch off main, then re-run this script." >&2
       exit 1
     fi
     ;;
@@ -312,6 +313,7 @@ fi
 
 if ! git checkout main; then
   echo "Error: could not switch to main (checked out in another worktree?)." >&2
+  echo "  Remedy: cd to the main worktree, switch off main, then re-run this script." >&2
   exit 1
 fi
 git pull --ff-only origin main
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -49,7 +49,10 @@ case "$CURRENT_BRANCH" in
   main) ;;
   chore/bump-v*)
     echo "Detected in-progress release branch '${CURRENT_BRANCH}'. Switching to main."
-    git checkout main
+    if ! git checkout main; then
+      echo "Error: could not switch to main (checked out in another worktree?)." >&2
+      exit 1
+    fi
     ;;
   *)
     echo "Error: must run from main or a 'chore/bump-v*' resume branch (currently on '${CURRENT_BRANCH}')." >&2
@@ -307,7 +310,10 @@ fi
 
 # ── Sync main + verify merge SHA ─────────────────────────────
 
-git checkout main
+if ! git checkout main; then
+  echo "Error: could not switch to main (checked out in another worktree?)." >&2
+  exit 1
+fi
 git pull --ff-only origin main
 
 MERGE_SHA=""
--- a/tests/test_release_sh_checkout_main.py
+++ b/tests/test_release_sh_checkout_main.py
@@ -41,7 +41,11 @@ def _guard_indices(lines: list[str]) -> list[int]:
 
 
 def _guard_block(lines: list[str], idx: int) -> list[str]:
-    """Return lines from idx up to and including the closing 'fi'."""
+    """Return lines from idx up to and including the closing 'fi'.
+
+    Assumes the guard body contains no nested if/fi pairs.
+    If guard bodies grow nested conditionals, switch to depth tracking.
+    """
     block = []
     for ln in lines[idx:]:
         block.append(ln)
--- a/tests/test_release_sh_checkout_main.py
+++ b/tests/test_release_sh_checkout_main.py
@@ -40,6 +40,16 @@ def _guard_indices(lines: list[str]) -> list[int]:
     ]
 
 
+def _guard_block(lines: list[str], idx: int) -> list[str]:
+    """Return lines from idx up to and including the closing 'fi'."""
+    block = []
+    for ln in lines[idx:]:
+        block.append(ln)
+        if ln.strip() == "fi":
+            break
+    return block
+
+
 class TestCheckoutMainGuarded:
     """Static: every git checkout main must be wrapped in an if ! guard."""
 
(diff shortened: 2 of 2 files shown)


scripts/release.sh
```diff
@@ -49,7 +49,10 @@ case "$CURRENT_BRANCH" in
   main) ;;
   chore/bump-v*)
     echo "Detected in-progress release branch '${CURRENT_BRANCH}'. Switching to main."
-    git checkout main
+    if ! git checkout main; then
+      echo "Error: could not switch to main (checked out in another worktree?)." >&2
```
lugassawan, line 53: **Medium:** Error message names the cause but omits the remediation step.

The existing error convention in this script always pairs a diagnosis with an action (e.g. line 13: `Run: gh auth login`, line 38: `Install with: brew install jq`). An operator hitting this in CI sees `checked out in another worktree?` and has no concrete next step.

**Suggested fix:**
```suggestion
      echo "Error: could not switch to main (checked out in another worktree?)." >&2
      echo "  Remedy: cd to the main worktree, switch off main, then re-run this script." >&2
```

(Same addition applies to the identical echo at line 314.)

tests/test_release_sh_checkout_main.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_each_guard_has_error_echo_and_exit(self):
+        """Every guard block must emit 'Error: could not switch to main' to stderr and exit 1."""
+        lines = _script_lines()
+        guards = _guard_indices(lines)
+        assert guards, "No 'if ! git checkout main; then' guards found — expected at least 2"
+
+        for idx in guards:
+            lineno = idx + 1
+            # Inspect up to 5 lines after the guard opener
+            block = lines[idx : idx + 5]
```
lugassawan: **Low:** The 5-line look-ahead window is a fragile proximity heuristic.

`block = lines[idx : idx + 5]` will silently pass if someone inserts a blank line or comment between the `if !` opener and the `echo`/`exit`, pushing the checks outside the window. Scanning up to the matching `fi` is more robust:

```python
def _guard_block(lines: list[str], idx: int) -> list[str]:
    block = []
    for ln in lines[idx:]:
        block.append(ln)
        if ln.strip() == "fi":
            break
    return block
```

Replace `block = lines[idx : idx + 5]` with `block = _guard_block(lines, idx)`.

scripts/release.sh
lugassawan: Addressed in 8d92837: added Remedy echo line to both guard sites (resume arm and post-merge sync).

tests/test_release_sh_checkout_main.py
lugassawan: Addressed in 8d92837: replaced `lines[idx : idx + 5]` with `_guard_block()` helper that scans to the matching `fi`; updated assert messages to remove the stale '5 lines' reference.

tests/test_release_sh_checkout_main.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _guard_block(lines: list[str], idx: int) -> list[str]:
+    """Return lines from idx up to and including the closing 'fi'."""
+    block = []
+    for ln in lines[idx:]:
+        block.append(ln)
+        if ln.strip() == "fi":
+            break
+    return block
```
lugassawan, line 54: **Low (nit):** `_guard_block` stops at the first `fi` regardless of nesting depth.

If a future guard body gains a nested `if` (e.g. an extra conditional message), the helper stops early and silently misses `exit 1`, producing a false-green test. Both current guard bodies are flat, so this is safe today — worth documenting the assumption:

```python
def _guard_block(lines: list[str], idx: int) -> list[str]:
    """Return lines from idx up to and including the closing 'fi'.

    Assumes the guard body contains no nested if/fi pairs.
    If guard bodies grow nested conditionals, switch to depth tracking.
    """
    block = []
    for ln in lines[idx:]:
        block.append(ln)
        if ln.strip() == "fi":
            break
    return block
```

lugassawan: Addressed in ac11949: added docstring noting the flat-body assumption and pointing to depth tracking as the upgrade path.

PR #390, opened 2026-05-31
[fix] invoke doctor.sh via $CLAUDE_PLUGIN_ROOT for plugin-install portability

### LINKED ISSUE #328: [bug] /doctor invokes scripts/doctor.sh via a non-portable relative path
## Product framing

- **Problem.** When I install swe-workbench as a plugin and run `/swe-workbench:doctor` from my own
  project, nothing runs — the command tells Claude to `bash scripts/doctor.sh`, but that relative
  path doesn't exist in my project's working directory.
- **Value.** Plugin-installed users (not just repo cloners) get a working `/doctor` preflight check.
- **Acceptance criteria.**
  - `commands/doctor.md` invokes the script via `$CLAUDE_PLUGIN_ROOT` so it resolves from the
    installed plugin dir.
  - Matches the convention used by hooks and other skills.
  - `tests/test_doctor_command.py` still passes.
- **Impact / Effort.** Impact: M — broken command for every plugin-installed user. Effort: S — one-line change.

## Surface affected

- [x] Slash command

**Specific surface:** `/swe-workbench:doctor` (`commands/doctor.md`)

## Claude Code version

N/A — static repo defect, version-independent.

## Plugin version

0.1.15

## Reproduction steps

1. Install swe-workbench as a plugin (not a local clone).
2. From any project directory, run `/swe-workbench:doctor`.
3. Claude runs `bash scripts/doctor.sh` relative to the project CWD.

## Expected behavior

The doctor script runs and prints the dependency table.

## Actual behavior

`scripts/doctor.sh` is not found — it lives under the plugin root, not the project CWD.

## Environment (optional)

Any plugin install.

## Logs / screenshots (optional)

N/A. Fix: change `commands/doctor.md:8` to `bash $CLAUDE_PLUGIN_ROOT/scripts/doctor.sh`.

lugassawan:
## Summary
- Fix `commands/doctor.md:8` to invoke `scripts/doctor.sh` via `$CLAUDE_PLUGIN_ROOT` instead of a bare relative path — resolves the "file not found" failure for all plugin-installed users
- Add regression test `test_doctor_invokes_script_portably()` to prevent the bare-path pattern from being reintroduced

## Test Plan
- [ ] `pytest tests/test_doctor_command.py -q` — all 6 tests pass (including new `test_doctor_invokes_script_portably`)

- [ ] Reproduce: install swe-workbench as a plugin, run `/swe-workbench:doctor` from any project dir — script now runs and prints the dependency table (previously: "file not found")
- [x] `pytest tests/ -q` — 1110/1110 pass, no regressions

Closes #328

### DIFF
--- a/commands/doctor.md
+++ b/commands/doctor.md
@@ -5,7 +5,7 @@ description: Read-only preflight check of runtime dependencies (gh, git, jq, rim
 Run the preflight diagnostic and print the results verbatim:
 
 ```
-bash scripts/doctor.sh
+bash "$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh"
 ```
 
 Print the full stdout output exactly as produced — do not summarise, truncate, or reformat it.
--- a/tests/test_doctor_command.py
+++ b/tests/test_doctor_command.py
@@ -70,3 +70,14 @@ def test_doctor_in_readme():
     assert "/swe-workbench:doctor" in commands_line, (
         "README.md '- **Commands**' bullet must include /swe-workbench:doctor"
     )
+
+
+def test_doctor_invokes_script_portably():
+    """doctor.md must invoke the script via $CLAUDE_PLUGIN_ROOT (portable for plugin installs, #328)."""
+    text = DOCTOR_CMD.read_text()
+    assert "$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh" in text, (
+        "doctor.md must invoke the script via $CLAUDE_PLUGIN_ROOT for plugin-install portability"
+    )
+    assert "bash scripts/doctor.sh" not in text, (
+        "doctor.md must not invoke the script via a non-portable bare relative path"
+    )


commands/doctor.md
```diff
@@ -5,7 +5,7 @@ description: Read-only preflight check of runtime dependencies (gh, git, jq, rim
 Run the preflight diagnostic and print the results verbatim:
 
 ```
-bash scripts/doctor.sh
+bash "$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh"
```
lugassawan, line 8: **High** — `$CLAUDE_PLUGIN_ROOT` is used bare with no fallback. If the variable is unset (local-clone dev environment, test runner without the env var), bash expands it to empty and the invocation becomes `bash "/scripts/doctor.sh"`, silently failing.

Note: `hooks/hooks.json` also uses the bare form for its five entries — so this matches the established command-invocation convention (the variable is always set when Claude Code runs plugin commands). However, `hooks/skill_usage_flush.sh` and `hooks/skill_usage_record.sh` both use `${CLAUDE_PLUGIN_ROOT:-$(cd "$(dirname "$0")/.."\  && pwd)}` as a dev-mode fallback. Worth considering whether this command warrants the same treatment.

Suggested fix if fallback is desired:
```
bash "${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}/scripts/doctor.sh"
```

tests/test_doctor_command.py
```diff
@@ -70,3 +70,14 @@ def test_doctor_in_readme():
     assert "/swe-workbench:doctor" in commands_line, (
         "README.md '- **Commands**' bullet must include /swe-workbench:doctor"
     )
+
+
+def test_doctor_invokes_script_portably():
+    """doctor.md must invoke the script via $CLAUDE_PLUGIN_ROOT (portable for plugin installs, #328)."""
+    text = DOCTOR_CMD.read_text()
+    assert "$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh" in text, (
```
lugassawan, line 78: **Medium** — The positive assertion checks for the exact string `"$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh"`. If the High finding above is addressed by adding the fallback form (`${CLAUDE_PLUGIN_ROOT:-...}`), this assertion will spuriously fail even though the command is correct.

Suggested fix — loosen the positive assertion to be resilient to the fallback form:
```python
assert "CLAUDE_PLUGIN_ROOT" in text and "scripts/doctor.sh" in text, (
    "doctor.md must invoke the script via $CLAUDE_PLUGIN_ROOT for plugin-install portability"
)
assert "bash scripts/doctor.sh" not in text, (
    "doctor.md must not invoke the script via a non-portable bare relative path"
)
```
The negative assertion on `"bash scripts/doctor.sh"` is correct as-is.

commands/doctor.md
lugassawan: Keeping the bare form intentionally — `commands/doctor.md` tells Claude Code what shell command to run, analogous to an entry in `hooks/hooks.json`, not a standalone script. Claude Code always sets `$CLAUDE_PLUGIN_ROOT` before executing plugin commands, so the variable is guaranteed to be present in this invocation context. The hook shell scripts (`skill_usage_flush.sh`, etc.) use the fallback because they can also be executed directly in dev environments without Claude Code setting the env var; command markdown files cannot.

tests/test_doctor_command.py
lugassawan: Keeping the strict assertion — since Thread 1 is CLARIFIED (bare `$CLAUDE_PLUGIN_ROOT` form stays unchanged), the exact string `"$CLAUDE_PLUGIN_ROOT/scripts/doctor.sh"` remains the canonical expected value. The spurious-failure scenario described only arises if the fallback form is introduced, which won't happen here.

PR #398, opened 2026-06-02
[chore] Add pi package port

arrokh:
## Issue / Task
N/A — pi package publishing support requested directly.

## Summary
Adds a publishable pi package manifest for `swe-workbench` at the repository root with pi package metadata, resource paths, and npm pack settings. Ports Claude Code commands into pi prompt templates, adapts Claude agent roles into pi skills, and adds a pi extension for safety hooks such as bash guarding, secret detection, and language-skill hints.

The package is intended to be installed from the repository/package root and published/discovered as `swe-workbench` for pi:

```bash
pi install npm:swe-workbench
```

## Included pi resources
- `pi-package/extensions` — TypeScript pi extension porting relevant Claude hook behavior.
- `skills` — original agnostic Agent Skills resources loaded directly from the root package.
- `pi-package/agent-skills` — pi-adapted Claude agent roles.
- `pi-package/prompts` — pi-adapted prompt templates generated from Claude commands.

## Test Plan
- [x] `./scripts/validate.sh`
- [x] TypeScript check for `pi-package/extensions/index.ts`
- [x] `npm pack --dry-run` confirms `swe-workbench-0.1.16-pi.0.tgz`
- [x] Local pi package install with `pi install .`
- [x] Packed-artifact install smoke using extracted tarball and `pi install "$tmp/package"`
- [x] pi startup smoke from installed packed artifact exits 0


pi-package/extensions/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const hintedSkills = new Set<string>();
+
+function stringField(input: ToolInput, key: string): string {
+  const value = input[key];
+  return typeof value === "string" ? value : "";
+}
+
+function normalizeCommand(command: string): string {
+  return command.replace(/[;|&]/g, " ").replace(/[\"'()\[\]{}]/g, "");
```
lugassawan: **Critical — Backtick and `$(…)` subshell bypass in `normalizeCommand`**

The function strips `()` and `{}` but leaves backticks untouched. A model generating:
```bash
echo `rm -rf /`
```
or `var=$(rm -rf /)` bypasses the guard entirely — backtick is not whitespace, so `(^|\s)rm\s+` never matches. The `$()` form partially survives (parens stripped → `$rm`) but `$rm` also evades the word-boundary check.

**Fix:** Include backtick in the separator set and replace `$(`  with a space:
```ts
function normalizeCommand(command: string): string {
  return command
    .replace(/[;|&`]/g, " ")      // backtick now sanitised
    .replace(/\$\s*\(/g, " ")     // $( becomes space, not stripped-parens
    .replace(/[\"'()\[\]{}]/g, "");
}
```

This is the primary safety guarantee of the extension — any bypass here defeats the entire bash-guard.

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    if (toolName === "read" || toolName === "edit" || toolName === "write") {
+      const path = stringField(input, "path") || stringField(input, "file_path");
+      const hint = path ? hintForPath(path) : null;
+      if (hint) ctx.ui.notify(hint, "info");
+    }
+  });
+
+  pi.on("session_start", async (_event, ctx) => {
```
lugassawan, line 159: **High — `hintedSkills` is module-level and never cleared on `session_start`**

`hintedSkills` is a `Set` at module scope. If the pi runtime keeps the extension alive across sessions (the common case for plugin hosts), the set fills up in session 1 and every subsequent session receives zero language-skill hints for file types already seen.

The `session_start` handler does not call `hintedSkills.clear()`.

**Fix:**
```ts
pi.on("session_start", async (_event, ctx) => {
  hintedSkills.clear();   // reset per-session dedup
  ctx.ui.setStatus("swe-workbench", "loaded");
});
```

pi-package/prompts/report-issue.md
```diff
@@ -0,0 +1,95 @@
...
+
+6. **Duplicate scan.** Extract 2–3 keywords from the thought. Strip each to `[a-zA-Z0-9_-]` only (drop shell metacharacters, quotes, and flags) and single-quote them when building the search string. Then run `gh issue list --repo lugassawan/swe-workbench --search '<sanitized keywords>' --state open --limit 5`. Surface matches. Ask before drafting if any look duplicative.
+
+7. **Draft.** With a template: fill its sections, prepend `## Product framing`. Without a template: use `## Problem` / `## Value` / `## Acceptance criteria` / `## Impact / Effort` / `## Additional context`.
+
+   Append the following footer to the body's last free-text section (or as a final `---` block):
+
+   ```markdown
+   ---
+   _Reported via `/swe-workbench:report-issue` — plugin v<version>, Claude Code <cli-version>._
```
lugassawan: **High — Hardcoded Claude Code paths in a pi prompt**

Three Claude-specific references survive in this pi-facing prompt:

1. **Line 21** — scans `~/.claude/projects/<slug>/memory/MEMORY.md` — this path only exists in Claude Code, not in pi.
2. **Line 71** — footer watermark: `Claude Code <cli-version>` — pi users will produce issues tagged as Claude Code bugs.
3. **Lines 75-76** — version detection runs `ls ~/.claude/plugins/cache/swe-workbench/…` and `claude --version` — neither exists in a pi environment; version strings will be empty or produce errors.

The port note at line 6 says "use pi's available tools" but this section retains concrete shell commands with Claude-specific paths.

**Fix:** Replace the footer watermark with `_Reported via swe-workbench — v<version>._` without mentioning Claude Code. For version detection, add a pi-specific block (e.g. reading `package.json` from the installed package location). For the memory scan step, either remove it for pi or document that pi uses a different memory location.

pi-package/prompts/review.md
```diff
@@ -0,0 +1,100 @@
...
+   - **Medium** — design smells, SOLID violations, maintainability risks.
+   - **Low** — naming, minor clarity.
+5. Each finding uses: `Severity | File:Line | Issue | Why it matters | Suggested fix`.
+6. Close with a short section summary: correctness bugs, security issues, design smells, test gaps.
+
+Ground judgements in SOLID and Clean Architecture principles. Do not nitpick formatting — that is the linter's job.
+
+## PR mode
+
+**When `--mode` is absent or `--mode general`:** invoke `swe-workbench:workflow-pr-review` via the `Skill` tool, passing the resolved PR number.
```
lugassawan: **High — `via the \`Skill\` tool` instruction has no meaning in pi**

Pi does not expose a "Skill tool" — skills are invoked differently. A model following this instruction literally in pi will hallucinate or error. The port note at line 6 says "use pi's available tools", but this section still references the Claude Code-specific `Skill` tool invocation pattern in two places.

**Fix:** Replace both occurrences of "via the `Skill` tool" with plain "invoke" or "activate":

Before:
```
invoke `swe-workbench:workflow-pr-review` via the `Skill` tool, passing the resolved PR number.
```
After:
```
invoke `swe-workbench:workflow-pr-review`, passing the resolved PR number.
```

Same change needed at line 98 for the followup-mode section.

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    const input = (event.input ?? {}) as ToolInput;
+    const toolName = event.toolName.toLowerCase();
+
+    if (toolName === "bash") {
+      const reason = blockReasonForBash(stringField(input, "command"));
+      if (reason) return { block: true, reason };
+    }
+
+    if (toolName === "write" || toolName === "edit") {
+      const reason = blockReasonForSecret(event.toolName, input);
```
lugassawan: **Medium — `blockReasonForSecret` receives `event.toolName` (original case) despite the outer handler already having a lowercased copy**

At line 148, the outer `tool_call` handler passes `event.toolName` (original case) to `blockReasonForSecret`, which then re-lowercases it internally on line 106 (`toolName.toLowerCase()`). The outer `toolName` variable (already lowercased, line 140) is not used for this call.

This works today but is fragile: anyone adding a new branch who forgets the inner re-lowercasing will get a silent miss.

**Fix:** Pass the already-lowercased `toolName` to `blockReasonForSecret` and remove the internal re-lowercasing:

```ts
// line 148 — change from:
const reason = blockReasonForSecret(event.toolName, input);
// to:
const reason = blockReasonForSecret(toolName, input);
```

Then update `blockReasonForSecret` line 106 to accept an already-lowercased name (remove the `.toLowerCase()` call there).

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,162 @@
...
+];
+
+function scanSecret(content: string): { name: string; line: number; needsContext: boolean } | null {
+  const lines = content.split(/\r?\n/);
+  for (let index = 0; index < lines.length; index += 1) {
+    const line = lines[index];
+    const suppressed = noSecretPattern.test(line);
+    const isReference = refPattern.test(line);
+    for (const item of secretPatterns) {
+      if (item.needsContext && (suppressed || isReference)) continue;
```
lugassawan: **Medium — `# nosecret` suppression is bypassed for `needsContext=false` patterns**

The `scanSecret` loop at line 88 only checks `suppressed || isReference` when `item.needsContext` is true. Patterns like `aws-access-key-id` (`AKIA…`) and `github-pat` (`ghp_…`) have `needsContext: false` — meaning a well-known fixture key used in test documentation (e.g. `AKIAIOSFODNN7EXAMPLE` from AWS SDK docs) will permanently block file writes with no escape hatch.

`needsContext: false` was intended to mean "suspicious even in env-var context", not "can never be suppressed".

**Fix:** Add a separate `suppressible` field, or honour `noSecretPattern` unconditionally for all patterns, letting test-fixture authors annotate their way out of false positives:

```ts
for (const item of secretPatterns) {
  if (suppressed) continue;  // # nosecret always wins
  if (item.needsContext && isReference) continue;
  if (item.pattern.test(line)) return { ... };
}
```

package.json
```diff
@@ -0,0 +1,62 @@
...
+    "README.md",
+    "docs/",
+    "skills/",
+    "pi-package/README.md",
+    "pi-package/extensions/",
+    "pi-package/prompts/",
+    "pi-package/agent-skills/"
+  ],
+  "peerDependencies": {
+    "@earendil-works/pi-coding-agent": "*"
```
lugassawan, line 40: **Medium — `@earendil-works/pi-coding-agent` is listed as `peerDependency` but not as `devDependency`, breaking `typecheck:pi` in a fresh checkout**

`tsc` needs the type declarations from this package to resolve the `ExtensionAPI` import at `pi-package/extensions/index.ts:2`. Without it in `devDependencies`, `npm install` in this repo does not pull it, and `npm run typecheck:pi` fails with:

```
error TS2307: Cannot find module '@earendil-works/pi-coding-agent' or its corresponding type declarations.
```

`--skipLibCheck` skips errors *inside* `node_modules` type files, not errors about *missing* modules.

**Fix:** Add to `devDependencies`:
```json
"devDependencies": {
  "@earendil-works/pi-coding-agent": "*",
  "@types/node": "^24.0.0",
  "typescript": "^5.9.0"
}
```

pi-package/prompts/implement.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Standard mode:** proceed with the command's existing lightweight clarify (a restatement and at most one clarifying question) — do not ask the mode question again.
+
+**Grill-me mode:** activate `swe-workbench:workflow-grill` and run its interrogation loop to completion (exit on shared understanding or when the user says "proceed"). Then thread the emitted `## Resolved decisions` block into the command's normal artifact/delegation step below — the same way a ticket-context summary is prepended — and continue as in standard mode.
+
+Activate the `swe-workbench:workflow-development` skill in **Mode B (Implementation-Time Behavior)**. Execute all five phases in order:
+
+**Phase 1 — Branch**
+Isolate this work in a worktree when the scope warrants it (non-trivial changes, risk of destabilizing main). For tiny single-file changes, skip with a written rationale. `workflow-development` Phase 1 will detect the worktree provider automatically: `rimba add <task>` when rimba is available (PATH or common install locations), otherwise `superpowers:using-git-worktrees`.
```
lugassawan: **Medium — Five `superpowers:*` skill invocations have no pi equivalent, and the port note provides no fallback mapping**

The following invocations from the Claude Code original survive verbatim in this pi prompt:

- Line 23: `superpowers:using-git-worktrees`
- Line 27: `superpowers:writing-plans`
- Line 28: `superpowers:subagent-driven-development` / `superpowers:executing-plans`
- Line 33: `superpowers:verification-before-completion`
- Line 36: `superpowers:requesting-code-review`

A pi model will either skip these silently or attempt to call non-existent tools. This affects `/implement` — the most complex command in the set.

**Fix (option A):** Replace each with prose describing the same intent without referencing the skill:

| Original | Pi fallback |
|---|---|
| `superpowers:writing-plans` | "Draft a written plan document" |
| `superpowers:executing-plans` | "Work through the plan steps sequentially" |
| `superpowers:subagent-driven-development` | "Delegate parallelizable units to sub-agents" |
| `superpowers:verification-before-completion` | "Verify with tests before claiming done" |
| `superpowers:requesting-code-review` | "Run a diff review before delivering" |
| `superpowers:using-git-worktrees` | "Use `rimba add <task>` when available, otherwise `git worktree add`" |

**Fix (option B):** Document that these phases are unsupported in pi and skip them with a note.

pi-package/agent-skills/shared/workflows.md
```diff
@@ -0,0 +1,19 @@
...
+- `swe-workbench:workflow-address-feedback` — PR-owner feedback loop: fetch open review threads, per-thread ADDRESSED/CLARIFIED/DEFERRED triage, Edit-tool fixes, workflow-commit-and-pr commit, REST reply posting, and GraphQL resolveReviewThread. Invoked by `/address-feedback`.
+- `swe-workbench:workflow-pr-review` — Remote-PR review orchestration: ephemeral worktree + reviewer agent + GraphQL thread dedup + REST inline-comment post + APPROVE/COMMENT submit. Invoked by `/review` PR mode.
+- `swe-workbench:workflow-pr-review-followup` — Reviewer follow-up re-check: re-runs reviewer agent against the updated diff, deduplicates against existing threads (Jaccard ±5-line), posts only truly-new inline comments, and submits APPROVE/COMMENT. Invoked by `/review --check-followup <N>`.
+- `swe-workbench:workflow-worktree-session` — Start, switch, or end a worktree-bound session via `EnterWorktree` / `ExitWorktree`. No claude restart.
```
lugassawan: **Low — Claude Code-specific tool names (`EnterWorktree` / `ExitWorktree`) in a pi-facing shared catalog**

The description for `workflow-worktree-session` at line 18 mentions "via `EnterWorktree` / `ExitWorktree`" — these are Claude Code internal tools that do not exist in pi. Since this is a shared catalog consumed by all agent-skill files, the misleading implementation detail affects every skill that loads it.

**Fix:** Drop the implementation detail and keep only the intent:

Before:
> `swe-workbench:workflow-worktree-session` — Start, switch, or end a worktree-bound session via `EnterWorktree` / `ExitWorktree`. No claude restart.

After:
> `swe-workbench:workflow-worktree-session` — Start, switch, or end a worktree-bound session.

package.json
```diff
@@ -0,0 +1,62 @@
...
+  },
+  "scripts": {
+    "validate": "./scripts/validate.sh",
+    "typecheck:pi": "tsc --noEmit --skipLibCheck --moduleResolution node16 --module Node16 --target ES2022 --types node pi-package/extensions/index.ts",
```
lugassawan: **Low — No `tsconfig.json` exists; compiler options are baked into the npm script only**

The `typecheck:pi` script at line 43 passes all compiler flags inline. There is no `tsconfig.json` anywhere in the repository. This means:

- IDE tooling (VS Code, WebStorm, etc.) does not pick up the compiler options automatically.
- Future contributors have no single source of truth for the TypeScript configuration.
- Options drift silently when `package.json` scripts diverge from editor defaults.

**Fix:** Create `pi-package/extensions/tsconfig.json` with the equivalent config, and replace the inline flags with `tsc --noEmit -p pi-package/extensions/tsconfig.json`.

lugassawan (commented):
**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 9 inline comments, deduped 0._

pi-package/extensions/index.ts
arrokh: Valid. The bash guard normalization now treats backticks and `$` as separators too, so command substitution cannot hide a protected `rm`/`git` token from the regex scan. Fixed in b5033ee.

pi-package/extensions/index.ts
arrokh: Valid. `hintedSkills` is now reset on each `session_start`, so hint de-duplication is session-scoped instead of process/module-lifetime scoped. Fixed in b5033ee.

pi-package/prompts/report-issue.md
arrokh: Valid. Removed the Claude Code memory/cache assumptions from the pi prompt. It now uses pi-accessible workspace/session context and captures package/pi versions instead of Claude cache/CLI paths. Fixed in b5033ee.

pi-package/prompts/review.md
arrokh: Valid. Replaced the Claude-specific `Skill` tool wording with pi skill-loading guidance, including `/skill:workflow-pr-review` examples when skill commands are enabled. Fixed in b5033ee.

pi-package/extensions/index.ts
arrokh: Valid. The handler now passes the already-normalized `toolName` into `blockReasonForSecret`, avoiding case drift between the outer dispatch and helper. Fixed in b5033ee.

pi-package/extensions/index.ts
arrokh: Valid. `# nosecret` now suppresses all configured secret patterns, including high-confidence fixture patterns, matching the documented cooperative-agent suppression behavior. Fixed in b5033ee.

package.json
arrokh: Valid. Added `@earendil-works/pi-coding-agent` to devDependencies so local/package typechecking can install the imported type package. Fixed in b5033ee.

pi-package/prompts/implement.md
arrokh: Valid. Added an explicit pi fallback mapping for Claude-only `superpowers:*` references in the implement prompt so pi users have concrete behavior instead of unresolved skill names. Fixed in b5033ee.

pi-package/agent-skills/shared/workflows.md
arrokh: Valid. Reworded the shared workflow catalog to avoid relying on Claude-only `EnterWorktree` / `ExitWorktree` tools and describe pi/git-worktree behavior instead. Fixed in b5033ee.

package.json
arrokh: Valid. Added `tsconfig.pi.json` and changed `typecheck:pi` to use it; the file is included in the package tarball. Fixed in b5033ee.

pi-package/prompts/implement.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+**Pi fallback mapping for Claude-only `superpowers:*` references:** if a referenced `superpowers:*` skill is unavailable in pi, perform the described behavior directly with pi tools and packaged SWE Workbench skills: use git worktrees or the current branch for work isolation, write an explicit plan in the conversation or a requested file, execute the plan step-by-step with tests first where practical, run concrete verification commands before claiming completion, and use packaged `agent-reviewer` / review skills for review.
+
+Activate the `swe-workbench:workflow-development` skill in **Mode B (Implementation-Time Behavior)**. Execute all five phases in order:
+
+**Phase 1 — Branch**
+Isolate this work in a worktree when the scope warrants it (non-trivial changes, risk of destabilizing main). For tiny single-file changes, skip with a written rationale. `workflow-development` Phase 1 will detect the worktree provider automatically: `rimba add <task>` when rimba is available (PATH or common install locations), otherwise `superpowers:using-git-worktrees`.
```
lugassawan: **Medium — Four live `superpowers:*` invocations remain in the pi prompt (partially fixed)**

The previous finding flagged five `superpowers:*` references with no pi equivalent. The fix adds a fallback note at line 21:

> _"if a referenced `superpowers:*` skill is unavailable in pi, perform the described behavior directly..."_

However, four references survive in the prompt body:
- **Line 26:** `superpowers:using-git-worktrees` (rimba fallback)
- **Line 31:** `superpowers:executing-plans` and `superpowers:subagent-driven-development`
- **Line 36:** `superpowers:verification-before-completion`
- **Line 39:** `superpowers:requesting-code-review`

The note is conditional ("if unavailable"), so the model must first attempt the invocation and discover unavailability at runtime. Lines 44-45 say "Do not skip Phase 3 or Phase 4 under any circumstances," creating pressure to execute these phases even when the skill is missing.

**Suggested fix:** Replace each reference with a pi-native direct instruction:
- Line 26: `otherwise create the worktree manually via git worktree commands`
- Line 31: `otherwise execute the plan step-by-step via pi tools`
- Line 36: `run the verification commands defined in the plan`
- Line 39: `invoke the packaged agent-reviewer skill`

lugassawan (approved):
**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment, deduped 0._

pi-package/prompts/implement.md
arrokh: Valid. The previous fallback note still left live Claude-only skill references in the pi-facing implement prompt. I replaced those with direct pi-native instructions for worktree creation, planning, execution, verification, and review, and also removed the remaining Claude-only fallback mention from the shared workflow catalog. Fixed in b83c612.

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,163 @@
...
+  } catch {
+    return false;
+  }
+}
+
+function blockReasonForBash(command: string): string | null {
+  const normalized = normalizeCommand(command);
+  if (!/(^|\s)rm\s+|git/.test(normalized)) return null;
+
+  if (/(^|\s)rm\s+-[a-zA-Z]*[rR][a-zA-Z]*[fF]?\s+(\/(\*|\s|$)|(~|\$HOME)(\/[^\s]*)?(\s|$)|(\/Users|\/home)(\/[^/\s]+)?(\s|\/|$))/.test(normalized)) {
```
lugassawan, line 52: **High — `$HOME` guard bypassed by `normalizeCommand`**

`normalizeCommand` (line 35) replaces every `$` with a space before this regex runs, so `rm -rf $HOME/secrets` becomes `rm -rf  HOME/secrets` — the `\$HOME` arm on this line never matches. The `~` arm still fires, but `$HOME` is a common shell idiom and the miss is silent.

**Suggested fix:** Run the `rm` check against the *raw* command before normalization, or match the post-normalization form:

```ts
// Check raw command before normalizing
const raw = command;
const normalized = normalizeCommand(command);
if (/(^|\s)rm\s+-[a-zA-Z]*[rR][a-zA-Z]*[fF]?\s+(\/(\*|\s|$)|(~|\$HOME|HOME)(\/[^\s]*)?(\s|$)|(\/Users|\/home)(\/[^\/\s]+)?(\s|\/|$))/.test(raw)) {
  return "BLOCKED: destructive rm against root or home";
}
```

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,163 @@
...
+  kts: "language-kotlin",
+  rb: "language-ruby",
+  swift: "language-swift",
+  sh: "language-bash",
+  bash: "language-bash",
+  sql: "language-sql",
+  cs: "language-csharp",
+};
+
+let hintedSkills = new Set<string>();
```
lugassawan, line 27: **Medium — `hintedSkills` module-level `let` persists across sessions; `session_start` reset is a partial fix**

The `session_start` handler at line 159 does reset `hintedSkills`, but the module-level `let` means the set persists if the event is missed or delayed (e.g. in a long-lived server process). This is a fragile invariant — the language does not enforce session scope; only event ordering does.

**Suggested fix:** Move `hintedSkills` inside the exported function closure:

```ts
export default function sweWorkbenchPi(pi: ExtensionAPI) {
  let hintedSkills = new Set<string>();
  // session_start reset is no longer needed — each call gets a fresh set
```

tsconfig.pi.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "module": "Node16",
+    "moduleResolution": "node16",
+    "target": "ES2022",
+    "types": ["node"],
+    "skipLibCheck": true,
+    "noEmit": true
```
lugassawan, line 8: **Medium — `strict` mode absent in `tsconfig.pi.json`**

Without `"strict": true`, `noImplicitAny`, `strictNullChecks`, and related checks are off. The extension handles nullability correctly today, but future contributors will not get that safety net when editing `pi-package/extensions/index.ts`.

**Suggested fix:**
```json
    "skipLibCheck": true,
    "strict": true,
    "noEmit": true
```

package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "swe-workbench",
+  "version": "0.1.16-pi.0",
```
lugassawan, line 3: **Medium — `version` is `0.1.16-pi.0`; live plugin is `0.1.20`**

The main plugin version (`.claude-plugin/plugin.json`) is `0.1.20` but this package.json was set to `0.1.16-pi.0`. The `-pi.0` suffix correctly distinguishes the npm package, but having the base version 4 patches behind means any tooling that surfaces `package.json#version` (including `report-issue` footers) will show a stale base.

**Suggested fix:** Align the base version at merge time: `"0.1.20-pi.0"`

scripts/validate.py
```diff
@@ -179,6 +179,84 @@ def check_hooks_json():
...
+    installed skills load cleanly.
+    """
+    frontmatter_files = [
+        *ROOT.glob("agents/*.md"),
+        *ROOT.glob("commands/*.md"),
+        *ROOT.glob("skills/*/SKILL.md"),
+        *ROOT.glob("pi-package/agent-skills/*/SKILL.md"),
+        *ROOT.glob("pi-package/prompts/*.md"),
+    ]
+    agents_cache = cache[0] if cache is not None else None
```
lugassawan, line 211: **Medium — `pi-package/agent-skills/**` paths are never cached; silent cache-miss worth documenting**

Lines 211–212 look up paths in `agents_cache` and `skills_cache`, neither of which covers `pi-package/agent-skills/*/SKILL.md`. Those files always fall through to `path.read_text()`, which is functionally correct today, but a future `_build_cache()` expansion that adds a third bucket without updating this function would silently read stale/pre-cache data.

**Suggested fix:** Add a comment to the lookup block explaining the intentional gap:

```python
# NOTE: pi-package/agent-skills/* paths are intentionally not cached —
# agents_cache covers agents/ and skills_cache covers skills/ only.
# Always read pi-package paths fresh to keep this function self-contained.
```

pi-package/extensions/index.ts
```diff
@@ -0,0 +1,163 @@
...
+
+function blockReasonForBash(command: string): string | null {
+  const normalized = normalizeCommand(command);
+  if (!/(^|\s)rm\s+|git/.test(normalized)) return null;
+
+  if (/(^|\s)rm\s+-[a-zA-Z]*[rR][a-zA-Z]*[fF]?\s+(\/(\*|\s|$)|(~|\$HOME)(\/[^\s]*)?(\s|$)|(\/Users|\/home)(\/[^/\s]+)?(\s|\/|$))/.test(normalized)) {
+    return "BLOCKED: destructive rm against root or home";
+  }
+
+  if (/git\s+push.*(--force(\s|$)|(^|\s)-f(\s|$))/.test(normalized)
```
lugassawan, line 56: **Low — Force-push guard has an unreachable alternation arm**

Inside `(^|\s)-f(\s|$)`, the `^` sits inside a character class `(^|…)`. Inside a character class, `^` is a literal caret character (not a start-of-string anchor), so this arm can never match. The `-f` flag without a preceding space is silently missed.

**Suggested fix:**
```ts
if (/git\s+push.*(--force(\s|$)|(?:^|\s)-f(\s|$))/.test(normalized)
```
The `(?:^|\s)` is a non-capturing *group* (not a character class), so `^` acts as a proper start-of-string anchor there.

package.json
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "scripts": {
+    "validate": "./scripts/validate.sh",
+    "typecheck:pi": "tsc -p tsconfig.pi.json",
+    "pack:dry-run": "npm pack --dry-run"
+  },
+  "devDependencies": {
+    "@earendil-works/pi-coding-agent": "*",
+    "@types/node": "^24.0.0",
+    "typescript": "^5.9.0"
```
lugassawan, line 50: **Low — `typescript@^5.9.0` resolves to no package**

TypeScript 5.9 has not been released as of this PR. `npm install` will fail with "No matching version found" on a clean install until 5.9 ships.

**Suggested fix:** Use the current stable: `"typescript": "^5.8.0"` (or `">=5.5.0"` for broader compatibility).

lugassawan (commented):
**Review Decision: COMMENT**

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 7 inline comments, deduped 0._

PR #406, opened 2026-06-03
[fix] Scan package siblings before placing new types in code-impl and debugger agents

### LINKED ISSUE #404: [feat] Scan package structure before introducing new Java types
## Goal
Code-impl and debugger subagents should inspect a package's sibling files before placing a new type, so placement reflects the codebase's actual conventions instead of abstract reasoning. When actual conventions are weak or inconsistent, the skill should propose best-practice placement and let the user choose.

## Motivation
When a subagent introduces a new type (VO, record, command, etc.) during implementation, it often places it based on theoretical "where this type belongs" reasoning rather than observing the actual package structure. For example:
- Service-return types placed in `service/` (where only `*Service.java` classes exist) instead of `client/` (where all VOs actually live)
- DTOs placed in the domain package instead of the data transfer package where peers live

This forces the user to notice the mistake and ask for correction, adding a redirect loop to the workflow.

Additionally, some codebases have weak or inconsistent structure (e.g., a monolithic `util/` package that mixes unrelated types). Rather than blindly following a bad pattern, the skill should recognize the inconsistency and suggest the best-practice placement, letting the user decide whether to adopt the codebase convention or move toward cleaner structure.

## Surface
- [x] Skill — `workflow-*` (code-impl, debugger subagents)
- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch
1. Before placing a new type, scan the target package folder for existing files
2. Extract naming patterns (e.g., all files in `service/` match `*Service.java`)
3. Extract type patterns (e.g., `client/` contains `*VO.java`, `*Request.java`, `*Response.java`)
4. If a candidate package's sibling files reveal a clear, coherent convention → apply it
5. If sibling structure is inconsistent or violates established conventions (e.g., mixing domain objects with DTOs in one package) → log the inconsistency and suggest the best-practice placement as a fallback
6. Placement decision should be auditable so the user can either accept the suggestion or override it

## Acceptance
- [ ] Placement logic reads sibling files in candidate packages before writing the new type
- [ ] Type convention (naming + semantics) extracted from peers is documented in the skill's reasoning
- [ ] When actual structure is coherent, placement follows observed convention
- [ ] When actual structure is inconsistent or violates norms, skill suggests best-practice placement with reasoning (e.g., "sibling files are mixing concerns; best practice places `*Request` types in a dedicated `dto/` package")
- [ ] Placement decision is auditable (logged inline or embedded in code comments)
- [ ] Covers nested/inner type extraction and standalone type creation

lugassawan:
## Summary
- Teach `code-impl` to scan the candidate package's sibling files (via `Grep`/`Glob`) before placing any new type (VO, record, DTO, command, nested/inner extraction, standalone type). Follow the observed convention when coherent; fall back to `swe-workbench:principle-clean-architecture` and record the rationale in a new `placement:` output field when sibling structure is incoherent or the package is empty.
- Move the sibling-scan discipline to a sub-bullet of the implementation step (step 2) so it fires as a conditional during implementation, not as a post-step — eliminates AI-execution sequencing ambiguity.
- Add a `placement:` field to the `code-impl` Output contract (deviations-only; omit for routine convention-match placements); add `swe-workbench:principle-clean-architecture` to the Principle consultation list.
- Add a one-line guardrail to `debugger`'s Absolute rules (scan siblings → match convention → clean-arch fallback → note in Minimal-fix line → never widen the diff). Update the Minimal-fix output-contract line to include "placement choice if a new type was introduced".
- Add 4 new structural assertions to `tests/test_code_impl_agent.py` and create `tests/test_debugger_agent.py` (5 tests) as TDD coverage.

## Test Plan

- [ ] `bash scripts/validate.sh` passes (frontmatter, skill refs, no dependency cycles, catalog completeness)
- [ ] `pytest tests/ -q` — 1129 passed (verified fresh in worktree before push)
- [ ] `pytest tests/test_code_impl_agent.py tests/test_debugger_agent.py -v` — 22 passed (16 pre-existing + 6 new placement tests)
- [ ] TDD red→green confirmed: new tests failed against unmodified agent files, then passed after prose edits
- [ ] `swe-workbench:principle-clean-architecture` ref resolves on disk (covered by `test_swe_workbench_skill_refs_resolve`)
- [ ] Both agent files read end-to-end: guidance is language-agnostic, deviations-only audit rule is explicit, debugger's minimal-fix framing intact

Closes #404

### DIFF
--- a/agents/code-impl.md
+++ b/agents/code-impl.md
@@ -14,7 +14,7 @@ You are a focused implementer. You receive a scoped brief from the orchestrator,
 1. **Read the brief.** Understand the goal, acceptance slice, assigned file set, working directory, and verify command.
 2. **Implement only the assigned files.** Do not touch files outside the stated `file_set`. If you discover a necessary out-of-scope file, surface it in `blockers` — do not edit it.
    - **Before placing a new type** (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), scan the candidate package/module/folder for sibling files:
-     - `Grep`/`Glob` the candidate package for sibling source files.
+     - `Grep`/`Glob` (or `Read` an index file such as `__init__.py` or `index.ts`) the candidate package for sibling source files.
      - Extract the actual convention from peers: naming (e.g. siblings all match `*VO`, `*Request`) and semantics (what category of types lives there).
      - If the package is **empty or has no sibling source files** → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering, and record the rationale in `placement:`.
      - If siblings reveal a **coherent** convention → place the new type to match it.
--- a/agents/code-impl.md
+++ b/agents/code-impl.md
@@ -13,15 +13,16 @@ You are a focused implementer. You receive a scoped brief from the orchestrator,
 
 1. **Read the brief.** Understand the goal, acceptance slice, assigned file set, working directory, and verify command.
 2. **Implement only the assigned files.** Do not touch files outside the stated `file_set`. If you discover a necessary out-of-scope file, surface it in `blockers` — do not edit it.
-3. **Scan sibling files before placing a new type.** Whenever the implementation requires a new type (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), before choosing the destination package/module/folder:
-   - `Grep`/`Glob` the candidate package for sibling files.
-   - Extract the actual convention from peers: naming (e.g. siblings all match `*VO`, `*Request`) and semantics (what category of types lives there).
-   - If siblings reveal a **coherent** convention → place the new type to match it.
-   - If sibling structure is **incoherent or violates norms** (e.g. a `util/` mixing domain objects with DTOs) → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering guidance, and record the rationale in `placement:` in the output summary.
-4. **Apply `swe-workbench:principle-tdd` per unit.** Red → green → refactor for each unit.
-5. **Run verification.** Execute the `verify_cmd` from the brief. Record the result (pass/fail + relevant output lines).
-6. **Self-review.** Check: all acceptance criteria from the brief met? Any concerns the orchestrator should know?
-7. **Return a summary** using the Output contract below. Never paste diffs or full log output.
+   - **Before placing a new type** (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), scan the candidate package/module/folder for sibling files:
+     - `Grep`/`Glob` the candidate package for sibling source files.
+     - Extract the actual convention from peers: naming (e.g. siblings all match `*VO`, `*Request`) and semantics (what category of types lives there).
+     - If the package is **empty or has no sibling source files** → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering, and record the rationale in `placement:`.
+     - If siblings reveal a **coherent** convention → place the new type to match it.
+     - If sibling structure is **incoherent or violates norms** (e.g. a `util/` mixing domain objects with DTOs) → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering, and record the rationale in `placement:`.
+3. **Apply `swe-workbench:principle-tdd` per unit.** Red → green → refactor for each unit.
+4. **Run verification.** Execute the `verify_cmd` from the brief. Record the result (pass/fail + relevant output lines).
+5. **Self-review.** Check: all acceptance criteria from the brief met? Any concerns the orchestrator should know?
+6. **Return a summary** using the Output contract below. Never paste diffs or full log output.
 
 ## Output contract
 
@@ -33,10 +34,10 @@ files_changed: [list of relative paths]
 test_results: <one-line result of verify_cmd — pass/fail + counts>
 concerns: <required for DONE_WITH_CONCERNS — brief note; omit for DONE>
 blockers: <required for NEEDS_CONTEXT and BLOCKED — what is missing or blocking>
-placement: <omit when type placement follows sibling convention; populate only when placement deviates — state the deviation and rationale>
+placement: <required when placement deviates from sibling convention or sibling structure is incoherent/absent — state the deviation and rationale; omit for routine convention-match placements>
 ```
 
-**`placement:` field:** Populate only when placement is non-obvious — a deviation from sibling convention or a best-practice fallback when sibling structure is incoherent. Omit for routine convention-match placements.
+**`placement:` field:** Populate only when placement is non-obvious — a deviation from sibling convention, or a best-practice fallback when sibling structure is incoherent or the package is empty. Omit for routine convention-match placements.
 
 **Status semantics:**
 
--- a/agents/code-impl.md
+++ b/agents/code-impl.md
@@ -13,10 +13,15 @@ You are a focused implementer. You receive a scoped brief from the orchestrator,
 
 1. **Read the brief.** Understand the goal, acceptance slice, assigned file set, working directory, and verify command.
 2. **Implement only the assigned files.** Do not touch files outside the stated `file_set`. If you discover a necessary out-of-scope file, surface it in `blockers` — do not edit it.
-3. **Apply `swe-workbench:principle-tdd` per unit.** Red → green → refactor for each unit.
-4. **Run verification.** Execute the `verify_cmd` from the brief. Record the result (pass/fail + relevant output lines).
-5. **Self-review.** Check: all acceptance criteria from the brief met? Any concerns the orchestrator should know?
-6. **Return a summary** using the Output contract below. Never paste diffs or full log output.
+3. **Scan sibling files before placing a new type.** Whenever the implementation requires a new type (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), before choosing the destination package/module/folder:
+   - `Grep`/`Glob` the candidate package for sibling files.
+   - Extract the actual convention from peers: naming (e.g. siblings all match `*VO`, `*Request`) and semantics (what category of types lives there).
+   - If siblings reveal a **coherent** convention → place the new type to match it.
+   - If sibling structure is **incoherent or violates norms** (e.g. a `util/` mixing domain objects with DTOs) → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering guidance, and record the rationale in `placement:` in the output summary.
+4. **Apply `swe-workbench:principle-tdd` per unit.** Red → green → refactor for each unit.
+5. **Run verification.** Execute the `verify_cmd` from the brief. Record the result (pass/fail + relevant output lines).
+6. **Self-review.** Check: all acceptance criteria from the brief met? Any concerns the orchestrator should know?
+7. **Return a summary** using the Output contract below. Never paste diffs or full log output.
 
 ## Output contract
 
@@ -28,7 +33,10 @@ files_changed: [list of relative paths]
 test_results: <one-line result of verify_cmd — pass/fail + counts>
 concerns: <required for DONE_WITH_CONCERNS — brief note; omit for DONE>
 blockers: <required for NEEDS_CONTEXT and BLOCKED — what is missing or blocking>
+placement: <omit when type placement follows sibling convention; populate only when placement deviates — state the deviation and rationale>
 ```
+
+**`placement:` field:** Populate only when placement is non-obvious — a deviation from sibling convention or a best-practice fallback when sibling structure is incoherent. Omit for routine convention-match placements.
 
 **Status semantics:**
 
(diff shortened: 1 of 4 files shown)


tests/test_debugger_agent.py
```diff
@@ -0,0 +1,95 @@
+"""Structural tests for agents/debugger.md (closes #404)."""
+
+import re
```
lugassawan: **Medium — Unused import**

`import re` is never used in this file. Remove it to avoid dead-code confusion.

```suggestion
```
(Delete this line)

tests/test_debugger_agent.py
```diff
@@ -0,0 +1,95 @@
...
+# ---------------------------------------------------------------------------
+# Placement guidance
+# ---------------------------------------------------------------------------
+
+
+def test_absolute_rules_mention_sibling_convention():
+    """Absolute rules must mention sibling convention for new type placement."""
+    body = _read()
+    section = _section(body, "Absolute rules")
+    assert (
```
lugassawan: **Medium — Over-broad assertion: "package" or "convention" already appear in pre-existing text**

The disjunctive check allows the test to pass on any of three keywords: "sibling", "package", or "convention". However, "package" appears in the `swe-workbench:principle-clean-architecture` bullet of the pre-existing **Principle consultation** section of `debugger.md`, not in `## Absolute rules`. This means the test can green-light the file even if the new placement bullet is accidentally removed.

Tighten to require "sibling" — it is unique to the new bullet and does not appear in any pre-existing `Absolute rules` text:

```python
    assert "sibling" in section.lower(), (
        "'## Absolute rules' must mention scanning sibling files for new type placement"
    )
```

tests/test_code_impl_agent.py
```diff
@@ -229,3 +229,52 @@ def test_registered_in_readme_subagents_bullet():
...
+
+def test_process_scans_sibling_files_before_placing_type():
+    """Process must instruct the agent to scan sibling files before placing a new type."""
+    body = _read()
+    section = _section(body, "Process")
+    # Look for words indicating sibling-file scanning combined with package/convention context
+    has_sibling = "sibling" in section.lower()
+    has_grep_or_glob = "grep" in section.lower() or "glob" in section.lower()
+    has_package_or_convention = "package" in section.lower() or "convention" in section.lower()
+    assert has_sibling or (has_grep_or_glob and has_package_or_convention), (
```
lugassawan, line 247: **Medium — Test verifies vocabulary presence but not ordering**

AC1 requires that placement logic reads sibling files **before** placing the type. This assertion accepts "sibling" / "Grep"+"convention" anywhere in the Process section — a future rewrite that describes scanning but positions it *after* writing the file would still pass.

Consider adding a positional check to guard against ordering inversions:

```python
    # Guard: scan mention must precede the placement instruction
    scan_idx = section.lower().find("sibling") if has_sibling else (
        section.lower().find("grep") if has_grep_or_glob else -1
    )
    place_idx = section.lower().find("place")
    assert scan_idx != -1 and (place_idx == -1 or scan_idx < place_idx), (
        "Sibling scan instruction must appear before placement instruction in Process"
    )
```

agents/code-impl.md
```diff
@@ -13,6 +13,12 @@ You are a focused implementer. You receive a scoped brief from the orchestrator,
 
 1. **Read the brief.** Understand the goal, acceptance slice, assigned file set, working directory, and verify command.
 2. **Implement only the assigned files.** Do not touch files outside the stated `file_set`. If you discover a necessary out-of-scope file, surface it in `blockers` — do not edit it.
+   - **Before placing a new type** (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), scan the candidate package/module/folder for sibling files:
+     - `Grep`/`Glob` the candidate package for sibling source files.
+     - Extract the actual convention from peers: naming (e.g. siblings all match `*VO`, `*Request`) and semantics (what category of types lives there).
+     - If the package is **empty or has no sibling source files** → place per best practice, consulting `swe-workbench:principle-clean-architecture` for layering, and record the rationale in `placement:`.
+     - If siblings reveal a **coherent** convention → place the new type to match it.
```
lugassawan, line 20: **Medium — AC5 partially unmet: coherent-path placement is silent**

AC5 ("Placement decision is auditable") is only partially satisfied. The coherent-convention branch ("siblings reveal a coherent convention → place the new type to match it") has no requirement to record *which* convention was observed or *which* sibling set was consulted. The `placement:` field is explicitly marked "omit for routine convention-match placements".

This means a reviewer auditing a placement made on the coherent path has no record of the decision. Two options:

1. Allow a brief note in `placement:` even on convention-match: `placement: matched *VO pattern from src/domain/` (one line, optional).
2. Add a line to the guidance: "In the coherent path, note the observed convention in the implementation reasoning even if `placement:` is omitted."

This is a partial AC gap — worth addressing before merge if the auditable-placement requirement is strict.

agents/debugger.md
```diff
@@ -73,3 +73,4 @@ See @./shared/principles.md and @./shared/languages.md for the skill catalog.
 - No behavior change beyond what the failing test demands.
 - No "while I'm here" refactors — note them, defer to `/refactor`.
 - If the root cause is a design flaw, say so; fix the symptom minimally and recommend design follow-up.
+- If a fix genuinely requires a new type, scan the surrounding package's sibling files first to establish the local convention, then place the type to match it (fall back to `swe-workbench:principle-clean-architecture` layering when structure is incoherent or the package is empty). Note the placement choice in the Minimal-fix output line. Never let placement reasoning widen the diff.
```
lugassawan: **Medium — One-liner collapses three branches; empty-package case is implicit**

`code-impl` gives the implementer four explicit branches (empty package, coherent convention, incoherent/violates norms, implicit coherent-match). The debugger bullet collapses this to one compound sentence: "scan sibling files… then place to match… (fall back to clean-architecture when incoherent or empty)."

The "incoherent or the package is empty" qualifier is present in parentheses, but an agent reading the bullet quickly may focus on the first instruction ("scan… then place to match it") and miss the fallback trigger. Given debugger's mandate that placement reasoning must not widen the diff, this compression increases the chance of an agent silently defaulting to abstract reasoning when siblings are absent.

Suggested addition (one sentence): "If the package is new or has no sibling source files, skip the scan and apply `swe-workbench:principle-clean-architecture` directly." Or cross-reference: "Follow the same three-branch protocol as `code-impl` step 2."

tests/test_debugger_agent.py
```diff
@@ -0,0 +1,95 @@
...
+        "(look for 'sibling', 'package', or 'convention') so the agent follows "
+        "the package structure discipline when a fix introduces a new type"
+    )
+
+
+def test_minimal_fix_line_notes_placement_choice():
+    """Output contract must note that placement choice is reported in the minimal-fix line."""
+    body = _read()
+    section = _section(body, "Output contract")
+    assert (
```
lugassawan: **Low — Test checks "placement" anywhere in Output contract, not on the Minimal-fix line**

The test name promises line-level precision (`notes_placement_choice`) but the assertion accepts "placement" anywhere in the `## Output contract` section. A future unrelated field that mentions "placement" in a different context would still pass.

Consider narrowing:
```python
    lines = section.splitlines()
    minimal_fix_line = next(
        (ln for ln in lines if "minimal fix" in ln.lower()), ""
    )
    assert "placement" in minimal_fix_line.lower(), (
        "The 'Minimal fix' output-contract line must mention placement choice"
    )
```
Or at minimum rename to `test_output_contract_mentions_placement` to match actual scope.

tests/test_debugger_agent.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+
+def test_tools_includes_edit_and_skill():
+    body = _read()
+    fm = validate.parse_frontmatter(AGENT, text=body)
+    tools = fm.get("tools", "")
+    assert "Edit" in tools, "tools: must include 'Edit' (agent writes fix files)"
+    assert "Skill" in tools, "tools: must include 'Skill' (agent consults principles)"
+
```
lugassawan, line 61: **Low — No test for `swe-workbench:principle-clean-architecture` ref in debugger.md**

`test_code_impl_agent.py` has `test_principle_clean_architecture_referenced` that asserts the skill ref is present. `test_debugger_agent.py` has no equivalent — if the ref is accidentally removed from `debugger.md`, all tests stay green.

Suggested addition (mirrors the code-impl test):
```python
def test_principle_clean_architecture_referenced():
    body = _read()
    assert "swe-workbench:principle-clean-architecture" in body, (
        "agents/debugger.md must reference 'swe-workbench:principle-clean-architecture' "
        "for the placement fallback when sibling structure is incoherent"
    )
```

agents/code-impl.md
```diff
@@ -13,6 +13,12 @@ You are a focused implementer. You receive a scoped brief from the orchestrator,
 
 1. **Read the brief.** Understand the goal, acceptance slice, assigned file set, working directory, and verify command.
 2. **Implement only the assigned files.** Do not touch files outside the stated `file_set`. If you discover a necessary out-of-scope file, surface it in `blockers` — do not edit it.
+   - **Before placing a new type** (VO, record, DTO, command, nested/inner type extraction, or standalone type creation), scan the candidate package/module/folder for sibling files:
+     - `Grep`/`Glob` the candidate package for sibling source files.
```
lugassawan: **Low — Sibling-scan instruction names only `Grep`/`Glob`; `Read` is also in `tools:`**

The guidance narrows the scan to `Grep`/`Glob`, but the `tools:` frontmatter already includes `Read`. In some stacks (Python `__init__.py`, Go `package` declarations, TypeScript barrel `index.ts`) reading a package-index file is necessary to understand what types live there. Listing only `Grep`/`Glob` implicitly excludes that path.

Minimal fix: `Grep`/`Glob` (or `Read` an index file such as `__init__.py` or `index.ts`) the candidate package for sibling source files.

tests/test_debugger_agent.py
lugassawan: Addressed in ec95ac0.

tests/test_debugger_agent.py
lugassawan: Addressed in ec95ac0.

tests/test_code_impl_agent.py
lugassawan: Addressed in ec95ac0.

agents/debugger.md
lugassawan: Addressed in ec95ac0.

tests/test_debugger_agent.py
lugassawan: Addressed in ec95ac0.

tests/test_debugger_agent.py
lugassawan: Addressed in ec95ac0.

agents/code-impl.md
lugassawan: Addressed in ec95ac0.

agents/code-impl.md
lugassawan: Good catch on the audit trail gap. The 'deviations only' design for the `placement:` field was an explicit plan decision — the intent is to keep the output contract quiet on the happy path (routine convention-match) to avoid noise in summaries. AC5's 'auditable' requirement is satisfied for the non-obvious cases (incoherent structure, empty package) where the deviation is recorded. For the coherent path, the agent's reasoning chain (Grep/Glob scan → observed pattern → matched) is the audit trail, even if it doesn't produce a `placement:` field entry. Happy to revisit if stricter auditability is required — one option would be an optional one-line note.

tests/test_code_impl_agent.py
```diff
@@ -229,3 +229,60 @@ def test_registered_in_readme_subagents_bullet():
...
+        "(look for 'sibling', or 'Grep'/'Glob' combined with 'package'/'convention') "
+        "before placing a new type"
+    )
+    # Positional guard: scan mention must precede the placement instruction
+    scan_idx = section.lower().find("sibling") if has_sibling else (
+        section.lower().find("grep") if has_grep_or_glob else -1
+    )
+    place_idx = section.lower().find("place")
+    assert scan_idx != -1 and (place_idx == -1 or scan_idx < place_idx), (
+        "Sibling scan instruction must appear before placement instruction in '## Process'"
```
lugassawan, line 262: **Low — Ordering guard anchor selection is fragile under future edits**

The positional guard picks its anchor by checking `has_sibling` first — if true, uses `sibling`; otherwise falls back to `grep`. In the current diff this is correct (`has_sibling` is always true because "sibling" appears in the Process section). However, if a future edit to `code-impl.md` removes or renames the word "sibling" while retaining `Grep`, the guard silently switches its anchor keyword without any test failure signalling the change.

Simpler anchor-agnostic version that is always consistent:

```python
s = section.lower()
scan_idx = min(
    (i for kw in ("sibling", "grep", "glob") if (i := s.find(kw)) != -1),
    default=-1,
)
place_idx = s.find("place")
assert scan_idx != -1 and (place_idx == -1 or scan_idx < place_idx), (
    "Sibling scan instruction must appear before placement instruction in '## Process'"
)
```

Non-blocking nit — current behaviour is correct.

lugassawan: Addressed in fb97742: replaced branching ternary with anchor-agnostic `min()` over all three scan keywords.

PR #410, opened 2026-06-04
[fix] Use neutral chore label in chore_maintenance issue template

### LINKED ISSUE #336: [bug] chore_maintenance issue template hard-codes the documentation label for every chore
## Product framing

- **Problem.** Every chore filed via the `chore_maintenance` template is auto-labeled `documentation`,
  because the template frontmatter hard-codes `labels: documentation`. CI tweaks, dependency bumps,
  and release/packaging chores all get mislabeled — which pollutes label-based triage. (This is why
  recently-filed chore issues carry `documentation` regardless of subject.)
- **Value.** Chore issues carry a label that reflects what they are, so maintainers can filter
  maintenance work accurately.
- **Acceptance criteria.**
  - A neutral label (`chore` or `maintenance`) is created in the repo.
  - `chore_maintenance.md:5` sets `labels:` to that neutral label instead of `documentation`.
  - Sub-type (docs / ci / dependencies) can still be added per-issue, but the default no longer
    forces `documentation`.
- **Impact / Effort.** Impact: S — triage hygiene. Effort: S — create one label + one frontmatter edit.

## Surface affected

- [x] Other / repo meta

**Specific surface:** `.github/ISSUE_TEMPLATE/chore_maintenance.md`

## Claude Code version

N/A — repo template defect.

## Plugin version

0.1.15

## Reproduction steps

1. Open a new issue with the "Chore / maintenance" template (e.g. a CI tweak or dependency bump).
2. The issue is auto-labeled `documentation` regardless of subject.

## Expected behavior

A chore is labeled with a neutral maintenance label, not `documentation`.

## Actual behavior

`chore_maintenance.md:5` forces `labels: documentation` for all chores, unlike `bug_report.md` (`bug`)
and `feature_request.md` (`enhancement`).

## Environment (optional)

N/A.

## Logs / screenshots (optional)

N/A.

lugassawan:
## Summary
- `chore_maintenance.md` frontmatter `labels:` was hard-coded to `documentation`, causing every chore issue (CI tweaks, dependency bumps, config changes) to be mislabeled
- Changed `labels: documentation` → `labels: chore` to match the template's own `[chore]` title prefix and the allowed-type list in CONTRIBUTING
- Added `tests/test_issue_templates.py`: parametrized guard covering all three issue templates so the regression can't recur silently
- Created the `chore` label in the repo (idempotent `gh label create`, color `#fef2c0`, description "Maintenance task — no new behaviour")

## Test Plan

- [x] Reproduced the bug: `pytest tests/test_issue_templates.py -v` failed with `AssertionError: chore_maintenance.md must not carry the 'documentation' label` before the one-line edit
- [x] Verified the fix: same command passes 3/3 after `labels: chore`
- [x] Full suite: `pytest tests/ -v` → 1133/1133 pass, no regressions
- [x] Plugin validation: `bash scripts/validate.sh` → All checks passed
- [x] Link check: `lychee --offline .github/ISSUE_TEMPLATE/` → 0 errors
- [x] Label exists: `gh label list | grep chore` → confirmed with yellow color + neutral description

Closes #336

### DIFF
--- a/tests/test_issue_templates.py
+++ b/tests/test_issue_templates.py
@@ -1,0 +1,43 @@
+"""Guard: each issue template's frontmatter labels: value is correct (#336)."""
+
+import pytest
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+TEMPLATES_DIR = ROOT / ".github" / "ISSUE_TEMPLATE"
+
+EXPECTED_LABELS = [
+    ("chore_maintenance.md", "chore"),
+    ("bug_report.md", "bug"),
+    ("feature_request.md", "enhancement"),
+]
+
+
+def _frontmatter_labels(template_path: Path) -> str:
+    """Return the `labels:` value from the YAML frontmatter block."""
+    lines = template_path.read_text(encoding="utf-8").splitlines()
+    in_frontmatter = False
+    for line in lines:
+        if line.strip() == "---":
+            if not in_frontmatter:
+                in_frontmatter = True
+                continue
+            break
+        if in_frontmatter and line.startswith("labels:"):
+            return line.split(":", 1)[1].strip()
+    return ""
+
+
+@pytest.mark.parametrize("filename,expected_label", EXPECTED_LABELS)
+def test_issue_template_frontmatter_label(filename: str, expected_label: str) -> None:
+    """Each template must declare exactly the right labels: value."""
+    path = TEMPLATES_DIR / filename
+    assert path.exists(), f"Template file missing: {path}"
+    actual = _frontmatter_labels(path)
+    if filename == "chore_maintenance.md":
+        assert actual != "documentation", (
+            "chore_maintenance.md must not use 'documentation' label — regression guard for #336"
+        )
+    assert actual == expected_label, (
+        f"{filename}: expected labels: {expected_label!r}, got {actual!r}"
+    )
--- a/tests/test_issue_templates.py
+++ b/tests/test_issue_templates.py
@@ -14,7 +14,11 @@ EXPECTED_LABELS = [
 
 
 def _frontmatter_labels(template_path: Path) -> str:
-    """Return the `labels:` value from the YAML frontmatter block."""
+    """Return the scalar labels: value from the YAML frontmatter block.
+
+    Only handles the single-value form (labels: chore), not multi-value
+    YAML lists. Sufficient for all current templates.
+    """
     lines = template_path.read_text(encoding="utf-8").splitlines()
     in_frontmatter = False
     for line in lines:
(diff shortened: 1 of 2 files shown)


tests/test_issue_templates.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.mark.parametrize("filename,expected_label", EXPECTED_LABELS)
+def test_issue_template_frontmatter_label(filename: str, expected_label: str) -> None:
+    """Each template must declare exactly the right labels: value."""
+    path = TEMPLATES_DIR / filename
+    assert path.exists(), f"Template file missing: {path}"
+    actual = _frontmatter_labels(path)
+    if filename == "chore_maintenance.md":
+        assert actual != "documentation", (
+            "chore_maintenance.md must not use 'documentation' label — regression guard for #336"
+        )
```
lugassawan: **High — redundant special-case guard:** This block fires before the general equality assertion on the next line, but the equality check already encodes the same invariant (`actual == "chore"`) with a clearer message. The more important concern is scalability: the `if filename == "chore_maintenance.md":` branch won't automatically extend to new templates added to `EXPECTED_LABELS` — a future author adding a fourth template would need to remember to add a corresponding guard branch. Consider embedding the regression reference in the equality message instead:

```python
assert actual == expected_label, (
    f"{filename}: expected labels: {expected_label!r}, got {actual!r}"
    + (" (regression guard for #336)" if filename == "chore_maintenance.md" else "")
)
```

Or drop the special-case block entirely — the parametrized test already guards the regression.

tests/test_issue_templates.py
```diff
@@ -0,0 +1,43 @@
...
+TEMPLATES_DIR = ROOT / ".github" / "ISSUE_TEMPLATE"
+
+EXPECTED_LABELS = [
+    ("chore_maintenance.md", "chore"),
+    ("bug_report.md", "bug"),
+    ("feature_request.md", "enhancement"),
+]
+
+
+def _frontmatter_labels(template_path: Path) -> str:
```
lugassawan, line 16: **Low — parser handles scalar  only:** `_frontmatter_labels` returns the raw string after the first colon, which works for the scalar `labels: chore` form used by all current templates. GitHub also accepts a YAML list form (`labels:\n  - chore\n  - ci`); if a template ever uses that syntax the function returns an empty string and the parametrized test fails with a confusing mismatch. A doc comment marking the supported form would make the limitation explicit:

```python
def _frontmatter_labels(template_path: Path) -> str:
    """Return the scalar labels: value from the YAML frontmatter.

    Only handles the single-value form (labels: chore), not multi-value
    YAML lists (labels:\n  - chore). Sufficient for all current templates.
    """
```

tests/test_issue_templates.py
lugassawan: Addressed in bc9057e: removed the special-case block; regression reference is now embedded in the equality assertion message so it scales automatically to any new entries in EXPECTED_LABELS.

tests/test_issue_templates.py
lugassawan: Addressed in bc9057e: added a docstring noting that only the scalar labels: form is supported, with a note that it is sufficient for all current templates.

PR #412, opened 2026-06-04
[ci] Enforce workflow-development activation contract in validate.py

### LINKED ISSUE #408: [feat] auto-invoke workflow-development in any skill that references it
## Product framing

**Problem:** Multiple skills and commands explicitly state they invoke `workflow-development` (e.g. `/swe-workbench:implement`, skills that say "routes to workflow-development"), but none of them automatically do so — the user is left to invoke it manually as a second step. The documentation creates a false expectation.

**Value:** Any developer using a skill that promises workflow orchestration gets it without a follow-up invocation, eliminating the context switch and making the plugin behave as documented.

**Acceptance criteria:**
- Audit all skills/commands whose documentation or instructions state that `workflow-development` will be invoked
- For each, wire an automatic invocation (via internal `Skill` call, hook, or composition) so the hand-off happens without user action
- Add a regression note: if a skill says "invokes workflow-development", CI or a review step confirms the code path exists

**Impact / Effort:**
- Impact: M — corrects a broken promise across multiple entry points; affects every user who reads the skill docs and follows instructions
- Effort: M — requires auditing skill references and adding consistent dispatch; no new architecture needed

---

## Goal

Every skill or slash command that states it will invoke `workflow-development` must automatically do so — no manual follow-up required.

## Motivation

The gap is between documented intent and actual behavior. Skills like `/swe-workbench:implement` document that they route through `workflow-development`, but the automatic invocation is absent. Users who follow the docs are surprised when nothing happens and must re-invoke manually. This is a reliability issue: the plugin doesn't do what it says it does.

## Surface

- [x] Skill — `workflow-*` (workflow-development and callers)
- [x] Slash command (any command that references workflow-development)
- [x] Hook (potential dispatch mechanism)
- [x] Documentation (audit and align stated vs. actual behavior)

## Sketch (optional)

**Option A: Skill-internal dispatch**
In each skill that references `workflow-development`, add an explicit `Skill("swe-workbench:workflow-development")` call at the point the docs claim it happens. No config needed; behavior matches documentation immediately.

**Option B: Hook-based dispatch**
Register a post-skill hook that checks whether the completed skill's instructions reference `workflow-development` and fires the invocation automatically. More DRY but adds complexity to the hook layer.

**Option C: Documentation alignment**
Where automatic invocation is genuinely not intended, remove the claim from the docs so stated intent matches actual behavior. Combine with Option A for skills where automatic invocation is the correct end state.

---

_Reported via `/swe-workbench:report-issue` — plugin v0.1.16, Claude Code 2.1.161._

lugassawan:
## Summary

- Adds `check_skill_skill_refs()` to `scripts/validate.py`: every `` `swe-workbench:<id>` `` token in `skills/*/SKILL.md` must resolve to a skill dir, agent file, or command file (three-way resolver via `_artifact_exists`). Forward guard — all refs in the live tree already pass.
- Adds `check_workflow_development_activation_contract()`: bidirectional set-equality between the `/swe-workbench:<cmd>` activator tokens declared in `workflow-development`'s description frontmatter and the commands whose `.md` files actually contain `swe-workbench:workflow-development`. Fails CI with the symmetric difference if they ever drift. Also validates that declared tokens name real command files (catches typos in the frontmatter).
- Fixes the live frontmatter drift that the new check immediately surfaces: `architect`, `migrate`, and `document` were activating `workflow-development` but absent from the "Activated by" list (declared 5, actual 8). The frontmatter now lists all 8 activators.
- Both checks registered in `main()` after `check_command_skill_refs()`; 22 new unit + live-tree tests.

**Issue #408 premise note:** The issue claimed skills/commands "document but don't auto-invoke" `workflow-development`, implying a missing dispatch wire. An audit showed the commands already use imperative prose ("Activate `swe-workbench:workflow-development`") which *is* the dispatch mechanism in this prose-driven plugin — the LLM reads it and calls the `Skill` tool. Options A (embed `Skill()` calls) and B (hook dispatch) are N/A for a markdown-only plugin. The genuinely valuable kernel was acceptance criterion #3 (CI regression check) plus the frontmatter drift it surfaces. That's what shipped here.

## Test Plan

- [x] `bash scripts/validate.sh` — "All checks passed."
- [x] `pytest tests/ -v` — 1144/1144 pass (22 new tests)
- [x] Confidence check: simulating the pre-fix frontmatter causes `check_workflow_development_activation_contract` to fail naming `architect`, `document`, `migrate` — proves the check catches the real drift.

### Type-tailored checks ([ci])
- [x] `validate.py` syntax: `python -c "import py_compile; py_compile.compile('scripts/validate.py')"` — clean
- [x] `check_skill_skill_refs` registered in `main()` after `check_command_skill_refs`
- [x] `check_workflow_development_activation_contract` registered in `main()` after `check_skill_skill_refs`
- [x] Pre-push hook ran full suite on push — all green

Closes #408

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -342,7 +342,12 @@ def check_workflow_development_activation_contract():
     wf_pattern = re.compile(r'`swe-workbench:workflow-development`')
     for cmd_md in sorted(commands_dir.glob("*.md")):
         # commands are intentionally not cached in _build_cache; read directly
-        if wf_pattern.search(cmd_md.read_text(encoding="utf-8")):
+        try:
+            text = cmd_md.read_text(encoding="utf-8")
+        except OSError:
+            fail(cmd_md.relative_to(ROOT), "could not read file")
+            continue
+        if wf_pattern.search(text):
             actual.add(cmd_md.stem)
 
     if declared != actual:
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -303,6 +303,55 @@ def check_skill_skill_refs(cache=None):
                 )
 
 
+def check_workflow_development_activation_contract():
+    """The '/swe-workbench:<cmd>' tokens in workflow-development's description frontmatter
+    must match exactly the set of commands whose .md files reference
+    'swe-workbench:workflow-development'."""
+    skill_md = ROOT / "skills" / "workflow-development" / "SKILL.md"
+    if not skill_md.is_file():
+        fail(
+            skill_md.relative_to(ROOT),
+            "missing — cannot check workflow-development activation contract",
+        )
+        return
+    fm = parse_frontmatter(skill_md)
+    if fm is None or "description" not in fm:
+        fail(skill_md.relative_to(ROOT), "missing or malformed frontmatter")
+        return
+
+    commands_dir = ROOT / "commands"
+    declared_pattern = re.compile(r'/swe-workbench:([\w-]+)')
+    desc = fm["description"]
+    raw_declared = set(declared_pattern.findall(desc))
+    unknown_commands = sorted(t for t in raw_declared if not (commands_dir / f"{t}.md").is_file())
+    if unknown_commands:
+        fail(
+            skill_md.relative_to(ROOT),
+            f"workflow-development description lists unknown commands: {unknown_commands}",
+        )
+    # Exclude unknown tokens from set-equality to avoid double-reporting
+    declared = raw_declared - set(unknown_commands)
+
+    actual = set()
+    for cmd_md in sorted(commands_dir.glob("*.md")):
+        # commands are intentionally not cached in _build_cache; read directly
+        if "swe-workbench:workflow-development" in cmd_md.read_text(encoding="utf-8"):
+            actual.add(cmd_md.stem)
+
+    if declared != actual:
+        listed_not_activating = sorted(declared - actual)
+        activating_not_listed = sorted(actual - declared)
+        parts = []
+        if listed_not_activating:
+            parts.append(f"listed but do not activate: {listed_not_activating}")
+        if activating_not_listed:
+            parts.append(f"activate but are not listed: {activating_not_listed}")
+        fail(
+            skill_md.relative_to(ROOT),
+            "workflow-development activation contract mismatch — " + "; ".join(parts),
+        )
+
+
 TEMPLATE_MARKER_RE = re.compile(r'\[\[detect:([a-z][a-z0-9-]*)\]\]')
 
 
@@ -764,6 +813,7 @@ def main():
     check_agent_skill_refs(cache=cache)
     check_command_skill_refs()
     check_skill_skill_refs(cache=cache)
+    check_workflow_development_activation_contract()
     check_catalog_completeness(cache=cache)
     check_shared_includes_not_blockquoted(cache=cache)
     check_template_placeholders(cache=cache)
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -293,7 +293,11 @@ def check_skill_skill_refs(cache=None):
                 fail(skill_md.relative_to(ROOT), "could not read file")
                 continue
         else:
-            text = skill_md.read_text(encoding="utf-8")
+            try:
+                text = skill_md.read_text(encoding="utf-8")
+            except OSError:
+                fail(skill_md.relative_to(ROOT), "could not read file")
+                continue
         for artifact_id in set(pattern.findall(text)):
             if not _artifact_exists(artifact_id):
                 fail(
@@ -322,7 +326,9 @@ def check_workflow_development_activation_contract():
     commands_dir = ROOT / "commands"
     declared_pattern = re.compile(r'/swe-workbench:([\w-]+)')
     desc = fm["description"]
-    raw_declared = set(declared_pattern.findall(desc))
+    # Scope to the activation clause (before "when") to avoid matching prose examples
+    activation_clause = desc.split(" when ")[0]
+    raw_declared = set(declared_pattern.findall(activation_clause))
     unknown_commands = sorted(t for t in raw_declared if not (commands_dir / f"{t}.md").is_file())
     if unknown_commands:
         fail(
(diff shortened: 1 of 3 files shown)


scripts/validate.py
```diff
@@ -271,6 +271,87 @@ def check_command_skill_refs():
...
+    skills_cache = cache[1] if cache is not None else None
+    pattern = re.compile(r'`swe-workbench:([\w-]+)`')
+    for skill_md in sorted(skills_dir.glob("*/SKILL.md")):
+        if skills_cache is not None and skill_md in skills_cache:
+            text = skills_cache[skill_md]
+            if text is None:
+                fail(skill_md.relative_to(ROOT), "could not read file")
+                continue
+        else:
+            text = skill_md.read_text(encoding="utf-8")
```
lugassawan: **Medium — fallback `read_text` has no `OSError` guard**

The `else` branch (cache miss) calls `skill_md.read_text()` without a try/except, while the `if` branch already has a guarded path. In the unlikely event of a race (file disappears between glob and read) this would crash the validator with an unhandled exception rather than recording a tidy `fail()`.

Suggest wrapping to match the pattern already used in `check_skills`:
```python
else:
    try:
        text = skill_md.read_text(encoding="utf-8")
    except OSError:
        fail(skill_md.relative_to(ROOT), "could not read file")
        continue
```

scripts/validate.py
```diff
@@ -271,6 +271,87 @@ def check_command_skill_refs():
...
+            "missing — cannot check workflow-development activation contract",
+        )
+        return
+    fm = parse_frontmatter(skill_md)
+    if fm is None or "description" not in fm:
+        fail(skill_md.relative_to(ROOT), "missing or malformed frontmatter")
+        return
+
+    commands_dir = ROOT / "commands"
+    declared_pattern = re.compile(r'/swe-workbench:([\w-]+)')
```
lugassawan, line 327: **Medium — `declared_pattern` matches anywhere in description, not just the activation clause**

`r'/swe-workbench:([\w-]+)'` will match any occurrence of `/swe-workbench:foo` in the full description string — including "Skip for..." or "can also be invoked for..." prose. The current description is safe, but a future edit adding a negative example ("do not activate for /swe-workbench:design in Mode C") would produce a spurious false failure.

Consider scoping the extraction to the activation clause only:
```python
activation_clause = desc.split(" when ")[0]
raw_declared = set(declared_pattern.findall(activation_clause))
```

scripts/validate.py
```diff
@@ -271,6 +271,87 @@ def check_command_skill_refs():
...
+            skill_md.relative_to(ROOT),
+            f"workflow-development description lists unknown commands: {unknown_commands}",
+        )
+    # Exclude unknown tokens from set-equality to avoid double-reporting
+    declared = raw_declared - set(unknown_commands)
+
+    actual = set()
+    for cmd_md in sorted(commands_dir.glob("*.md")):
+        # commands are intentionally not cached in _build_cache; read directly
+        if "swe-workbench:workflow-development" in cmd_md.read_text(encoding="utf-8"):
```
lugassawan: **Medium — activation detection uses bare substring match, not backtick-wrapped form**

`"swe-workbench:workflow-development" in text` matches any occurrence — including comments, strikethrough Markdown, or prose examples. `check_command_skill_refs` (the sibling check) uses the stricter backtick form. Asymmetry here could produce false activators from non-invocation prose.

Suggest using the same regex as the sibling check:
```python
wf_pattern = re.compile(r'`swe-workbench:workflow-development`')
for cmd_md in sorted(commands_dir.glob("*.md")):
    if wf_pattern.search(cmd_md.read_text(encoding="utf-8")):
        actual.add(cmd_md.stem)
```

scripts/validate.py
lugassawan: Addressed in c8f0bc6ff81c5a208a0eabe1034ff1655209397b: wrapped the else-branch read_text in try/except OSError, matching the pattern in check_skills.

scripts/validate.py
lugassawan: Addressed in c8f0bc6ff81c5a208a0eabe1034ff1655209397b: scoped the extraction to `desc.split(" when ")[0]` so only the activation clause is matched.

scripts/validate.py
lugassawan: Addressed in c8f0bc6ff81c5a208a0eabe1034ff1655209397b: replaced bare substring match with `re.compile(r'`swe-workbench:workflow-development`')`, consistent with check_command_skill_refs.

scripts/validate.py
```diff
@@ -271,6 +271,94 @@ def check_command_skill_refs():
...
+            f"workflow-development description lists unknown commands: {unknown_commands}",
+        )
+    # Exclude unknown tokens from set-equality to avoid double-reporting
+    declared = raw_declared - set(unknown_commands)
+
+    actual = set()
+    wf_pattern = re.compile(r'`swe-workbench:workflow-development`')
+    for cmd_md in sorted(commands_dir.glob("*.md")):
+        # commands are intentionally not cached in _build_cache; read directly
+        if wf_pattern.search(cmd_md.read_text(encoding="utf-8")):
```
lugassawan: **Medium — `cmd_md.read_text()` in `check_workflow_development_activation_contract` is unguarded against `OSError`**

The `OSError` guard added to `check_skill_skill_refs`'s else branch isn't mirrored in the sibling command-file loop. A permission error or dangling symlink here aborts the whole validation run instead of recording a tidy `fail()`.

Suggest wrapping consistently:
```python
for cmd_md in sorted(commands_dir.glob("*.md")):
    try:
        text = cmd_md.read_text(encoding="utf-8")
    except OSError:
        fail(cmd_md.relative_to(ROOT), "could not read file")
        continue
    if wf_pattern.search(text):
        actual.add(cmd_md.stem)
```

lugassawan: Addressed in fef9a7942cbe3a273cc97ff56a23d6d7c346f8e5: wrapped cmd_md.read_text() in try/except OSError, matching the guard added to check_skill_skill_refs in the previous pass.

PR #421, opened 2026-06-04
[feat] Auto-redact confidential identifiers in /report-issue before filing

### LINKED ISSUE #415: [feat] `report-issue`: auto-redact or warn on confidential identifiers before filing
## Goal

Automatically redact or warn users about confidential identifiers (internal company names, repo names, service endpoints, email addresses) in the drafted issue body before the preview gate, so sensitive information is never accidentally exposed to the public GitHub repo.

## Motivation

When a user files a GitHub issue via `/swe-workbench:report-issue` during development, the drafted body may contain internal identifiers gathered from conversation context — company names, monorepo names, internal service identifiers, email addresses, internal domain names. The skill has no redaction or warning step, so users must manually review and edit the preview to strip sensitive data. This creates two problems:
1. **User friction**: Extra manual step before filing, slowing down issue capture.
2. **Privacy risk**: If a user forgets to edit, confidential information leaks to the public repo and cannot be easily retracted.

## Surface

- [x] Skill — `integration-*` (report-issue is an integration skill)
- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch

**Option A — Automatic redaction (recommended):**
- Before step 8 (preview gate), scan the drafted body for internal identifiers using a configurable pattern list (regex or keyword lists).
- Common patterns: company names, internal repo slugs, internal service names (`*-api`, `*-service`), internal email domains, internal hostnames/domains.
- Replace detected identifiers with generic placeholders: "a downstream consumer", "the monorepo", "an internal service", "[internal-email-domain]", etc.
- Preserve all non-sensitive text and structure.
- Add a brief note in the preview: "_Automatically redacted X internal identifier(s)._"

**Option B — Warning + user confirmation:**
- Before step 8, scan for identifiers and list any found.
- Show a warning in the preview: "_⚠️ X potential internal identifier(s) detected. Review before filing: [list]._"
- Ask user to confirm: "Continue with these identifiers?" or "Edit body to remove/replace?".

**Implementation notes:**
- Pattern list should be maintainable (separate config or hardcoded list of common internal naming patterns).
- Should handle variations: camelCase, snake_case, kebab-case.
- Should be conservative (prefer false positives to false negatives, so users can edit if needed).
- Should not redact user-intentional disclosures (e.g., if the user explicitly decides a service name is OK for public discussion).

---

_Reported via `/swe-workbench:report-issue` — plugin v0.1.16, Claude Code 2.1.162._

lugassawan:
## Summary

- Add a **Redaction pass (privacy guard)** sub-step to step 7 of `commands/report-issue.md`: before writing the draft body to the temp `.md` file, the model scans it for confidential identifiers from conversation context and replaces each with a generic placeholder.
- Allowlist keeps plugin-own identifiers (`swe-workbench`, `lugassawan`, command/skill/agent names, plugin paths) and a closed list of public tech names (GitHub, Claude Code, `gh`, Python, pytest, Node, npm, pip) intact.
- Redaction catalog covers: email → `[internal-email]`; hostnames/URLs → `[internal-host]`; IPs → `[internal-ip]`; API keys/tokens → `[redacted-token]`; service names → `an internal service`; org/repo slugs → prose placeholders.
- Add `Redacted: <N identifier(s) → placeholders | none detected>` line to the step 8 preview block; when N > 0, also render a preview-only italic note above the `Body:` fence (note is never written to the `.md` sidecar).
- Cover with 5 new deliberate-break-proof assertions in `tests/test_report_issue_command.py`.

## Test Plan

- [x] `pytest tests/test_report_issue_command.py` — all 11 assertions pass (6 pre-existing + 5 new)
- [x] `pytest tests/` — full suite of 1158 tests green
- [x] `python scripts/validate.py` — frontmatter, layout, and no-cycle checks pass
- [x] Deliberate-break proof: each new assertion fails when its corresponding instruction phrase is removed from the command file
- [x] Pre-push CI passed (all 1158 tests)

- [ ] Invoke `/swe-workbench:report-issue` with a draft that embeds a fake email (`user@corp.internal`) and a service name (`payments-service`): confirm preview shows both replaced with placeholders and `Redacted: 2 internal identifier(s) → placeholders`
- [ ] Confirm that `swe-workbench`, `lugassawan`, `report-issue` in the same draft are NOT redacted (allowlist intact)
- [ ] Confirm the `_Automatically redacted N identifier(s)._` note does not appear in the filed body (abort at preview — do not file)
- [ ] Invoke with a clean draft (no identifiers): confirm `Redacted: none detected`

Closes #415

### DIFF
--- a/commands/report-issue.md
+++ b/commands/report-issue.md
@@ -76,12 +76,12 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
 
    **Redaction pass (privacy guard).** Before writing the body in step 8, scan the drafted body for confidential identifiers pulled from conversation context and replace each with a generic placeholder. Prefer over-redaction — the user restores false positives at the preview gate.
 
-   - **Allowlist — NEVER redact** (these are the legitimate subject of the issue): the target repo `lugassawan/swe-workbench` and the bare string `swe-workbench`; this plugin's command/skill/agent names (e.g. `report-issue`, `capture`, `product-manager`, `workflow-*`, `principle-*`, `language-*`); plugin-internal file paths (e.g. `commands/report-issue.md`); the repo owner handle `lugassawan`; and the following public tech names exactly: GitHub, Claude Code, the `gh` CLI, Python, pytest, Node, npm, pip. When uncertain whether a name is public, prefer to redact it.
+   - **Allowlist — NEVER redact** (these are the legitimate subject of the issue): the target repo `lugassawan/swe-workbench` and the bare string `swe-workbench`; this plugin's command/skill/agent names (e.g. `report-issue`, `capture`, `product-manager`, `workflow-*`, `principle-*`, `language-*`); plugin-internal file paths (e.g. `commands/report-issue.md`); the repo owner handle `lugassawan`; and the following public tech names and their canonical domains: GitHub (`github.com`, `api.github.com`), Claude Code, the `gh` CLI, Python, pytest, Node, npm, pip. When uncertain whether a name is public, prefer to redact it.
    - **Redact when NOT allowlisted** (handle camelCase / snake_case / kebab-case variants):
      - Email addresses → `[internal-email]`
      - URLs, hostnames, internal domains (e.g. `*.corp`, `*.internal`, company domains) → `[internal-host]`
      - IP addresses → `[internal-ip]`
-     - API keys, tokens, and credential strings matching common patterns (e.g. `sk-…`, `ghp_…`, `AKIA…`, `Bearer …`, UUID-format secrets) → `[redacted-token]`
+     - API keys, tokens, and credential strings matching common patterns (e.g. `sk-…`, `ghp_…`, `AKIA…`, `Bearer <token>`) → `[redacted-token]`. Do NOT redact bare UUIDs without surrounding context that identifies them as secrets (e.g., a UUID in a JSON field named `secret`, `key`, or `token`).
      - Internal service identifiers (`*-api`, `*-service`, `*-svc`, `*-worker`, `*-gateway`) that are not plugin names → `an internal service`
      - Apparent company / product / org names, and other-repo or monorepo slugs surfaced from conversation → `a downstream consumer` / `the monorepo` / `an internal repository` as fits.
    - Never redact text the user explicitly typed as their own thought intending it to be public.
@@ -103,6 +103,6 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
 
    Reply 'confirm' to file, or edit any of the above (including the label) and I'll redraft.
    ```
-   When a label was matched, render `Command:` with `--label "<chosen-label>"`. When no label was matched, omit the `--label` segment entirely. When N > 0 from the redaction pass, also render the line `_Automatically redacted N internal identifier(s)._` as a plain Markdown italic line **outside and immediately above the opening triple-backtick of the `Body:` fence** — this note is **preview-only** and must NOT be written into the `.md` body file; write the `.md` file first, then append the note only to the printed preview output. **Wait for the user to reply `confirm`. Do NOT run `gh issue create` on this turn.**
+   When a label was matched, render `Command:` with `--label "<chosen-label>"`. When no label was matched, omit the `--label` segment entirely. **Wait for the user to reply `confirm`. Do NOT run `gh issue create` on this turn.**
 
 9. **File on confirm.** Only when the user replies `confirm`, read the command from the `.cmd` sidecar written in step 8 and run it exactly as written — do not regenerate the title or path. Return the issue URL.
--- a/commands/report-issue.md
+++ b/commands/report-issue.md
@@ -74,12 +74,25 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
    - **Plugin version:** list `~/.claude/plugins/cache/swe-workbench/swe-workbench/` version directories, sort semantically (`sort -V`), take the highest with `tail -1`, then read `plugin.json` from it: `ls ~/.claude/plugins/cache/swe-workbench/swe-workbench/ 2>/dev/null | sort -V | tail -1 | xargs -I{} python3 -c "import json; print(json.load(open('$HOME/.claude/plugins/cache/swe-workbench/swe-workbench/{}/.claude-plugin/plugin.json'))['version'])"`. If this returns no output (not installed from cache), fall back to: `gh api repos/lugassawan/swe-workbench/contents/.claude-plugin/plugin.json --repo lugassawan/swe-workbench --jq '.content' | python3 -c "import base64,sys,json; print(json.loads(base64.b64decode(sys.stdin.read().strip()))['version'])"`.  
    - **CLI version:** `claude --version` — strip the leading `Claude Code ` prefix to get the bare semver.
 
+   **Redaction pass (privacy guard).** Before writing the body in step 8, scan the drafted body for confidential identifiers pulled from conversation context and replace each with a generic placeholder. Prefer over-redaction — the user restores false positives at the preview gate.
+
+   - **Allowlist — NEVER redact** (these are the legitimate subject of the issue): the target repo `lugassawan/swe-workbench` and the bare string `swe-workbench`; this plugin's command/skill/agent names (e.g. `report-issue`, `capture`, `product-manager`, `workflow-*`, `principle-*`, `language-*`); plugin-internal file paths (e.g. `commands/report-issue.md`); the repo owner handle `lugassawan`; and common public tech names (GitHub, Claude Code, gh, Python, pytest, etc.).
+   - **Redact when NOT allowlisted** (handle camelCase / snake_case / kebab-case variants):
+     - Email addresses → `[internal-email]`
+     - URLs, hostnames, internal domains (e.g. `*.corp`, `*.internal`, company domains) → `[internal-host]`
+     - IP addresses → `[internal-ip]`
+     - Internal service identifiers (`*-api`, `*-service`, `*-svc`, `*-worker`, `*-gateway`) that are not plugin names → `an internal service`
+     - Apparent company / product / org names, and other-repo or monorepo slugs surfaced from conversation → `a downstream consumer` / `the monorepo` / `an internal repository` as fits.
+   - Never redact text the user explicitly typed as their own thought intending it to be public.
+   - Count the replacements made; carry the count into the step 8 preview.
+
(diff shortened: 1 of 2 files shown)


commands/report-issue.md
```diff
@@ -74,12 +74,26 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
...
    - **CLI version:** `claude --version` — strip the leading `Claude Code ` prefix to get the bare semver.
 
+   **Redaction pass (privacy guard).** Before writing the body in step 8, scan the drafted body for confidential identifiers pulled from conversation context and replace each with a generic placeholder. Prefer over-redaction — the user restores false positives at the preview gate.
+
+   - **Allowlist — NEVER redact** (these are the legitimate subject of the issue): the target repo `lugassawan/swe-workbench` and the bare string `swe-workbench`; this plugin's command/skill/agent names (e.g. `report-issue`, `capture`, `product-manager`, `workflow-*`, `principle-*`, `language-*`); plugin-internal file paths (e.g. `commands/report-issue.md`); the repo owner handle `lugassawan`; and the following public tech names exactly: GitHub, Claude Code, the `gh` CLI, Python, pytest, Node, npm, pip. When uncertain whether a name is public, prefer to redact it.
```
lugassawan: **Medium — allowlist gap: `github.com` URLs will be redacted as `[internal-host]`**

The allowlist names the product "GitHub" but not its domain. A body that contains `https://github.com/lugassawan/swe-workbench/issues/123` (a perfectly legitimate cross-reference) will have the host replaced with `[internal-host]`, silently corrupting the link.

Suggested fix — add the domain alongside the product name:
```
…the following public tech names and their canonical domains: GitHub (`github.com`, `api.github.com`), Claude Code, the `gh` CLI…
```

commands/report-issue.md
```diff
@@ -74,12 +74,26 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
...
+   - **Redact when NOT allowlisted** (handle camelCase / snake_case / kebab-case variants):
+     - Email addresses → `[internal-email]`
+     - URLs, hostnames, internal domains (e.g. `*.corp`, `*.internal`, company domains) → `[internal-host]`
+     - IP addresses → `[internal-ip]`
+     - API keys, tokens, and credential strings matching common patterns (e.g. `sk-…`, `ghp_…`, `AKIA…`, `Bearer …`, UUID-format secrets) → `[redacted-token]`
```
lugassawan: **Medium — "UUID-format secrets" will match all UUIDs, not just secrets**

UUIDs appear legitimately as GitHub run IDs, trace IDs, correlation IDs, and issue references. An LLM following this rule literally will redact any UUID it encounters, producing noisy false positives that obscure useful diagnostic content.

Suggested fix — narrow to prefix-anchored patterns only:
```
- API keys, tokens, and credential strings matching common patterns (e.g. `sk-…`, `ghp_…`, `AKIA…`, `Bearer <token>`) → `[redacted-token]`. Do NOT redact bare UUIDs without surrounding context that identifies them as secrets (e.g., a UUID in a field named `secret`, `key`, or `token`).
```

commands/report-issue.md
```diff
@@ -74,12 +74,26 @@ Delegate to the `product-manager` subagent. Its response must deliver all of the
...
+   - Never redact text the user explicitly typed as their own thought intending it to be public.
+   - Count the replacements made; carry the count into the step 8 preview.
+
 8. **Preview gate.** Obtain a Unix timestamp once (`date +%s`) and reuse it. Write the body to `/tmp/report-issue-lugassawan-swe-workbench-<unix-timestamp>.md` using the `Write` tool (not a Bash heredoc). Also write a one-line command to `/tmp/report-issue-lugassawan-swe-workbench-<unix-timestamp>.cmd` using the `Write` tool: when a label was matched, write `gh issue create --repo lugassawan/swe-workbench --title "..." --body-file <path> --label "<matched-label>"`; when no label was matched, write `gh issue create --repo lugassawan/swe-workbench --title "..." --body-file <path>` (no `--label` segment). Then print:
    ```
    Filing into: lugassawan/swe-workbench
    Template: <chosen template> | none — default body
    Title: <drafted title>
    Label: <chosen label> | none — no matching label found
+   Redacted: <N internal identifier(s) → placeholders | none detected>
```
lugassawan, line 96: **Medium — redundant redaction count: two different representations shown simultaneously**

When N > 0 the preview will show:
1. `Redacted: 2 internal identifier(s) → placeholders` (line 96, inside the summary block)
2. `_Automatically redacted 2 internal identifier(s)._` (line 106 prose, italic line above `Body:` fence)

Both convey the same count in different formats. A user reading the preview sees the count twice, and a model following the instructions is unsure which representation is authoritative. The `N = 0` case is also asymmetric: the summary line says "none detected" but the italic note simply doesn't appear — which is inconsistent.

Suggested fix — pick one. The summary-line approach (`Redacted: <N …>`) is more consistent with the existing `Label:`/`Template:` pattern and handles zero cleanly without a conditional. Drop the prose italic note instruction from line 106, or keep only the italic note and remove line 96.

tests/test_report_issue_command.py
```diff
@@ -79,3 +79,73 @@ def test_report_issue_supports_blank_argument():
...
+    text = REPORT_ISSUE_MD.read_text()
+    assert "Redacted:" in text, (
+        "commands/report-issue.md must include a 'Redacted:' line in the preview gate block"
+    )
+
+
+def test_report_issue_redaction_before_preview():
+    """commands/report-issue.md: redaction pass → Redacted: line → confirm gate, in that order."""
+    text = REPORT_ISSUE_MD.read_text()
+    redaction_pos = text.find("Redaction pass")
```
lugassawan, line 140: **Low — ordering test anchors could silently pass against the wrong position**

`.find()` always returns the first occurrence. If `"Redacted:"` or `"Reply 'confirm'"` ever appears earlier in the file (e.g. in a step-0 example or a new confirmation flow), the ordering assertion would pass against the wrong position while the actual elements are out of order.

Suggested fix — chain the searches so each starts after the previous match:
```python
redaction_pos = text.find("Redaction pass")
redacted_line_pos = text.find("Redacted:", redaction_pos)   # start after redaction_pos
confirm_pos = text.find("Reply 'confirm'", redacted_line_pos)  # start after redacted_line_pos
```
This makes the test self-documenting (it reads as "Redacted: appears after Redaction pass") and immune to earlier occurrences.

tests/test_report_issue_command.py
```diff
@@ -79,3 +79,73 @@ def test_report_issue_supports_blank_argument():
...
+def test_report_issue_redaction_has_allowlist():
+    """commands/report-issue.md must define an allowlist for the redaction pass."""
+    text = REPORT_ISSUE_MD.read_text()
+    assert "Allowlist" in text, (
+        "commands/report-issue.md must include an 'Allowlist' section in the redaction instructions"
+    )
+    assert "NEVER redact" in text, (
+        "commands/report-issue.md must include a 'NEVER redact' directive in the allowlist"
+    )
+    assert "swe-workbench" in text[text.find("NEVER redact"):text.find("NEVER redact") + 400], (
```
lugassawan: **Low — hardcoded 400-char window in allowlist token assertion**

`text[text.find("NEVER redact"):text.find("NEVER redact") + 400]` uses an arbitrary character count that silently breaks if the allowlist prose is reorganised or expanded beyond 400 characters.

Suggested fix — use the allowlist block boundaries instead:
```python
never_redact_pos = text.find("NEVER redact")
allowlist_end = text.find("Redact when NOT allowlisted", never_redact_pos)
allowlist_block = text[never_redact_pos:allowlist_end]
assert "swe-workbench" in allowlist_block, (
    "commands/report-issue.md must name 'swe-workbench' as an allowlisted token"
)
```
This makes the window self-describing and resilient to prose changes.

commands/report-issue.md
lugassawan: Addressed in 0d1294c98f267930e467bb6a4475ff573ad1e1a2: added `github.com` and `api.github.com` to the allowlist domain list alongside the product name.

commands/report-issue.md
lugassawan: Addressed in 0d1294c98f267930e467bb6a4475ff573ad1e1a2: narrowed to prefix-anchored patterns only; added explicit 'Do NOT redact bare UUIDs' guard.

commands/report-issue.md
lugassawan: Addressed in 0d1294c98f267930e467bb6a4475ff573ad1e1a2: removed the redundant italic prose note; the `Redacted:` summary line in the preview block is sufficient and consistent with the existing `Label:`/`Template:` pattern.

tests/test_report_issue_command.py
lugassawan: Addressed in 0d1294c98f267930e467bb6a4475ff573ad1e1a2: chained the `.find()` calls so each search starts from the previous match position, making the ordering check immune to earlier occurrences.

tests/test_report_issue_command.py
lugassawan: Addressed in 0d1294c98f267930e467bb6a4475ff573ad1e1a2: replaced the hardcoded 400-char window with block-boundary search using `text.find("Redact when NOT allowlisted", never_redact_pos)` as the end anchor.

tests/test_report_issue_command.py
```diff
@@ -79,3 +79,76 @@ def test_report_issue_supports_blank_argument():
...
+        "commands/report-issue.md must include a 'Redacted:' line in the preview gate block"
+    )
+
+
+def test_report_issue_redaction_before_preview():
+    """commands/report-issue.md: redaction pass → Redacted: line → confirm gate, in that order."""
+    text = REPORT_ISSUE_MD.read_text()
+    redaction_pos = text.find("Redaction pass")
+    redacted_line_pos = text.find("Redacted:", redaction_pos)
+    confirm_pos = text.find("Reply 'confirm'", redacted_line_pos)
```
lugassawan: **Low — chained `.find()` uses `redacted_line_pos` before its `-1` guard is checked**

`text.find("Reply 'confirm'", redacted_line_pos)` passes `redacted_line_pos` as the `start` argument before the `assert redacted_line_pos != -1` guard runs. In Python, `str.find(sub, -1)` starts from the last character (not position 0), making `confirm_pos` semantically unreliable if `"Redacted:"` is ever absent. The subsequent `-1` guards still catch the failure, so this doesn't affect correctness today — but it's a latent confusion.

Suggested fix — interleave guards with searches so each chained call uses a validated position:
```python
redaction_pos = text.find("Redaction pass")
assert redaction_pos != -1, "..."
redacted_line_pos = text.find("Redacted:", redaction_pos)
assert redacted_line_pos != -1, "..."
confirm_pos = text.find("Reply 'confirm'", redacted_line_pos)
assert confirm_pos != -1, "..."
assert redaction_pos < redacted_line_pos < confirm_pos, "..."
```
Non-blocking — does not affect correctness on the current file.

lugassawan: Addressed in 397c86f1dbcac038434e9100cf1d00a23ad8b43e: interleaved the `assert != -1` guards with the chained `.find()` calls so each search operates on a validated position.

PR #426, opened 2026-06-05
[fix] /implement embeds ## Workflow section under plan mode

### LINKED ISSUE #423: [bug] /implement drops mandated ## Workflow section when run inside plan mode
## Product framing

**Problem.** When a user runs `/swe-workbench:implement` while Claude Code's built-in plan mode is active, the command's mandated `## Workflow` section is silently dropped from the resulting plan file — the user gets a plan that looks complete but is missing the project-detected branch/verify/review/PR workflow block they were promised.

**Value.** Anyone running `/implement` inside plan mode (the common path for code-modifying tasks) benefits: the plan they execute against will actually contain the workflow guidance the command guarantees, instead of a silently incomplete one.

**Acceptance criteria.**
- Given `/swe-workbench:implement <issue-url>` is invoked while plan mode is active, when the plan file is written, then it contains a `## Workflow` section per the plan-workflow-section template.
- Given Mode A always applies to `/implement`, when the command finalizes the plan, then project-detection placeholders are substituted (no raw placeholders).
- Given plan mode's own workflow takes precedence, the command either reconciles with it or surfaces a clear failure rather than silently omitting the section.

**Impact / Effort.** Impact: M — affects every `/implement` run under plan mode. Effort: M — reconcile Mode A activation with plan-mode precedence + a verification check.

---

## Surface affected
- [x] Skill — `workflow-*`
- [x] Slash command

**Specific surface**: `/swe-workbench:implement` + `swe-workbench:workflow-development` (Mode A)

## Claude Code version
2.1.163

## Plugin version
0.1.17

## Reproduction steps
1. Invoke `/swe-workbench:implement <github issue URL>` while plan mode is active.
2. Let the session proceed: grill mode runs, then plan mode's own workflow (Explore → Plan → write plan → ExitPlanMode) takes over.
3. Open the produced plan file at `~/.claude/plans/...`.

## Expected behavior
Mode A always applies for `/implement`; the plan file should contain a `## Workflow` section with project-detection placeholders substituted.

## Actual behavior
`workflow-development` Mode A was never activated. Plan mode's workflow superseded the command's finalization step, so the plan file has NO `## Workflow` section. The omission is silent.

## Diagnosis / root cause (from the session that hit this)
The failure is an **instruction-precedence collision**, not a logic bug in the command:

1. **Two workflows were active at once.** `/swe-workbench:implement` was invoked while Claude Code's built-in **plan mode** was already on. Plan mode injects its own mandatory plan workflow (Phase 1 Explore → Phase 2 Design → Phase 3 Review → Phase 4 Final Plan → Phase 5 `ExitPlanMode`) and explicitly states: *"This supersedes any other instructions you have received."*

2. **Plan mode wins the precedence contest.** The command tells the agent to "activate `workflow-development` in Mode A and embed the rendered `## Workflow` section … before finalizing the plan." But plan mode's superseding clause + the `using-superpowers` precedence ladder push the agent onto the plan-mode track, where the command's Mode-A step never fires.

3. **The embed hook is structurally bypassed.** The command's Mode A is gated on *"if no written plan exists, draft one via `superpowers:writing-plans`"* and then embed the Workflow section into **that** plan. Under plan mode the plan is instead written to the harness's single designated plan file via a *different* workflow with a *different* prescribed structure (Context / approach / critical files / verification — no Workflow section). So the "embed into the writing-plans output" hook has nothing to attach to.

4. **Edit restriction reinforces it.** Plan mode forbids editing anything except the one designated plan file, so even an attempt to render the Workflow section into a separate `writing-plans` artifact would be blocked.

5. **The drop is silent.** Nothing reconciles the two workflows and no gate verifies the `## Workflow` section exists before `ExitPlanMode`, so the omission passes unnoticed.

**Suggested fix direction:** have `/implement` detect active plan mode and either (a) fold the Mode-A `## Workflow` section into the plan-mode Final-Plan structure (write it into the designated plan file before `ExitPlanMode`), or (b) add a pre-`ExitPlanMode` gate that asserts the `## Workflow` section is present and fails loudly if not.

## Environment (optional)
Claude Code plan mode active during the run.

_Reported via /swe-workbench:report-issue — plugin v0.1.17, Claude Code 2.1.163._

lugassawan:
## Summary
- Fix `/implement` silently dropping the mandated `## Workflow` section under Claude Code's built-in plan mode
- Add `ExitPlanMode` robustness clause to `commands/implement.md` Phase 2, bringing it in line with its 7 sibling commands (`document`, `debug`, `architect`, `migrate`, `refactor`, `design`, `test`)
- Add `check_plan_mode_workflow_embedding()` to `scripts/validate.py` to contract-gate this class of bug across all current and future Mode-A activators

## Test Plan
- [ ] `bash scripts/validate.sh` → 0 failures (all 8 Mode-A activators now carry the clause)
- [ ] `python3 -m pytest tests/test_validate.py::TestCheckPlanModeWorkflowEmbedding` → 4 pass (3 synthetic + 1 live-tree)
- [ ] `python3 -m pytest` → 1162 pass

- [ ] Reproduce: confirm that without the clause, `validate.sh` flags `implement.md` (the TDD RED step proved this)
- [ ] Verify: confirm the fixed `implement.md` passes the new gate clean

Closes #423


scripts/validate.py
```diff
@@ -805,6 +805,32 @@ def dfs(node):
...
+    wf_ref = re.compile(r'`swe-workbench:workflow-development`')
+    mode_a = re.compile(r'\bMode A\b')
+    clause = "ExitPlanMode"
+    for cmd_md in sorted(commands_dir.glob("*.md")):
+        try:
+            text = cmd_md.read_text(encoding="utf-8")
+        except OSError:
+            fail(cmd_md.relative_to(ROOT), "could not read file")
+            continue
+        if wf_ref.search(text) and mode_a.search(text) and clause not in text:
```
lugassawan, line 831: **Medium** — Gate condition fires on `wf_ref AND mode_a AND NOT clause`. A future command that mentions "Skip Mode A if the plan is pure analysis" (without activating it) would false-positive here if `ExitPlanMode` is also absent. No current false-positive exists (live-tree scan is clean), but documenting the limitation in the docstring reduces future maintenance friction.

```suggestion
def check_plan_mode_workflow_embedding():
    """Every command that activates workflow-development Mode A must instruct embedding the
    ## Workflow section covering the ExitPlanMode path — otherwise the section is silently
    dropped under built-in plan mode (#423).

    Note: gate fires when all three signals appear anywhere in the document. A command that
    mentions 'Mode A' only in a 'Skip Mode A if …' clause without activating it will produce
    a false positive if ExitPlanMode is also absent. Today no such command exists; add the
    clause if one is introduced.
    """
```

tests/test_validate.py
```diff
@@ -1973,3 +1973,60 @@ def test_onboarding_docs_point_at_real_slice(self):
...
+            "---\ndescription: d\n---\n\n"
+            "Activate `swe-workbench:workflow-development` in **Mode A** "
+            "whether saved to a plan file or passed to `ExitPlanMode`.\n",
+            encoding="utf-8",
+        )
+        validate.check_plan_mode_workflow_embedding()
+        assert len(validate.FAILURES) == 0
+
+    def test_no_mode_a_no_failure(self, reset_validate):
+        """A command referencing workflow-development without Mode A is not flagged."""
```
lugassawan: **Low** — Test name implies Mode-B commands are exempt, but the gate only checks whether `\bMode A\b` matches (not whether Mode B is present). The test is logically correct but the name is misleading.

```suggestion
    def test_wf_ref_without_mode_a_no_failure(self, reset_validate):
        """A command that references workflow-development but contains no 'Mode A' token is not flagged."""
```

commands/implement.md
```diff
@@ -23,7 +23,7 @@ Isolate this work in a worktree when the scope warrants it (non-trivial changes,
...
+- If no written plan exists, draft one via `superpowers:writing-plans`. **Whenever you author or finalize a plan that modifies the codebase — whether saved to a plan file or passed to `ExitPlanMode`** — first activate `swe-workbench:workflow-development` in **Mode A** and embed the rendered `## Workflow` section per `skills/workflow-development/templates/plan-workflow-section.md` **before the plan is finalized (i.e. before `ExitPlanMode`)**. Run the skill's project-detection (`git branch -a`, `git log --oneline -20`, Makefile grep, PR-template lookup) so placeholders are substituted from this repo, not left as `[[detect:…]]`. Since `/implement` always modifies the codebase, Mode A always applies here — **if the plan being finalized lacks a `## Workflow` section, that is a defect: add it before finalizing rather than emitting an incomplete plan.**
```
lugassawan: **Low (readability)** — The new sentence appends the loud-failure assertion inline, making the bullet harder to scan. The 7 sibling commands separate the main instruction from the defect guard. Consider breaking the guard into a sub-bullet:

```suggestion
- If no written plan exists, draft one via `superpowers:writing-plans`. **Whenever you author or finalize a plan that modifies the codebase — whether saved to a plan file or passed to `ExitPlanMode`** — first activate `swe-workbench:workflow-development` in **Mode A** and embed the rendered `## Workflow` section per `skills/workflow-development/templates/plan-workflow-section.md` **before the plan is finalized (i.e. before `ExitPlanMode`)**. Run the skill's project-detection (`git branch -a`, `git log --oneline -20`, Makefile grep, PR-template lookup) so placeholders are substituted from this repo, not left as `[[detect:…]]`. Since `/implement` always modifies the codebase, Mode A always applies here.
  - **Defect guard:** if the plan being finalized lacks a `## Workflow` section, add it before emitting the plan rather than passing an incomplete plan to `ExitPlanMode`.
```

scripts/validate.py
lugassawan: Addressed in b65ae4060bbe92c8f5228c63fa34764f0347878d: extended docstring with false-positive caveat for 'Skip Mode A if …' phrasing.

tests/test_validate.py
lugassawan: Addressed in b65ae4060bbe92c8f5228c63fa34764f0347878d: renamed to `test_wf_ref_without_mode_a_no_failure` and updated docstring to describe what the gate actually tests.

commands/implement.md
lugassawan: Addressed in b65ae4060bbe92c8f5228c63fa34764f0347878d: split the defect guard into a sub-bullet `- **Defect guard:** …` for improved scannability.

PR #432, opened 2026-06-10
[refactor] Move plugin-runtime scripts to runtime/ + 5 skill refinements

lugassawan:
## Summary
- Move `clean-ephemeral.sh`, `clean-state-files.sh`, and `doctor.sh` from `scripts/` to new `runtime/` dir; update all skill/command/agent/test references
- Add `runtime/fetch-pr.sh` (collapses 3 duplicated `gh pr view` fetch trios) and `runtime/reply-and-resolve.sh` (encapsulates Phase-5 reply+resolve for address-feedback)
- Remove stray `---` from review-summary `printf` in pr-review + pr-review-followup; add Gradle marker rows to both workflow-development detection tables; add `rimba version` NOTE to `resolve-rimba.sh` + guarded version probe in workflow-development

## Test Plan
- [x] `bash scripts/validate.sh` → All checks passed
- [x] `python -m pytest tests/` → 1247 passed (including pre-push hook re-run)
- [x] `grep -rn '/scripts/clean-ephemeral.sh|/scripts/clean-state-files.sh|/scripts/doctor.sh' skills commands agents tests` → zero hits
- [x] `bash -n runtime/*.sh` → no syntax errors
- [x] New test file `tests/test_runtime_scripts.py` asserts existence, executable bit, bash syntax, and skill references for all 5 runtime scripts

### Type-tailored checks ([refactor])
- [x] Same outputs: runtime scripts are identical byte-for-byte to the originals (git mv, no content change to the 3 moved scripts)
- [x] All callers updated atomically in the same commit — no partial-migration window
- [x] `validate.py` does not scan `scripts/` (by design); runtime-script references are guarded by pytest substring assertions which went red then green

Issue: N/A — user-requested runtime/dev script split + skill refinements


runtime/fetch-pr.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Fetch a PR's metadata JSON; exit 1 if the PR is missing or unreadable.
+# Usage: fetch-pr.sh <PR> <out_path> <json_fields>
+set -euo pipefail
+PR="${1:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
+OUT="${2:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
+FIELDS="${3:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
+mkdir -p "$(dirname "$OUT")"
+gh pr view "$PR" --json "$FIELDS" > "$OUT"
+[ -s "$OUT" ] || { echo "PR #$PR not found or not accessible."; exit 1; }
```
lugassawan: **Medium + Low** — Two issues on this line:

1. **`[ -s ]` unreachable under `set -euo pipefail`** — when `gh pr view` exits non-zero (PR not found/inaccessible), `set -e` aborts at line 9 before reaching this guard. The check only fires for an empty-but-zero-exit case (rare). The header comment "exit 1 if the PR is missing" misleads readers about what actually gates the error. Consider:
```suggestion
gh pr view "$PR" --json "$FIELDS" > "$OUT" || { echo "PR #$PR not found or not accessible." >&2; exit 1; }
[ -s "$OUT" ] || { echo "PR #$PR returned empty JSON." >&2; exit 1; }
```

2. **Error message goes to stdout** — callers capturing stdout for JSON parsing may swallow this silently. `>&2` needed.

runtime/reply-and-resolve.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Reply to a PR review thread (REST) and optionally resolve it (GraphQL).
+# Usage: reply-and-resolve.sh <owner> <repo> <pr> <comment_databaseid> <thread_id> <reply_body>
+# Empty reply_body skips the reply; empty thread_id skips the resolve
+# (ADDRESSED -> both, CLARIFIED -> reply only, DEFERRED -> neither).
+set -euo pipefail
+OWNER="${1:?}"; REPO="${2:?}"; PR="${3:?}"
+COMMENT_DATABASEID="${4:-}"; THREAD_ID="${5:-}"; REPLY_BODY="${6:-}"
+if [ -n "$REPLY_BODY" ]; then
+  gh api "repos/${OWNER}/${REPO}/pulls/${PR}/comments/${COMMENT_DATABASEID}/replies" -F body="$REPLY_BODY"
```
lugassawan, line 13: **Medium** — When `$REPLY_BODY` is non-empty (ADDRESSED or CLARIFIED) but `$COMMENT_DATABASEID` is accidentally empty, the script calls `gh api "repos/.../pulls/.../comments//replies"` — an empty path segment that GitHub will return a 4xx on. A cheap guard prevents a confusing silent failure:

```suggestion
if [ -n "$REPLY_BODY" ]; then
  [ -n "$COMMENT_DATABASEID" ] || { echo "reply-and-resolve: REPLY_BODY set but COMMENT_DATABASEID is empty" >&2; exit 1; }
  gh api "repos/${OWNER}/${REPO}/pulls/${PR}/comments/${COMMENT_DATABASEID}/replies" -F body="$REPLY_BODY"
fi
```

runtime/reply-and-resolve.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Reply to a PR review thread (REST) and optionally resolve it (GraphQL).
+# Usage: reply-and-resolve.sh <owner> <repo> <pr> <comment_databaseid> <thread_id> <reply_body>
+# Empty reply_body skips the reply; empty thread_id skips the resolve
+# (ADDRESSED -> both, CLARIFIED -> reply only, DEFERRED -> neither).
```
lugassawan, line 5: **Low** — The comment says "DEFERRED -> neither" which is correct, but a reader might infer `$COMMENT_DATABASEID` is also empty for DEFERRED. Since `$COMMENT_DATABASEID` is positional and always passed by callers (only `$REPLY_BODY` and `$THREAD_ID` are empty for DEFERRED), this is subtle. Clarify:

```suggestion
# Empty reply_body skips the reply (COMMENT_DATABASEID is then ignored); empty thread_id skips the resolve
```

tests/test_runtime_scripts.py
```diff
@@ -0,0 +1,72 @@
...
+RUNTIME_SCRIPTS = [
+    "clean-ephemeral.sh",
+    "clean-state-files.sh",
+    "doctor.sh",
+    "fetch-pr.sh",
+    "reply-and-resolve.sh",
+]
+
+
+def test_all_runtime_scripts_exist_and_executable():
```
lugassawan, line 21: **Low** — `RUNTIME_SCRIPTS` is a hardcoded list. New scripts added to `runtime/` will silently skip the existence+executable check until this list is updated. Consider adding a glob-based catch-all to complement the explicit list:

```python
def test_all_runtime_sh_files_are_tracked():
    """Every *.sh in runtime/ must be listed in RUNTIME_SCRIPTS (prevents silent coverage gaps)."""
    on_disk = {p.name for p in RUNTIME.glob("*.sh")}
    missing = on_disk - set(RUNTIME_SCRIPTS)
    assert not missing, f"New runtime scripts not in RUNTIME_SCRIPTS: {missing}"
```

runtime/README.md
```diff
@@ -0,0 +1,33 @@
+# runtime/
+
+Scripts in this directory are **plugin-runtime** artifacts: they ship to end-user machines as part of
+the swe-workbench plugin and are invoked at runtime by skills, commands, and agents via
```
lugassawan, line 4: **Nit** — The opening prose uses "plugin-runtime artifacts" while the table header says "Plugin-runtime scripts". Could simplify both to "runtime scripts" to match the directory name and avoid introducing a term not defined elsewhere.

runtime/fetch-pr.sh
lugassawan: Addressed in 4f9600e794b0faedf377e35d5f7ddca671d4f375: split into two distinct guards (`command-failed` vs `empty-JSON`); both redirect error messages to `>&2` so callers capturing stdout for JSON parsing won't swallow them.

runtime/reply-and-resolve.sh
lugassawan: Addressed in 4f9600e794b0faedf377e35d5f7ddca671d4f375: added `[ -n "$COMMENT_DATABASEID" ]` guard inside the `if [ -n "$REPLY_BODY" ]` block; exits 1 with a descriptive error rather than silently calling `gh api` with an empty path segment.

runtime/reply-and-resolve.sh
lugassawan: Addressed in 4f9600e794b0faedf377e35d5f7ddca671d4f375: comment now reads `Empty reply_body skips the reply (COMMENT_DATABASEID is then ignored)` — makes clear the databaseid is always present positionally and only ignored when reply_body is empty.

tests/test_runtime_scripts.py
lugassawan: Addressed in 4f9600e794b0faedf377e35d5f7ddca671d4f375: added `test_all_runtime_sh_files_are_tracked` which globs `runtime/*.sh` and asserts every result is present in `RUNTIME_SCRIPTS` — any future script added to the directory will fail CI until the list is updated.

runtime/README.md
lugassawan: Addressed in 4f9600e794b0faedf377e35d5f7ddca671d4f375: changed to `runtime scripts` in the opening paragraph — matches the directory name and the table header, no new term introduced.

runtime/reply-and-resolve.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# Reply to a PR review thread (REST) and optionally resolve it (GraphQL).
+# Usage: reply-and-resolve.sh <owner> <repo> <pr> <comment_databaseid> <thread_id> <reply_body>
+# Empty reply_body skips the reply (COMMENT_DATABASEID is then ignored); empty thread_id skips the resolve
+# (ADDRESSED -> both, CLARIFIED -> reply only, DEFERRED -> neither).
+set -euo pipefail
+OWNER="${1:?}"; REPO="${2:?}"; PR="${3:?}"
+COMMENT_DATABASEID="${4:-}"; THREAD_ID="${5:-}"; REPLY_BODY="${6:-}"
```
lugassawan, line 10: **Medium** — DEFERRED threads cause a no-op subprocess launch. When both `REPLY_BODY` and `THREAD_ID` are empty, the script forks a process, evaluates the `${1:?`}/`${2:?`}/`${3:?`} guards (non-fatal only because they're guarded by the `if` branches, not positional-at-start), then exits 0 — zero effect but nonzero cost per DEFERRED thread. More importantly, a caller that accidentally omits OWNER/REPO/PR still hits the `:?` guards even though the call would do nothing. Consider a fast-exit guard before them:

```suggestion
# Fast-exit for DEFERRED: nothing to do
if [ -z "${6:-}" ] && [ -z "${5:-}" ]; then exit 0; fi
OWNER="${1:?}"; REPO="${2:?}"; PR="${3:?}"
```

Alternatively, document in `workflow-address-feedback` Phase 5 to skip the call for DEFERRED threads entirely.

tests/test_runtime_scripts.py
```diff
@@ -0,0 +1,83 @@
...
+    on_disk = {p.name for p in RUNTIME.glob("*.sh")}
+    tracked = set(RUNTIME_SCRIPTS)
+    missing = on_disk - tracked
+    assert not missing, (
+        f"runtime/ contains untracked scripts not in RUNTIME_SCRIPTS: {sorted(missing)}. "
+        "Add them to RUNTIME_SCRIPTS so they get existence, executable, and syntax checks."
+    )
+
+
+def test_runtime_scripts_pass_bash_syntax_check():
```
lugassawan, line 40: **Low** — `bash -n` syntax check is the only behavioral coverage for `fetch-pr.sh` and `reply-and-resolve.sh`. The new scripts carry conditional logic that deserves functional tests (e.g. the `[ -s "$OUT" ]` empty-JSON guard in `fetch-pr.sh`, and the ADDRESSED/CLARIFIED/DEFERRED paths in `reply-and-resolve.sh`). `test_clean_ephemeral.py` and `test_clean_state_files.py` demonstrate the subprocess+PATH-stub pattern. Not blocking — the glob catch-all added above protects against structural gaps — but worth a follow-up `test_fetch_pr_script.py` and `test_reply_and_resolve_script.py`.

runtime/reply-and-resolve.sh
lugassawan: Addressed in 2cb90a9b223a5fb4297086eff4b10a16b7d5b9f5: added fast-exit guard before the :? checks — when both REPLY_BODY and THREAD_ID are empty the script now exits 0 immediately, skipping the fork and preventing confusing :? errors when OWNER/REPO/PR are accidentally omitted on a DEFERRED call. Also tightened the Phase 5 prose in workflow-address-feedback to 'For each ADDRESSED or CLARIFIED thread...' to make the skip explicit.

tests/test_runtime_scripts.py
lugassawan: Addressed in 2cb90a9b223a5fb4297086eff4b10a16b7d5b9f5: added test_fetch_pr_script.py (7 tests: happy path, gh-failure/stderr routing, empty-JSON guard, missing-args) and test_reply_and_resolve_script.py (8 tests: DEFERRED fast-exit, ADDRESSED reply+resolve, CLARIFIED reply-only, missing-COMMENT_DATABASEID guard). Pattern mirrors test_clean_ephemeral.py with a recording gh stub prepended to PATH.

tests/test_fetch_pr_script.py
```diff
@@ -0,0 +1,119 @@
...
+ROOT = Path(__file__).parent.parent
+
+
+def _make_gh_stub(stub_dir: Path, *, exit_code: int, output: str) -> None:
+    """Write a fake gh binary that emits `output` to stdout and exits with `exit_code`."""
+    stub_dir.mkdir(exist_ok=True)
+    output_file = stub_dir / "_gh_output"
+    output_file.write_text(output)
+    stub = stub_dir / "gh"
+    stub.write_text(f"#!/bin/sh\ncat {output_file}\nexit {exit_code}\n")
```
lugassawan: **Medium** — Unquoted path in the `cat` command inside the gh stub: `cat {output_file}`. On any `tmp_path` containing a space (macOS user with space in username, custom `TMPDIR`, CI runner path), the shell word-splits the path and `cat` fails — the happy-path tests then fail with a misleading "Expected exit 0" rather than a clear stub error.

Single-quote the path in the generated script:

```suggestion
    stub.write_text(f"#!/bin/sh\ncat '{output_file}'\nexit {exit_code}\n")
```

tests/test_reply_and_resolve_script.py
```diff
@@ -0,0 +1,149 @@
...
+ROOT = Path(__file__).parent.parent
+
+
+def _make_recording_gh_stub(stub_dir: Path, log_file: Path) -> None:
+    """gh stub that appends one line per invocation to log_file and exits 0."""
+    stub_dir.mkdir(exist_ok=True)
+    stub = stub_dir / "gh"
+    # Record "$1 $2" only: "api <url>" or "api graphql". The GraphQL query arg
+    # spans multiple lines so recording $@ would produce extra log lines.
+    stub.write_text(f'#!/bin/sh\nprintf \'%s %s\\n\' "$1" "$2" >> {log_file}\nexit 0\n')
```
lugassawan: **Medium** — Same pattern: the log-file path is unquoted in the `>>` redirect: `>> {log_file}`. A space in `tmp_path` causes the redirect to fail silently, leaving `log_file` empty — every `_gh_calls()` assertion then returns `[]` regardless of actual behaviour, masking bugs rather than catching them.

Single-quote the path in the generated script:

```suggestion
    stub.write_text(f'#!/bin/sh\nprintf \'%s %s\\n\' "$1" "$2" >> \'{log_file}\'\nexit 0\n')
```

tests/test_fetch_pr_script.py
lugassawan: Addressed in cb8349a6661a13697facb15746d240669146058d: single-quote the embedded path in the cat command so spaces in tmp_path don't cause word-splitting.

tests/test_reply_and_resolve_script.py
lugassawan: Addressed in cb8349a6661a13697facb15746d240669146058d: single-quote the log_file path in the >> redirect so spaces in tmp_path don't cause shell word-splitting.

PR #434, opened 2026-06-10
[feat] Add worked bounded fan-out examples to principle-concurrency (9 languages)

### LINKED ISSUE #377: [feat] Add worked examples to principle-concurrency (9 OO languages)
## Goal

Add an `examples/` dir to `principle-concurrency` with one shared scenario implemented across all 9 OO languages.

## Motivation

Concurrency idioms are among the most language-specific in the OO set — Go's goroutines/channels, Rust's `Arc<Mutex<T>>`/`tokio`, Java's `CompletableFuture`/`ExecutorService`, Kotlin's coroutines, Swift's `async/await`/actors, TypeScript's `Promise.all`. Prose alone can't convey the structural differences. A single worked scenario (e.g. bounded fan-out: fetch N URLs concurrently, collect results) makes the cross-language shape concrete.

The `examples/` convention is documented in `docs/extending.md` and enforced by `scripts/validate.py` (≤120 lines/file, on-demand pointer in SKILL.md, never auto-loaded). `principle-clean-architecture` is currently the only skill using this pattern.

## Surface

- [x] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch

    skills/principle-concurrency/examples/
      <scenario>.cs.md
      <scenario>.go.md
      <scenario>.java.md
      <scenario>.kt.md
      <scenario>.py.md
      <scenario>.rb.md
      <scenario>.rs.md
      <scenario>.swift.md
      <scenario>.ts.md

Pick one shared scenario (e.g. bounded fan-out: fetch many items concurrently with a concurrency limit, collect results) that exercises the language's idiomatic concurrency primitives. Follow the existing convention: "Problem / Implementation / Common Mistake", `# file:` headers, ≤120 lines per file.

Languages (9, OO set): C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript.
Bash and SQL are out of scope.

## Acceptance

- [ ] `principle-concurrency/examples/` dir created with one shared scenario × 9 language files.
- [ ] SKILL.md gains an on-demand pointer line; nothing is auto-loaded.

lugassawan:
## Summary
- Added `skills/principle-concurrency/examples/` with 9 language implementations of the **bounded fan-out** scenario: fetch N item IDs concurrently with cap K=5, collect results in original order.
- Files: `bounded-fan-out.{cs,go,java,kt,py,rb,rs,swift,ts}.md` — each ≤75 lines, matching the `principle-solid/examples/` format convention (em-dash title, three `##` sections, single fenced block, `✗`-annotated Common Mistake).
- Added one read-on-demand blockquote pointer in `skills/principle-concurrency/SKILL.md` after the existing Go-idioms reference; no `@examples/...` auto-include added.

## Test Plan
- [x] `bash scripts/validate.sh` passes clean (incl. `check_examples` ≤120-line cap)
- [x] `pytest` — 1280/1280 pass
- [x] `wc -l skills/principle-concurrency/examples/*.md` — all 9 files ≤120 lines (range 46–75)
- [x] Examples manually reviewed: each uses the idiomatic bounded primitive for its language (errgroup+SetLimit, buffer_unordered, newFixedThreadPool+invokeAll, Semaphore+awaitAll, withThrowingTaskGroup drain-as-you-add, SemaphoreSlim+WhenAll, K-worker cursor pool, asyncio.Semaphore+gather, Queue+K threads)

- [ ] Each example demonstrates a genuine K=5 concurrency cap (not just concurrent execution)
- [ ] Ordering invariant is present in all 9 files (indexed-slot or gather-order guarantee)
- [ ] Common Mistake in each file shows the unbounded anti-pattern with ✗ markers and a `bad`/`Bad`-prefixed identifier

Closes #377

### DIFF
--- a/skills/principle-concurrency/examples/bounded-fan-out.cs.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.cs.md
@@ -1,0 +1,63 @@
+# Bounded Fan-out — C# — SemaphoreSlim + Task.WhenAll
+
+## Problem
+
+Fetch N items concurrently in C# using `SemaphoreSlim` to cap inflight work at K=5.
+Each task calls `await sem.WaitAsync()` before fetching and releases the semaphore in a
+`finally` block to guarantee release on both success and exception. Pre-allocating
+`results[i]` means each task writes to its own slot — no locking needed. `Task.WhenAll`
+collects everything once all tasks complete, in original order.
+
+## Implementation
+
+```csharp
+// file: bounded-fan-out.cs
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+class BoundedFanOut
+{
+    static async Task<string> FetchAsync(string id)
+    {
+        await Task.Delay(10);
+        return $"result-{id}";
+    }
+
+    static async Task Main()
+    {
+        var ids = new[] { "a", "b", "c", "d", "e", "f", "g", "h" };
+        const int K = 5;
+        var sem = new SemaphoreSlim(K);
+        var results = new string[ids.Length];
+
+        var tasks = ids.Select(async (id, i) =>
+        {
+            await sem.WaitAsync();
+            try
+            {
+                results[i] = await FetchAsync(id); // each task owns its index slot
+            }
+            finally
+            {
+                sem.Release(); // always release, even on exception
+            }
+        });
+
+        await Task.WhenAll(tasks);
+        Console.WriteLine(string.Join(", ", results));
+    }
+}
+```
+
+## Common Mistake
+
+`Task.WhenAll` over a direct projection launches all tasks at once with no cap.
+
+```csharp
+// ✗ all N tasks start immediately — no semaphore, no limit
+static async Task BadFanOut(string[] ids) {
+    await Task.WhenAll(ids.Select(FetchAsync)); // ✗ unbounded inflight
+}
+```
--- a/skills/principle-concurrency/examples/bounded-fan-out.cs.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.cs.md
@@ -29,7 +29,7 @@ class BoundedFanOut
     {
         var ids = new[] { "a", "b", "c", "d", "e", "f", "g", "h" };
         const int K = 5;
-        var sem = new SemaphoreSlim(K);
+        using var sem = new SemaphoreSlim(K); // IDisposable: using ensures release on scope exit
         var results = new string[ids.Length];
 
         var tasks = ids.Select(async (id, i) =>
--- a/skills/principle-concurrency/examples/bounded-fan-out.go.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.go.md
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -52,7 +53,7 @@ func main() {
 	}
 
 	if err := g.Wait(); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 	fmt.Println(results)
 }
--- a/skills/principle-concurrency/examples/bounded-fan-out.go.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.go.md
@@ -3,9 +3,10 @@
 ## Problem
 
 Fetch N items concurrently but cap inflight work at K=5 using `errgroup.SetLimit`.
-Pre-allocate `results[i]` so each goroutine writes its own index slot — no mutex needed.
-`errgroup.WithContext` wires cancellation: if any fetch fails the remaining goroutines see
-a cancelled context. Order is preserved by index, not by completion time.
+Pre-allocate `results[i]` and shadow the loop variables so each goroutine captures its own
+`i` and writes exactly one slot — the pre-allocation, the shadow (`i, id := i, id`), and
+the indexed write are three load-bearing parts of the same ordering invariant. `errgroup.WithContext`
+propagates cancellation: a failed fetch signals the remaining goroutines to stop early.
 
 ## Implementation
 
@@ -39,13 +40,13 @@ func main() {
 	g.SetLimit(K) // at most K goroutines running at once
 
 	for i, id := range ids {
-		i, id := i, id
+		i, id := i, id // required for Go <1.22: range vars are shared across iterations
 		g.Go(func() error {
 			val, err := fetch(ctx, id)
 			if err != nil {
 				return err
 			}
-			results[i] = val // safe: each goroutine owns its index
+			results[i] = val // safe: pre-allocated slice + one goroutine per index slot
 			return nil
 		})
 	}
--- a/skills/principle-concurrency/examples/bounded-fan-out.go.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.go.md
@@ -1,0 +1,75 @@
+# Bounded Fan-out — Go — errgroup with concurrency limit
+
+## Problem
+
+Fetch N items concurrently but cap inflight work at K=5 using `errgroup.SetLimit`.
+Pre-allocate `results[i]` so each goroutine writes its own index slot — no mutex needed.
+`errgroup.WithContext` wires cancellation: if any fetch fails the remaining goroutines see
+a cancelled context. Order is preserved by index, not by completion time.
+
+## Implementation
+
+```go
+// file: bounded-fan-out.go
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+)
+
+func fetch(ctx context.Context, id string) (string, error) {
+	select {
+	case <-ctx.Done():
+		return "", ctx.Err()
+	case <-time.After(10 * time.Millisecond):
+		return "result-" + id, nil
+	}
+}
+
+func main() {
+	ids := []string{"a", "b", "c", "d", "e", "f", "g", "h"}
+	const K = 5
+	results := make([]string, len(ids))
+
+	g, ctx := errgroup.WithContext(context.Background())
+	g.SetLimit(K) // at most K goroutines running at once
+
+	for i, id := range ids {
+		i, id := i, id
+		g.Go(func() error {
+			val, err := fetch(ctx, id)
+			if err != nil {
+				return err
+			}
+			results[i] = val // safe: each goroutine owns its index
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		panic(err)
+	}
+	fmt.Println(results)
+}
+```
+
+## Common Mistake
+
+Fire-and-forget goroutines with no WaitGroup and no concurrency cap.
+
+```go
+// ✗ no WaitGroup — main exits before goroutines finish
+// ✗ no limit — all N goroutines start simultaneously
+func badFanOut(ids []string) {
+	for _, id := range ids {
+		go func(id string) { // ✗ unbounded goroutine launch
+			fmt.Println(fetch(context.Background(), id))
+		}(id)
+	}
+	// ✗ returns immediately; results are lost
+}
+```
--- a/skills/principle-concurrency/examples/bounded-fan-out.java.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.java.md
@@ -19,7 +19,7 @@ import java.util.concurrent.*;
 public class BoundedFanOut {
 
     static String fetch(String id) throws InterruptedException {
-        Thread.sleep(10);
+        Thread.sleep(10); // InterruptedException propagates up; don't swallow it in callers
         return "result-" + id;
     }
 
--- a/skills/principle-concurrency/examples/bounded-fan-out.java.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.java.md
@@ -43,7 +43,7 @@ public class BoundedFanOut {
             }
             System.out.println(results);
         } finally {
-            executor.shutdown();
+            executor.shutdown(); // safe: invokeAll already drained all tasks
         }
     }
 }
--- a/skills/principle-concurrency/examples/bounded-fan-out.java.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.java.md
@@ -57,7 +57,10 @@ public class BoundedFanOut {
 // ✗ cached pool creates an unbounded number of threads
 ExecutorService badExecutor = Executors.newCachedThreadPool();
 List<CompletableFuture<String>> futures = ids.stream()
-    .map(id -> CompletableFuture.supplyAsync(() -> fetch(id), badExecutor)) // ✗ all at once
+    .map(id -> CompletableFuture.supplyAsync(() -> {
+        try { return fetch(id); }
+        catch (InterruptedException e) { throw new RuntimeException(e); } // Supplier<T> can't throw checked
+    }, badExecutor)) // ✗ all at once
     .toList();
 CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join(); // ✗ no limit
 ```
--- a/skills/principle-concurrency/examples/bounded-fan-out.java.md
+++ b/skills/principle-concurrency/examples/bounded-fan-out.java.md
@@ -1,0 +1,63 @@
+# Bounded Fan-out — Java — fixed thread pool with invokeAll
+
+## Problem
+
+Fetch N items concurrently on a fixed thread pool of K=5 threads using `invokeAll`.
+The executor enforces the concurrency cap structurally — threads block in the pool queue
+rather than starting without limit. `invokeAll` blocks until every `Callable` completes
+and returns `List<Future<String>>` in the same order as submission, so results are
+already ordered without extra sorting.
+
+## Implementation
+
+```java
+// file: bounded-fan-out.java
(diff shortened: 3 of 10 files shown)


skills/principle-concurrency/examples/bounded-fan-out.go.md
```diff
@@ -0,0 +1,75 @@
...
+func main() {
+	ids := []string{"a", "b", "c", "d", "e", "f", "g", "h"}
+	const K = 5
+	results := make([]string, len(ids))
+
+	g, ctx := errgroup.WithContext(context.Background())
+	g.SetLimit(K) // at most K goroutines running at once
+
+	for i, id := range ids {
+		i, id := i, id
```
lugassawan: **Medium** — The `i, id := i, id` shadow declaration has no teaching comment. Without explanation, Go 1.22+ readers will read this as dead code, and ≤1.21 readers may omit it in their own code.

Suggested fix:
```suggestion
		i, id := i, id // required for Go <1.22: range vars are shared across iterations
```

skills/principle-concurrency/examples/bounded-fan-out.rs.md
```diff
@@ -0,0 +1,52 @@
...
+    format!("result-{}", id)
+}
+
+#[tokio::main]
+async fn main() {
+    let ids = vec!["a", "b", "c", "d", "e", "f", "g", "h"];
+    const K: usize = 5;
+
+    // Enumerate to carry the original index through the async boundary.
+    let mut pairs: Vec<(usize, String)> = stream::iter(ids.iter().enumerate())
```
lugassawan: **Medium** — `ids.iter().enumerate()` yields `(usize, &&str)` items. Inside the `async move` closure `id` is `&&str`, not `&str`. Deref coercion silently handles this for `'static`-string literals, but learners copying this into non-`'static` contexts get confusing lifetime errors with no obvious cause.

Suggested fix — use `.copied()` to yield `(usize, &str)` pairs directly:
```suggestion
    let mut pairs: Vec<(usize, String)> = stream::iter(ids.iter().copied().enumerate())
        .map(|(i, id)| async move { (i, fetch(id).await) })
```

skills/principle-concurrency/examples/bounded-fan-out.swift.md
```diff
@@ -0,0 +1,70 @@
...
+            }
+            group.addTask { (i, try await fetch(id: id)) }
+            active += 1
+        }
+        // Drain remainder.
+        for try await (idx, val) in group {
+            results[idx] = val
+        }
+    }
+    return (0..<ids.count).map { results[$0]! }
```
lugassawan: **Low** — `results[$0]!` force-unwraps without a documented invariant. Contradicts the Swift skill's "never force-unwrap without a documented invariant" rule and will surface as a SwiftLint warning.

Suggested fix — add an inline comment stating the invariant:
```suggestion
    return (0..<ids.count).map { results[$0]! } // safe: every index 0..<ids.count is populated by exactly one task
```

skills/principle-concurrency/examples/bounded-fan-out.go.md
```diff
@@ -0,0 +1,75 @@
...
+		g.Go(func() error {
+			val, err := fetch(ctx, id)
+			if err != nil {
+				return err
+			}
+			results[i] = val // safe: each goroutine owns its index
```
lugassawan: **Low** — The comment `// safe: each goroutine owns its index` explains the write safety but not the mechanism (pre-allocated slice + goroutine-per-index). A Go developer may not connect the allocation on line 33, the capture on line 42, and the indexed write on line 48 as the three load-bearing parts of the same invariant. Consider moving or expanding the rationale in the Problem section.

skills/principle-concurrency/examples/bounded-fan-out.java.md
```diff
@@ -0,0 +1,63 @@
...
+
+            // invokeAll blocks until all tasks finish; order matches submission.
+            List<Future<String>> futures = executor.invokeAll(tasks);
+
+            List<String> results = new ArrayList<>();
+            for (Future<String> f : futures) {
+                results.add(f.get()); // already in original order
+            }
+            System.out.println(results);
+        } finally {
```
lugassawan, line 45: **Low** — `executor.shutdown()` is called without `awaitTermination()`, which is atypical Java teardown. For this specific use it is correct (`invokeAll` already drained all tasks), but learners copying the pattern to `submit()`-based code will see tasks silently dropped.

Suggested fix — add a one-line comment:
```suggestion
        } finally {
            executor.shutdown(); // safe: invokeAll already drained all tasks
        }
```

skills/principle-concurrency/examples/bounded-fan-out.go.md
lugassawan: Addressed in 01b4b683f111b01a5570bdd1095931b0cb3793cc: added `// required for Go <1.22: range vars are shared across iterations` comment to the shadow declaration.

skills/principle-concurrency/examples/bounded-fan-out.rs.md
lugassawan: Addressed in 01b4b683f111b01a5570bdd1095931b0cb3793cc: switched to `ids.iter().copied().enumerate()` to yield `(usize, &str)` directly — no hidden deref coercion.

skills/principle-concurrency/examples/bounded-fan-out.swift.md
lugassawan: Addressed in 01b4b683f111b01a5570bdd1095931b0cb3793cc: added inline comment `// safe: every index 0..<ids.count is populated by exactly one task`.

skills/principle-concurrency/examples/bounded-fan-out.go.md
lugassawan: Addressed in 01b4b683f111b01a5570bdd1095931b0cb3793cc: expanded the Problem section to explain the pre-allocation, the shadow, and the indexed write as three load-bearing parts of the same ordering invariant.

skills/principle-concurrency/examples/bounded-fan-out.java.md
lugassawan: Addressed in 01b4b683f111b01a5570bdd1095931b0cb3793cc: added `// safe: invokeAll already drained all tasks` comment to the shutdown() call.

skills/principle-concurrency/examples/bounded-fan-out.ts.md
```diff
@@ -0,0 +1,53 @@
...
+  return `result-${id}`;
+}
+
+async function boundedFanOut(ids: string[], K: number): Promise<string[]> {
+  const results: string[] = new Array(ids.length);
+  let cursor = 0;
+
+  async function worker(): Promise<void> {
+    while (true) {
+      const i = cursor++; // safe: JS is single-threaded; no two workers overlap here
```
lugassawan: **Medium** — The comment `// safe: JS is single-threaded; no two workers overlap here` states the conclusion rather than the mechanism. "JS is single-threaded" is true of the event loop, but a learner translating this pattern to another language (or using `SharedArrayBuffer` with Worker threads) may cargo-cult the justification incorrectly. The real invariant is that `cursor++` is a synchronous read-modify-write and JS only yields at `await`.

```suggestion
      const i = cursor++; // safe: cursor++ is synchronous; JS only yields at `await`, so no two workers can read the same index
```

skills/principle-concurrency/examples/bounded-fan-out.rb.md
```diff
@@ -0,0 +1,49 @@
+# Bounded Fan-out — Ruby — thread pool consuming a Queue
+
+## Problem
+
+Fetch N items concurrently in Ruby using K=5 worker threads that drain a shared `Queue`.
+Pre-enqueue `[i, id]` pairs so each worker captures the original index alongside the item.
+Sentinel `nil` values (one per worker) signal shutdown. Workers write `results[i] = fetch(id)`
```
lugassawan, line 7: **Medium** — The Problem section asserts "no mutex needed" without explaining why. Developers coming from Java or C++ will ask whether `Array` element assignment is atomic. The actual guard is CRuby's GVL (Global VM Lock), which serializes thread execution — combined with disjoint index slots, it makes per-element writes safe. Adding one sentence makes the claim self-contained.

Suggested addition to the Problem section: _"CRuby's GVL serializes thread execution; combined with disjoint index slots, this makes the per-element write safe without an explicit mutex."_

skills/principle-concurrency/examples/bounded-fan-out.go.md
```diff
@@ -0,0 +1,76 @@
...
+		g.Go(func() error {
+			val, err := fetch(ctx, id)
+			if err != nil {
+				return err
+			}
+			results[i] = val // safe: pre-allocated slice + one goroutine per index slot
+			return nil
+		})
+	}
+
```
lugassawan, line 54: **Medium** — `panic(err)` in `main()` is the wrong idiom for expected error paths in Go. The language skill and idiomatic Go reserve `panic` for unrecoverable invariant violations; the standard pattern for `main()` is `log.Fatal(err)`. A student copying this into a library or service function would propagate a panic across a package boundary.

```suggestion
	if err := g.Wait(); err != nil {
		log.Fatal(err)
	}
```
(Requires `"log"` added to the import block.)

skills/principle-concurrency/examples/bounded-fan-out.swift.md
```diff
@@ -0,0 +1,70 @@
...
+
+func boundedFanOut(ids: [String], limit: Int) async throws -> [String] {
+    var results = [Int: String]()
+    var active = 0
+
+    try await withThrowingTaskGroup(of: (Int, String).self) { group in
+        for (i, id) in ids.enumerated() {
+            if active == limit {
+                // Drain one before adding the next to stay at limit.
+                if let (idx, val) = try await group.next() {
```
lugassawan, line 31: **Low** — Inside the `active == limit` guard, `group.next()` should never return `nil` (the group has exactly `limit` live tasks). The `if let` silently skips the case where it does, which would cause `active` to not decrement and the loop to stall. In practice the error propagation from task cancellation will unwind the group anyway, but a comment noting the invariant would make the reasoning explicit for learners.

Suggested: add `// group.next() is non-nil here: exactly `limit` tasks are live` as an inline comment.

skills/principle-concurrency/examples/bounded-fan-out.java.md
```diff
@@ -0,0 +1,63 @@
...
+
+```java
+// file: bounded-fan-out.java
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+
+public class BoundedFanOut {
+
+    static String fetch(String id) throws InterruptedException {
```
lugassawan, line 21: **Low** — `Thread.sleep(10)` throws `InterruptedException`, which propagates up as `ExecutionException` in this example (correct). However, a student adapting this by catching `InterruptedException` inside the worker lambda may swallow it without calling `Thread.currentThread().interrupt()`, breaking the thread interruption contract. A one-line comment on the `fetch` signature noting this would pre-empt the common mistake.

skills/principle-concurrency/examples/bounded-fan-out.cs.md
```diff
@@ -0,0 +1,63 @@
...
+    {
+        await Task.Delay(10);
+        return $"result-{id}";
+    }
+
+    static async Task Main()
+    {
+        var ids = new[] { "a", "b", "c", "d", "e", "f", "g", "h" };
+        const int K = 5;
+        var sem = new SemaphoreSlim(K);
```
lugassawan: **Low** — `SemaphoreSlim` implements `IDisposable`. Not disposing it is a minor resource leak in production code. The idiomatic pattern is `using var sem = new SemaphoreSlim(K)` — it makes the scope explicit and is the pattern learners should copy.

```suggestion
        using var sem = new SemaphoreSlim(K);
```

skills/principle-concurrency/examples/bounded-fan-out.ts.md
lugassawan: Addressed in a574400: updated comment to explain the mechanism — `cursor++` is synchronous and JS only yields at `await`, so no two workers can read the same index.

skills/principle-concurrency/examples/bounded-fan-out.rb.md
lugassawan: Addressed in a574400: expanded the Problem section to name CRuby's GVL as the actual guard and explain why disjoint index slots remove the shared-write race.

skills/principle-concurrency/examples/bounded-fan-out.go.md
lugassawan: Addressed in a574400: replaced `panic(err)` with `log.Fatal(err)` and added `"log"` to the import block.

skills/principle-concurrency/examples/bounded-fan-out.swift.md
lugassawan: Addressed in a574400: added comment `// group.next() is non-nil here: exactly `limit` tasks are live` to document the invariant.

skills/principle-concurrency/examples/bounded-fan-out.java.md
lugassawan: Addressed in a574400: added inline comment on the `Thread.sleep` call warning callers not to swallow `InterruptedException`.

skills/principle-concurrency/examples/bounded-fan-out.cs.md
lugassawan: Addressed in a574400: changed to `using var sem = new SemaphoreSlim(K)` with a comment noting it's `IDisposable`.

skills/principle-concurrency/examples/bounded-fan-out.java.md
```diff
@@ -0,0 +1,63 @@
...
+## Common Mistake
+
+`newCachedThreadPool` spins up one thread per task with no cap.
+
+```java
+// ✗ cached pool creates an unbounded number of threads
+ExecutorService badExecutor = Executors.newCachedThreadPool();
+List<CompletableFuture<String>> futures = ids.stream()
+    .map(id -> CompletableFuture.supplyAsync(() -> fetch(id), badExecutor)) // ✗ all at once
+    .toList();
```
lugassawan, line 64: **High** — `CompletableFuture.supplyAsync(() -> fetch(id), badExecutor)` does not compile: `fetch` throws checked `InterruptedException`, but `CompletableFuture.supplyAsync` expects a `Supplier<T>` which cannot throw checked exceptions.

**Why it matters:** The Common Mistake block is meant to teach about unbounded threads, but it also fails to compile for an unrelated reason — a checked-exception violation. A learner trying to understand or reproduce the snippet will hit a compile error that obscures the intended lesson.

**Suggested fix:** Wrap the checked exception in the lambda:
```java
.map(id -> CompletableFuture.supplyAsync(() -> {
    try { return fetch(id); }
    catch (InterruptedException e) { throw new RuntimeException(e); }
}, badExecutor))
```
Or simplify `fetch` in the bad-example block to a non-throwing variant (e.g., replace `Thread.sleep(10)` with `return "result-" + id` to keep focus on the unbounded-threads lesson).

lugassawan: Addressed in a15e679f1493ca5f580f4de0a95fd2d39b550296: wrapped `InterruptedException` in `RuntimeException` inside the `supplyAsync` lambda — `Supplier<T>` cannot propagate checked exceptions, so the wrapping is required for compilation.

skills/principle-concurrency/examples/bounded-fan-out.rs.md
```diff
@@ -0,0 +1,52 @@
...
+## Common Mistake
+
+`join_all` launches every future at once with no concurrency cap.
+
+```rust
+use futures::future::join_all;
+
+// ✗ all N futures are polled simultaneously — no cap
+async fn bad_fan_out(ids: &[&str]) -> Vec<String> {
+    join_all(ids.iter().map(|id| fetch(id))).await // ✗ unbounded inflight
```
lugassawan: **High** — `ids.iter().map(|id| fetch(id))` in the Common Mistake block has a type mismatch: `ids.iter()` on `&[&str]` yields `&&str` items, but `fetch` expects `&str`. This is a compile error unrelated to the intended lesson (unbounded concurrency), which would confuse a learner trying to reproduce the anti-pattern.

**Why it matters:** Same issue class as the Java fix in the previous pass — the bad example fails to compile for a reason orthogonal to the concurrency lesson, obscuring the teaching point.

**Suggested fix:** Add `.copied()` to convert `&&str` → `&str`:
```rust
join_all(ids.iter().copied().map(|id| fetch(id))).await // ✗ unbounded inflight
```
`.copied()` is valid here because `&str: Copy`.

lugassawan: Addressed in 2b88f110df2effcbc14461b6f7020c97abffa72e: added `.copied()` to convert `&&str` → `&str` before mapping, matching the same fix already applied in the implementation block (`ids.iter().copied().enumerate()` on line 29).

Issue #428, opened 2026-06-08
[bug] workflow-* skills leave residual files in /tmp/ after completion


lugassawan:
## Surface affected

- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — integration (`ticket-context`)
- [x] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

**Specific surface**: `workflow-pr-review`, `workflow-pr-review-followup`, `workflow-address-feedback` (and potentially other `workflow-*` skills — audit needed)

## Claude Code version

2.1.168

## Plugin version

0.1.18

## Reproduction steps

1. Run `/swe-workbench:review <PR number>` (or `--check-followup <N>`)
2. Let the skill complete successfully
3. Inspect `/tmp/swe-workbench-pr-review/`

## Expected behavior

Temporary state files written to `/tmp/swe-workbench-pr-review/` are deleted after the skill completes successfully.

## Actual behavior

Files accumulate in `/tmp/swe-workbench-pr-review/` indefinitely. After multiple PR review sessions, the directory grows unbounded:

```
18736-followup.json
18736-followup-threads.json
18736-followup-diff.patch
18736-followup-diff.txt
...
```

The worktree is cleaned up (via `rimba remove` or `git worktree remove`) but the `/tmp/` state files are never deleted. The same pattern likely exists in `workflow-address-feedback` and may extend to other workflow skills.

## Environment (optional)

macOS, zsh

## Logs / screenshots (optional)

N/A

---
_Reported via `/swe-workbench:report-issue` — plugin v0.1.18, Claude Code 2.1.168._

PR #438, opened 2026-06-12
[fix] Foreground-verify state-file reap + preflight-pr.sh + _RT root guard

lugassawan:
## Summary

- **Fix A (preflight-pr.sh):** new `runtime/preflight-pr.sh` consolidates the ~20-line inline `gh auth + fetch-pr + jq` block across three PR-workflow skills into a single script call; emits `BASE`, `HEAD_SHA`, `AUTHOR_LOGIN`, `OWNER`, `REPO`, `STATE` as `printf %q`-quoted eval-able `KEY=VALUE` lines — mirrors the `eval "$(sync-and-verify.sh)"` contract already used in `workflow-cleanup-merged`. Applies to `workflow-pr-review`, `workflow-pr-review-followup`, `workflow-address-feedback`.
- **Fix B (_RT root guard):** bind `_RT="${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}"` once before worktree entry in all 6 skills; add a hard-fail guard that aborts if `runtime/` is missing — converts "script not found → silent inline fallback" into a loud, actionable error. Replaces 22 scattered `${CLAUDE_PLUGIN_ROOT:-$(git rev-parse ...)}` occurrences.
- **Fix C (foreground reap):** pull `clean-state-files.sh` out of the backgrounded `( ... ) &` subshell in 5 skills; drop `2>/dev/null` and `|| true`; add a per-file post-check report line (`✓ state file reaped`). The worktree teardown (`rimba remove`) stays backgrounded — only the state-file reap moved to the foreground.
- **Tests:** invert the two #428 assertions that encoded the bug as correct (assert reap is NOT inside background subshell, NOT suppressed); add `test_preflight_pr_script.py`, `test_runtime_root_binding.py`; harden 5 skill test files + `test_runtime_scripts.py`.

## Test Plan

- [x] Changed skills/commands/agents load without errors (`bash scripts/validate.sh` passes)
- [x] Examples in the diff were exercised manually (`bash -n runtime/preflight-pr.sh` passes)
- [x] 1327/1327 pytest tests pass (pre-push hook verified)
- [x] `test_preflight_pr_script.py` (9 tests): verifies `preflight-pr.sh` uses `dirname "$0"`, emits `printf %q` for all 6 scalars, never echoes `title`/`body`, and passes `bash -n`
- [x] `test_runtime_root_binding.py`: verifies each of the 6 skills has a single `_RT=` binding, a hard-fail guard, and no raw `${CLAUDE_PLUGIN_ROOT:-$(git rev-parse` at script call sites
- [x] Inverted `test_pr_review_skill_state_cleanup_outside_background_subshell` and `test_*_no_suppression` confirm foreground reap in pr-review + followup

- [ ] Reproduce the original orphan: run a PR review in a worktree where `CLAUDE_PLUGIN_ROOT` is unset and confirm Fix B aborts loudly rather than resolving to the worktree root
- [ ] Run a full review flow end-to-end and confirm `/tmp/swe-workbench-pr-review/*.json` files are removed on success with `✓ state file reaped` lines visible

Issue: N/A — recurrence of #428/#429

### DIFF
--- a/runtime/preflight-pr.sh
+++ b/runtime/preflight-pr.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Consolidated pre-flight for PR-review workflows.
+# Usage: preflight-pr.sh <PR> <out_json> [fields]
+#
+# Outputs KEY=VALUE lines (printf %q quoted) for eval:
+#   BASE  HEAD_SHA  AUTHOR_LOGIN  OWNER  REPO  STATE
+#
+# title/body are NOT emitted — they are free-text and echoing them into
+# eval "$(...)" enables code injection.  Read them from <out_json> with jq.
+#
+# Resolves sibling scripts via dirname "$0" so the script works correctly
+# even when cwd is an ephemeral PR worktree (CLAUDE_PLUGIN_ROOT may be unset).
+set -euo pipefail
+
+PR="${1:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
+OUT_JSON="${2:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
+FIELDS="${3:-state,number,headRefName,baseRefName,headRefOid,title,body,author,reviewDecision}"
+
+SCRIPT_DIR="$(dirname "$0")"
+
+gh auth status >/dev/null || {
+  echo "gh not authenticated. Run 'gh auth login'." >&2
+  exit 1
+}
+
+bash "$SCRIPT_DIR/fetch-pr.sh" "$PR" "$OUT_JSON" "$FIELDS"
+
+BASE=$(jq -r .baseRefName     "$OUT_JSON")
+HEAD_SHA=$(jq -r .headRefOid  "$OUT_JSON")
+AUTHOR_LOGIN=$(jq -r .author.login "$OUT_JSON")
+STATE=$(jq -r .state          "$OUT_JSON")
+
+OWNER=$(gh repo view --json owner -q .owner.login)
+REPO=$(gh repo view --json name   -q .name)
+if [ -z "$OWNER" ] || [ "$OWNER" = "null" ] || [ -z "$REPO" ] || [ "$REPO" = "null" ]; then
+  echo "Could not determine base repo owner/name. Run 'gh repo view' to verify the current remote is set correctly." >&2
+  exit 1
+fi
+
+printf 'BASE=%q\n'         "$BASE"
+printf 'HEAD_SHA=%q\n'     "$HEAD_SHA"
+printf 'AUTHOR_LOGIN=%q\n' "$AUTHOR_LOGIN"
+printf 'OWNER=%q\n'        "$OWNER"
+printf 'REPO=%q\n'         "$REPO"
+printf 'STATE=%q\n'        "$STATE"
--- a/runtime/preflight-pr.sh
+++ b/runtime/preflight-pr.sh
@@ -16,7 +16,7 @@ PR="${1:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
 OUT_JSON="${2:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
 FIELDS="${3:-state,number,headRefName,baseRefName,headRefOid,title,body,author,reviewDecision}"
 
-SCRIPT_DIR="$(dirname "$0")"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 
 gh auth status >/dev/null || {
   echo "gh not authenticated. Run 'gh auth login'." >&2
@@ -29,6 +29,14 @@ BASE=$(jq -r .baseRefName     "$OUT_JSON")
 HEAD_SHA=$(jq -r .headRefOid  "$OUT_JSON")
 AUTHOR_LOGIN=$(jq -r .author.login "$OUT_JSON")
 STATE=$(jq -r .state          "$OUT_JSON")
+
+for _var in BASE HEAD_SHA AUTHOR_LOGIN STATE; do
+  _val="${!_var}"
+  if [ -z "$_val" ] || [ "$_val" = "null" ]; then
+    echo "preflight-pr.sh: field $_var is empty/null in $OUT_JSON — PR data may be incomplete." >&2
+    exit 1
+  fi
+done
 
 OWNER=$(gh repo view --json owner -q .owner.login)
 REPO=$(gh repo view --json name   -q .name)
--- a/tests/test_workflow_audit_emit_issues_skill.py
+++ b/tests/test_workflow_audit_emit_issues_skill.py
@@ -272,9 +272,11 @@ def test_audit_emit_cleanup_on_confirm_success():
     text = SKILL_MD.read_text()
     confirm_idx = text.find("`confirm` (literal)")
     # Use the executable invocation, not the guard check ([ -f "$_RT/runtime/clean-state-files.sh" ])
-    cleanup_idx = text.find('bash "$_RT/runtime/clean-state-files.sh"')
+    # re.search is resilient to minor quoting/whitespace changes vs. text.find()
+    cleanup_match = re.search(r'bash\s+"\$_RT/runtime/clean-state-files\.sh"', text)
     assert confirm_idx != -1, "SKILL.md must have a 'confirm' table row"
-    assert cleanup_idx != -1, "SKILL.md must reference bash invocation of clean-state-files.sh"
+    assert cleanup_match, "SKILL.md must reference bash invocation of clean-state-files.sh"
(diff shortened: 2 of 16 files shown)


runtime/preflight-pr.sh
```diff
@@ -0,0 +1,45 @@
...
+  echo "gh not authenticated. Run 'gh auth login'." >&2
+  exit 1
+}
+
+bash "$SCRIPT_DIR/fetch-pr.sh" "$PR" "$OUT_JSON" "$FIELDS"
+
+BASE=$(jq -r .baseRefName     "$OUT_JSON")
+HEAD_SHA=$(jq -r .headRefOid  "$OUT_JSON")
+AUTHOR_LOGIN=$(jq -r .author.login "$OUT_JSON")
+STATE=$(jq -r .state          "$OUT_JSON")
```
lugassawan, line 31: **Medium — null guard missing for `BASE`, `HEAD_SHA`, `AUTHOR_LOGIN`, `STATE`**

`jq -r` silently outputs the string `"null"` for absent fields. `OWNER`/`REPO` already have a guard; the four other scalars do not. A partial JSON response (e.g. draft PR with no `headRefOid`) would emit `HEAD_SHA=null` into the callers shell and produce a confusing `git diff null...HEAD` error rather than an actionable message.

Suggested addition after the four `jq` assignments:

```bash
for var_name in BASE HEAD_SHA AUTHOR_LOGIN STATE; do
  val="${!var_name}"
  if [ -z "$val" ] || [ "$val" = "null" ]; then
    echo "preflight-pr.sh: field $var_name is empty/null in $OUT_JSON — PR data may be incomplete." >&2
    exit 1
  fi
done
```

tests/test_workflow_pr_review_skill.py
```diff
@@ -26,14 +26,15 @@ def test_pr_review_skill_file_exists():
...
     assert re.search(r"OWNER\s*=.*\$\(gh repo view[^\n]*owner", text), (
-        "SKILL.md must derive OWNER via 'gh repo view --json owner' — "
+        "runtime/preflight-pr.sh must derive OWNER via 'gh repo view --json owner' — "
         "gh pr view --json has no baseRepository field; gh repo view resolves the base remote correctly"
     )
     assert re.search(r"REPO\s*=.*\$\(gh repo view[^\n]*name", text), (
-        "SKILL.md must derive REPO via 'gh repo view --json name' — "
+        "runtime/preflight-pr.sh must derive REPO via 'gh repo view --json name' — "
         "gh pr view --json has no baseRepository field; gh repo view resolves the base remote correctly"
     )
```
lugassawan: **Medium — redirected owner/repo tests are now duplicates of `test_preflight_pr_script.py`**

`test_pr_review_skill_owner_repo_from_gh_repo_view`, `test_followup_skill_owner_repo_from_gh_repo_view`, and `test_address_feedback_skill_owner_repo_from_gh_repo_view` all read `runtime/preflight-pr.sh` after Fix A moved the logic there. They are now identical assertions against the same file, already covered by `test_preflight_pr_script.py`. More critically, none of them verify that each skill actually invokes `preflight-pr.sh` (that check lives in `test_runtime_scripts.py:test_preflight_pr_referenced_in_*`).

Options:
1. Delete the three redirected tests (redundancy with `test_preflight_pr_script.py` + `test_runtime_scripts.py` means no coverage gap).
2. If keeping them, rename them to `test_preflight_pr_script_owner_derivation` and consolidate into `test_preflight_pr_script.py`.

tests/test_workflow_pr_review_skill.py
```diff
@@ -117,21 +119,44 @@ def test_pr_review_skill_cleanup_deletes_pr_json():
...
+def test_pr_review_skill_state_cleanup_outside_background_subshell():
+    """clean-state-files.sh must NOT appear inside the background ( ... ) & subshell.
+
+    The reap must run in the foreground so failures surface immediately rather than being
+    silently dropped by the backgrounded, output-suppressed worktree-teardown subshell.
+    This is the inverse of the previous #428 assertion, which encoded the bug as correct.
+    """
     text = SKILL_MD.read_text()
-    # The subshell is opened with '(' and closed with ') &' — find it
     subshell_match = re.search(r'\(\s*bash.*?clean-state-files\.sh.*?\)\s*&', text, re.DOTALL)
```
lugassawan: **Medium — `test_*_cleanup_outside_background_subshell` DOTALL regex is fragile**

```python
subshell_match = re.search(r'\(\s*bash.*?clean-state-files\.sh.*?\)\s*&', text, re.DOTALL)
```

The non-greedy `.*?` with `re.DOTALL` stops at the first `)&` following any `(bash` token. A future code example containing `(bash ... ) &` anywhere before the foreground reap block would cause the regex to match that example instead — making the negative assertion vacuously true (it would stop scanning before reaching `clean-state-files.sh`). Suggested tighter form that anchors on the script name:

```python
subshell_match = re.search(
    r'\([^)]*clean-state-files\.sh[^)]*\)\s*&',
    text
)
```

No `re.DOTALL` needed — the match is now constrained to a single balanced `(...)` pair.

runtime/preflight-pr.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+# Resolves sibling scripts via dirname "$0" so the script works correctly
+# even when cwd is an ephemeral PR worktree (CLAUDE_PLUGIN_ROOT may be unset).
+set -euo pipefail
+
+PR="${1:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
+OUT_JSON="${2:?Usage: preflight-pr.sh <PR> <out_json> [fields]}"
+FIELDS="${3:-state,number,headRefName,baseRefName,headRefOid,title,body,author,reviewDecision}"
+
+SCRIPT_DIR="$(dirname "$0")"
```
lugassawan: **Low — `dirname "$0"` returns a relative path when invoked as `bash runtime/preflight-pr.sh`**

Non-blocking since skills always invoke via `"$_RT/runtime/preflight-pr.sh"` (absolute), but the idiomatic fix is safer for future callers:

```bash
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
```

This ensures `$SCRIPT_DIR` is always an absolute path regardless of invocation style.

tests/test_workflow_audit_emit_issues_skill.py
```diff
@@ -263,16 +263,20 @@ def test_audit_emit_uses_clean_state_files_for_deletion():
 
 
 def test_audit_emit_cleanup_on_confirm_success():
-    """clean-state-files.sh must appear in the confirm-success row of the Phase 4 table."""
+    """clean-state-files.sh invocation must appear after the Phase 4 confirm table row.
+
+    After Fix C, the reap block is a separate section below the table (not inline in the row).
+    We search for the bash invocation (not the [ -f ... ] guard check in the Preamble) so
+    the ordering check finds the correct occurrence.
+    """
```
lugassawan, line 271: **Low — `text.find('bash "$_RT/runtime/clean-state-files.sh"')` is brittle on quoting changes**

If the invocation line is ever reformatted (different quote style, extra whitespace), `find()` returns `-1` and the test fails with a misleading message. A regex search is more resilient:

```python
import re
cleanup_idx_match = re.search(r'bash\s+"\$_RT/runtime/clean-state-files\.sh"', text)
assert cleanup_idx_match, "SKILL.md must reference bash invocation of clean-state-files.sh"
cleanup_idx = cleanup_idx_match.start()
```

runtime/preflight-pr.sh
lugassawan: Addressed in 83a6e0767a6d9ea60fcb72698d9280908921170c: added null guard loop for BASE/HEAD_SHA/AUTHOR_LOGIN/STATE after jq assignments; also fixed dirname to use cd+pwd.

tests/test_workflow_pr_review_skill.py
lugassawan: Addressed in 83a6e0767a6d9ea60fcb72698d9280908921170c: deleted 3 redundant *_owner_repo_from_gh_repo_view tests (covered by test_preflight_pr_script.py + test_runtime_scripts.py).

tests/test_workflow_pr_review_skill.py
lugassawan: Addressed in 83a6e0767a6d9ea60fcb72698d9280908921170c: tightened regex from DOTALL non-greedy to r'\([^)]*clean-state-files\.sh[^)]*\)\s*&' in both pr-review and followup tests.

runtime/preflight-pr.sh
lugassawan: Addressed in 83a6e0767a6d9ea60fcb72698d9280908921170c: fixed SCRIPT_DIR to use cd+pwd for absolute path.

PR #439, opened 2026-06-12
[feat] Add worked Core-4 design-pattern examples (9 OO languages)

lugassawan:
## Summary
- Add 36 worked example files (`skills/principle-design-patterns/examples/`) covering the Core-4 design patterns × 9 OO languages
- Patterns: Strategy (checkout discount pricing), Observer (order-status notifications), Decorator (retry + logging fetch), Factory Method (notification channel)
- Languages: C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript
- Each file follows the repo's example template: H1 / Problem / Implementation / Common Mistake, ≤120 lines
- Add one `>` pointer blockquote to `SKILL.md` (on-demand, not auto-loaded)
- Mirrors precedent set by `principle-concurrency/examples/` (PR #434) and `principle-clean-architecture/examples/`

## Test Plan
- [x] Changed skills/commands/agents load without errors
- [x] Examples in the diff were exercised manually
- [x] `bash scripts/validate.sh` — 0 issues (all 36 files ≤120 lines; existing checks green)
- [x] `pytest tests/ -v` — 1324/1324 pass
- [x] `ls skills/principle-design-patterns/examples/ | wc -l` → 36
- [x] Each file opens with H1/Problem/Implementation/Common-Mistake structure and a language-tagged fence
- [x] SKILL.md carries exactly one new `>` pointer line; `examples/` absent from any frontmatter auto-load path

### Type-tailored checks ([feat])
- [x] New behaviour (examples dir) is present on disk under the expected path
- [x] No regressions in existing skill triggers or catalog checks

Closes #375

PR #441, opened 2026-06-12
[feat] Add worked error-handling examples to principle-error-handling (9 OO languages)

### LINKED ISSUE #376: [feat] Add worked examples to principle-error-handling (9 OO languages)
## Goal

Add an `examples/` dir to `principle-error-handling` with one shared scenario implemented across all 9 OO languages.

## Motivation

Error handling is where language idioms differ most fundamentally — Go's explicit multi-return errors, Rust's `Result`/`?`, Java's checked exceptions, Swift's `throws`, Python's `try/except`. Prose alone leaves these contrasts abstract. A single worked scenario (e.g. parse-and-validate a config file) across 9 languages makes the idiom differences concrete.

The `examples/` convention is documented in `docs/extending.md` and enforced by `scripts/validate.py` (≤120 lines/file, on-demand pointer in SKILL.md, never auto-loaded). `principle-clean-architecture` is currently the only skill using this pattern.

## Surface

- [x] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — `integration-*`
- [ ] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch

    skills/principle-error-handling/examples/
      <scenario>.cs.md
      <scenario>.go.md
      <scenario>.java.md
      <scenario>.kt.md
      <scenario>.py.md
      <scenario>.rb.md
      <scenario>.rs.md
      <scenario>.swift.md
      <scenario>.ts.md

Pick one shared scenario (e.g. parse-and-validate a config file: missing keys, wrong types, IO errors) that exercises the language's idiomatic error-handling path end-to-end. Follow the existing convention: "Problem / Implementation / Common Mistake", `# file:` headers, ≤120 lines per file.

Languages (9, OO set): C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript.
Bash and SQL are out of scope.

## Acceptance

- [ ] `principle-error-handling/examples/` dir created with one shared scenario × 9 language files.
- [ ] SKILL.md gains an on-demand pointer line; nothing is auto-loaded.

lugassawan:
## Summary

- Add 27 worked error-handling example files to `skills/principle-error-handling/examples/` — 3 scenarios × 9 OO languages (C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript)
- **Scenario 1 — `config.<lang>.md`:** parse & validate a `key=value` config file; three error tiers (IO / parse / validation); each language uses its idiomatic style (Go `%w` wrap, Rust `Result`+`?`+typed enum, TypeScript discriminated union, Kotlin sealed class, Java/C# checked exception hierarchy, Python `raise…from`, Ruby `rescue` re-raise, Swift `throws`+`do/catch`)
- **Scenario 2 — `fetch.<lang>.md`:** HTTP fetch with retry; injected `Transport` interface + deterministic `FakeTransport`; classifies transient (5xx/timeout) vs permanent (4xx); exponential backoff + jitter; per-attempt timeout modelled; exhausted sentinel after `maxRetries`
- **Scenario 3 — `withdraw.<lang>.md`:** in-memory account withdrawal; three typed errors in order (invalid amount → frozen → insufficient funds); domain returns errors **without logging**; caller logs once at the boundary (log-once discipline)
- Add one pointer blockquote to `SKILL.md` immediately before `## Red Flags`, listing all 3 scenarios + 9 languages + "(read on demand — not auto-loaded)"

> **Scope note:** Issue #376's written acceptance said "one shared scenario × 9 files." During planning review the scope was expanded to three scenarios × 9 languages = 27 files, mirroring the Core-4 precedent in PR #439. This expansion is intentional.

## Test Plan

- [x] `bash scripts/validate.sh` → `All checks passed.` (enforces ≤120 lines/example, ≤150 lines SKILL.md)
- [x] `python3 -m pytest tests/` → 1324/1324 pass (no regression)
- [x] `wc -l skills/principle-error-handling/examples/*.md` → max 119 lines, all ≤120
- [x] `ls skills/principle-error-handling/examples/ | wc -l` → 27 files
- [x] `wc -l skills/principle-error-handling/SKILL.md` → 102 lines (cap 150)
- [x] Spot-read `config.rs.md`, `fetch.go.md`, `withdraw.py.md` — idiom and `✗`-annotated Common Mistake render correctly
- [x] Two-stage review (spec compliance + code quality) per scenario; all Critical and Important issues fixed before push

- [x] New behaviour present: 27 example files exist under `examples/` with correct naming convention
- [x] SKILL.md pointer added and loads only on demand (not in frontmatter `description:`)
- [x] Each file exercises the skill sections it is mapped to (config → Errors as Values / Wrap Don't Swallow; fetch → Retries and Backoff / Timeouts; withdraw → Sentinel vs Typed / Where to Handle)

Closes #376

### DIFF
--- a/skills/principle-error-handling/examples/config.swift.md
+++ b/skills/principle-error-handling/examples/config.swift.md
@@ -1,0 +1,88 @@
+# Error Handling — Swift — Config Parse & Validate
+
+## Problem
+
+Swift's `throws` + typed `enum Error` make every failure site explicit in the
+function signature. A `ConfigError` enum with associated values for IO, parse, and
+validation lets callers `catch` each case individually in a `do/catch` block, while
+`try` propagation keeps the call chain concise without hiding which tier failed.
+
+## Implementation
+
+```swift
+// file: config.swift
+import Foundation
+
+enum ConfigError: Error {
+    case io(Error)
+    case parse(line: Int, reason: String)
+    case validation(field: String, reason: String)
+}
+
+struct Config { let host: String; let port: Int }
+
+func parseConfig(path: String) throws -> Config {
+    let text: String
+    do {
+        text = try String(contentsOfFile: path, encoding: .utf8)
+    } catch {
+        throw ConfigError.io(error)
+    }
+
+    var kv = [String: String]()
+    let lines = text.components(separatedBy: "\n")
+    for (idx, raw) in lines.enumerated() {
+        let line = raw.trimmingCharacters(in: .whitespaces)
+        if line.isEmpty || line.hasPrefix("#") { continue }
+        guard let eq = line.firstIndex(of: "="), eq != line.startIndex else {
+            throw ConfigError.parse(line: idx + 1, reason: "missing '=' separator")
+        }
+        let key = String(line[..<eq]).trimmingCharacters(in: .whitespaces)
+        guard !key.isEmpty else {
+            throw ConfigError.parse(line: idx + 1, reason: "empty key")
+        }
+        kv[key] = String(line[line.index(after: eq)...]).trimmingCharacters(in: .whitespaces)
+    }
+
+    return try validateConfig(kv)
+}
+
+func validateConfig(_ kv: [String: String]) throws -> Config {
+    guard let host = kv["host"], !host.isEmpty else {
+        throw ConfigError.validation(field: "host", reason: "required key missing")
+    }
+    guard let portStr = kv["port"] else {
+        throw ConfigError.validation(field: "port", reason: "required key missing")
+    }
+    guard let port = Int(portStr) else {
+        throw ConfigError.validation(field: "port", reason: "'\(portStr)' is not an integer")
+    }
+    guard (1...65535).contains(port) else {
+        throw ConfigError.validation(field: "port", reason: "\(port) out of range 1-65535")
+    }
+    return Config(host: host, port: port)
+}
+```
+
+```swift
+// file: main.swift
+do {
+    let cfg = try parseConfig(path: "app.conf")
+    print("host=\(cfg.host) port=\(cfg.port)")
+} catch ConfigError.io(let e) {
+    fputs("IO error: \(e.localizedDescription)\n", stderr)
+} catch ConfigError.parse(let line, let reason) {
+    fputs("Parse error line \(line): \(reason)\n", stderr)
+} catch ConfigError.validation(let field, let reason) {
+    fputs("Validation error '\(field)': \(reason)\n", stderr)
+}
+```
+
+## Common Mistake
+
+Using `try?` — the typed `ConfigError` is discarded and the caller receives `nil` with no information about which tier failed or why.
+
+```swift
+let cfg = try? parseConfig(path: "app.conf")  // ✗ all three error cases collapsed to nil
+// caller cannot distinguish missing file from invalid port — no diagnostic available
+```
--- a/skills/principle-error-handling/examples/config.swift.md
+++ b/skills/principle-error-handling/examples/config.swift.md
@@ -34,13 +34,14 @@ func parseConfig(path: String) throws -> Config {
     for (idx, raw) in lines.enumerated() {
         let line = raw.trimmingCharacters(in: .whitespaces)
         if line.isEmpty || line.hasPrefix("#") { continue }
-        guard let eq = line.firstIndex(of: "="), eq != line.startIndex else {
+        guard line.contains("=") else {
             throw ConfigError.parse(line: idx + 1, reason: "missing '=' separator")
         }
-        let key = String(line[..<eq]).trimmingCharacters(in: .whitespaces)
-        guard !key.isEmpty else {
+        let eq = line.firstIndex(of: "=")!
+        guard eq != line.startIndex else {
             throw ConfigError.parse(line: idx + 1, reason: "empty key")
         }
+        let key = String(line[..<eq]).trimmingCharacters(in: .whitespaces)
         kv[key] = String(line[line.index(after: eq)...]).trimmingCharacters(in: .whitespaces)
     }
 
--- a/skills/principle-error-handling/examples/config.swift.md
+++ b/skills/principle-error-handling/examples/config.swift.md
@@ -34,15 +34,15 @@ func parseConfig(path: String) throws -> Config {
     for (idx, raw) in lines.enumerated() {
         let line = raw.trimmingCharacters(in: .whitespaces)
         if line.isEmpty || line.hasPrefix("#") { continue }
-        guard line.contains("=") else {
+        let parts = line.split(separator: "=", maxSplits: 1)
+        guard parts.count == 2 else {
             throw ConfigError.parse(line: idx + 1, reason: "missing '=' separator")
         }
-        let eq = line.firstIndex(of: "=")!
-        guard eq != line.startIndex else {
+        let key = parts[0].trimmingCharacters(in: .whitespaces)
+        guard !key.isEmpty else {
             throw ConfigError.parse(line: idx + 1, reason: "empty key")
         }
-        let key = String(line[..<eq]).trimmingCharacters(in: .whitespaces)
-        kv[key] = String(line[line.index(after: eq)...]).trimmingCharacters(in: .whitespaces)
+        kv[key] = parts[1].trimmingCharacters(in: .whitespaces)
     }
 
     return try validateConfig(kv)
--- a/skills/principle-error-handling/examples/fetch.go.md
+++ b/skills/principle-error-handling/examples/fetch.go.md
@@ -40,7 +40,7 @@ func (f *FakeTransport) Fetch(url string) (*Response, error) {
 // file: fetch.go
 package fetch
 
-import ("errors"; "fmt"; "math/rand")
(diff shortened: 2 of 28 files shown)


skills/principle-error-handling/examples/config.swift.md
```diff
@@ -0,0 +1,89 @@
...
+
+    var kv = [String: String]()
+    let lines = text.components(separatedBy: "\n")
+    for (idx, raw) in lines.enumerated() {
+        let line = raw.trimmingCharacters(in: .whitespaces)
+        if line.isEmpty || line.hasPrefix("#") { continue }
+        guard line.contains("=") else {
+            throw ConfigError.parse(line: idx + 1, reason: "missing '=' separator")
+        }
+        let eq = line.firstIndex(of: "=")!
```
lugassawan: **High** — Force-unwrap `!` in idiomatic example code contradicts the principle being taught.

`line.firstIndex(of: "=")!` will crash if `firstIndex` returns nil. The preceding guard makes it safe in this snapshot, but teaching `!` after a guard is the anti-pattern the Swift skill warns against — readers will copy this idiom.

**Suggested fix** — use `split(separator:maxSplits:)` instead:
```swift
let parts = line.split(separator: "=", maxSplits: 1)
guard parts.count == 2, !parts[0].trimmingCharacters(in: .whitespaces).isEmpty else {
    throw ConfigError.parse(line: idx + 1, reason: "empty key")
}
let key = parts[0].trimmingCharacters(in: .whitespaces)
kv[key]  = parts[1].trimmingCharacters(in: .whitespaces)
```

skills/principle-error-handling/examples/fetch.kt.md
```diff
@@ -0,0 +1,107 @@
...
+    transport: Transport,
+    url: String,
+    maxRetries: Int,
+    timeoutMs: Int,
+): Result<Response> {
+    val baseMs = 100.0
+    for (attempt in 0 until maxRetries) {
+        val result = transport.fetch(url)
+        if (result.isSuccess) return result
+        val err = result.exceptionOrNull()!!
```
lugassawan: **Medium** — `!!` (double-bang) force-unwrap in idiomatic example; Kotlin skill bans this in production code.

`result.exceptionOrNull()!!` is safe here by invariant, but `!!` is banned precisely because such invariants erode over time — and this is a teaching file. Readers will copy the pattern.

**Suggested fix:**
```kotlin
val err = result.exceptionOrNull() ?: return result   // isSuccess already returned above
if (!isTransient(err)) return result                   // permanent — bubble immediately
```

skills/principle-error-handling/examples/fetch.ts.md
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+```
+
+```typescript
+// file: fetch.ts
+import type { FetchResult, Transport } from "./transport";
+
+function isTransient(result: FetchResult): boolean {
+  if ("error" in result) return result.error === "timeout" || result.error === "network";
```
lugassawan: **Medium** — Presence-check (`"error" in result`) is inconsistent with `config.ts.md` which uses the `ok: true/false` discriminant pattern recommended by the TypeScript skill.

`"error" in result` can narrowing-fail on malformed inputs and teaches a different pattern than the sibling config example — readers won't know which to prefer.

**Suggested fix** — align with `config.ts.md`:
```typescript
export type FetchResult =
  | { ok: true;  status: number; body: string }
  | { ok: false; error: "timeout" | "network" | "exhausted" | "permanent"; statusCode?: number };
```
Update `isTransient`, `isPermanent`, and all consumers to use `result.ok` as the discriminant.

skills/principle-error-handling/examples/fetch.go.md
```diff
@@ -0,0 +1,111 @@
...
+	if f.Attempt < 2 { return nil, ErrTimeout }
+	return &Response{Status: 200, Body: "OK"}, nil
+}
+```
+
+```go
+// file: fetch.go
+package fetch
+
+import ("errors"; "fmt"; "math/rand")
```
lugassawan: **Low** — `math/rand` top-level functions (`rand.Float64()`) are deprecated in Go 1.20+ in favour of `math/rand/v2`. Since the jitter line is already a stand-in, consider a note:
```go
// real impl: rand.New(rand.NewSource(time.Now().UnixNano())).Float64()
// or math/rand/v2: rand.Float64()
```

skills/principle-error-handling/examples/fetch.rs.md
```diff
@@ -0,0 +1,119 @@
...
+## Implementation
+
+```rust
+// file: transport.rs
+#[derive(Debug)]
+pub struct Response { pub status: u16, pub body: String }
+
+#[derive(Debug)]
+pub enum FetchError {
+    Transient(u16),   // 5xx status
```
lugassawan, line 19: **Low** — `FetchError::Transient(u16)` is declared but never constructed anywhere in the example (`FakeTransport` only emits `Timeout` and `Permanent`). Rustc would emit an `unused variant` warning. Either have `FakeTransport` emit `Transient(503)` on one attempt, or remove the variant and note that 5xx cases are modelled via `Timeout` in this simplified example.

skills/principle-error-handling/examples/fetch.rb.md
```diff
@@ -0,0 +1,103 @@
...
+class FetchError < StandardError; end
+class TransientError < FetchError
+  attr_reader :status
+  def initialize(status) = (@status = status; super("transient #{status}"))
+end
+class PermanentError < FetchError
+  attr_reader :status
+  def initialize(status) = (@status = status; super("permanent #{status}"))
+end
+class TimeoutError  < FetchError; end
```
lugassawan: **Low** — Class named `TimeoutError` may shadow `Timeout::Error` (available as `::TimeoutError` in some Ruby versions when `require 'timeout'` is loaded). Renaming to `FetchTimeoutError` (consistent with the Python example) removes any potential confusion.

skills/principle-error-handling/examples/config.swift.md
lugassawan: Addressed in 14080c0: replaced force-unwrap `!` with `split(separator:maxSplits:)` — no force-unwrap needed.

skills/principle-error-handling/examples/fetch.kt.md
lugassawan: Addressed in 14080c0: replaced `!!` with `?: return result` safe unwrap.

skills/principle-error-handling/examples/fetch.ts.md
lugassawan: Addressed in 14080c0: aligned to `ok`-discriminated union matching `config.ts.md`.

skills/principle-error-handling/examples/fetch.go.md
lugassawan: Addressed in 14080c0: added deprecation comment on `math/rand` import pointing to `math/rand/v2`.

skills/principle-error-handling/examples/fetch.rs.md
lugassawan: Addressed in 14080c0: FakeTransport now emits Transient(503) on attempt 0 so the variant is exercised.

skills/principle-error-handling/examples/fetch.rb.md
lugassawan: Addressed in 14080c0: renamed TimeoutError to FetchTimeoutError throughout, matching the Python example.

skills/principle-error-handling/examples/fetch.ts.md
```diff
@@ -0,0 +1,115 @@
...
+  maxRetries: number,
+  timeoutMs: number,
+): FetchResult {
+  const BASE_MS = 100;
+
+  for (let attempt = 0; attempt < maxRetries; attempt++) {
+    const result = transport.fetch(url);
+
+    if (result.ok && result.status < 400) return result;
+    if (isPermanent(result)) return result;           // bubble permanent immediately
```
lugassawan, line 63: **Important** — `isTransient` is defined but never called in `fetchWithRetry`.

The loop at lines 59–67 only gates on `isPermanent` (line 63); everything else is retried implicitly. `isTransient` (lines 38–41) is dead code.

This means the example does not actually demonstrate transient-vs-permanent classification — a transport returning `{ ok: false, error: "exhausted" }` would be silently retried rather than bubbled, and any unknown error variant is treated as retryable without going through the classify step.

All peer language examples (`fetch.go.md`, `fetch.kt.md`, `fetch.rs.md`, `fetch.py.md`, `fetch.rb.md`) use an explicit classify gate. Add one here for consistency:

```typescript
if (result.ok && result.status < 400) return result;
if (isPermanent(result)) return result;           // bubble permanent immediately
if (!isTransient(result)) return result;          // non-transient — bubble
```

Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)).

lugassawan: Addressed in 74a83d5: added explicit `isTransient` gate after `isPermanent` check — classify step is now wired into the retry loop.

PR #442, opened 2026-06-12
[feat] Add worked Light DDD examples to principle-ddd (9 OO languages)

### LINKED ISSUE #440: [feat] Add worked examples to principle-ddd (9 OO languages)
## Goal

Add an `examples/` dir to `principle-ddd` with one shared "Light DDD" scenario implemented across all 9 OO languages, grounding the skill's currently code-free building-block definitions in worked implementations.

## Motivation

`principle-ddd/SKILL.md` is the most abstract skill in the `principle-*` set — ~71 lines of glossary (value object, aggregate, repository, domain event, bounded context) with **zero code blocks**. Translating these definitions into real types is exactly where DDD newcomers stall: what *is* an immutable value object in Rust vs Java vs Python? How does an aggregate root enforce an invariant as the sole mutation entry point? Worked code makes the tactical patterns concrete in a way prose cannot.

This extends the worked-examples convention already shipped for `principle-clean-architecture`, `principle-design-patterns` (#375), `principle-concurrency` (#377), and `principle-solid` (#378). The pattern is documented in `docs/extending.md` and enforced by `scripts/validate.py` (`check_examples()`: ≤120 lines/file; on-demand pointer in SKILL.md; never auto-loaded).

DDD was not part of the original series (#373), which scoped the four principles where *per-language idioms diverge most*. DDD is proposed here on a different axis: it is the principle whose **abstraction gap** is largest, making it the highest-value remaining candidate for concretization.

## Surface

- [x] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — integration (`ticket-context`)
- [ ] Skill — `workflow-*`
- [ ] Subagent
- [ ] Slash command
- [ ] Hook
- [ ] Documentation
- [ ] Build / CI / release

## Sketch

A deliberately scoped **"Light DDD"** scenario — value object + one aggregate enforcing one invariant + repository port — chosen to fit the ≤120-line/file cap. Full DDD (domain events, context maps, cross-aggregate coordination) is intentionally out of scope: it overruns the line budget in verbose languages and dilutes per-file focus.

Proposed scenario: an **`Order`** aggregate.
- `Money` value object — immutable, compared by value. This is the most idiom-divergent piece: Rust newtype, Java/C# `record`, Python frozen dataclass, Kotlin `data class`, Go struct with no setters, Ruby `Comparable`, Swift struct, TS `readonly`.
- `Order` aggregate root — the only mutation entry point; enforces the invariant *"lines cannot be added once the order is submitted."*
- `OrderRepository` — collection-like port interface in the domain layer (interface only, no implementation).

```
skills/principle-ddd/examples/
  order.cs.md
  order.go.md
  order.java.md
  order.kt.md
  order.py.md
  order.rb.md
  order.rs.md
  order.swift.md
  order.ts.md
```

Follow the existing convention: "Problem / Implementation / Common Mistake", `# file:` (or `// file:`) per-component headers, ≤120 lines per file.

Languages (9, OO set): C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript. Bash and SQL are out of scope (not OO-shaped).

## Acceptance

- [ ] `principle-ddd/examples/` created with the `Order` scenario × 9 language files.
- [ ] SKILL.md gains a single on-demand pointer line (blockquote naming the scenario + 9 languages + "read on demand — not auto-loaded"); nothing is auto-loaded.
- [ ] Each example demonstrates an immutable value object, an aggregate root enforcing one invariant, and a repository port.
- [ ] "Common Mistake" sections show a real anti-pattern (e.g. mutating an aggregate's internals directly, or a value object carrying identity/mutability).

lugassawan:
## Summary
- Add `skills/principle-ddd/examples/` with 9 worked "Light DDD" Order-aggregate examples (one per OO language: C#, Go, Java, Kotlin, Python, Ruby, Rust, Swift, TypeScript)
- Each file shows: `Money` value object (immutable, value equality, same-currency `plus()` guard), `Order` aggregate root (private `lines`, invariant enforced at the sole `addLine` entry point, Draft→Submitted state transition), `OrderRepository` domain port (interface/trait/protocol only), usage snippet (construct → addLine → submit → rejected addLine), and a per-language Common Mistake
- Append a single on-demand pointer blockquote to `skills/principle-ddd/SKILL.md` after the "Light DDD is the floor" sentence — nothing auto-loaded

## Test Plan
- [x] Changed skills/commands/agents load without errors (`bash scripts/validate.sh` — 0 failures, 1324 tests pass)
- [x] Examples in the diff were exercised manually — read 3 files (Rust, Go, Python); confirmed immutable VO, aggregate root invariant, port-only repository, and real Common Mistake in each
- [x] All 9 example files ≤120 lines (`wc -l skills/principle-ddd/examples/*.md` — max 120 for Go, min 82 for Kotlin)

- [x] New behaviour demonstrated: invariant-guarded `addLine` and `submit()` transition verified in the usage snippet of each language file
- [x] `OrderRepository` is a port only (no implementation) in all 9 files
- [x] `Money.plus()` same-currency guard present in all 9 languages
- [x] `✗` Common Mistake annotation present in all 9 files

Closes #440

### DIFF
--- a/skills/principle-ddd/examples/order.go.md
+++ b/skills/principle-ddd/examples/order.go.md
@@ -12,6 +12,7 @@ exceptions) and `OrderRepository` is a plain `interface` — a domain port with
 zero infrastructure coupling.
 
 ## Implementation
+
 ```go
 // file: money.go
 package ddd
@@ -111,7 +112,6 @@ func main() {
 ## Common Mistake
 
 Exporting `Lines` as a public field lets callers `append` after `Submit()`, bypassing the aggregate root.
-
 ```go
 type Order struct {
 	ID     string
--- a/skills/principle-ddd/examples/order.go.md
+++ b/skills/principle-ddd/examples/order.go.md
@@ -1,0 +1,120 @@
+# Light DDD — Go — Order Aggregate
+
+## Problem
+
+Go has no `immutable` or `final` keyword — value object semantics come from
+unexported fields, value receivers, and the absence of setters. Callers cannot
+mutate `Money` directly; they go through constructors or methods that return new
+values. The `Order` aggregate root hides `lines []orderLine` behind an unexported
+field so only `AddLine` and `Submit` can mutate it, keeping the draft-only
+invariant enforceable. `AddLine` returns `error` (Go's idiomatic flow — no
+exceptions) and `OrderRepository` is a plain `interface` — a domain port with
+zero infrastructure coupling.
+
+## Implementation
+```go
+// file: money.go
+package ddd
+
+import "fmt"
+
+type Money struct {
+	minorUnits int64
+	currency   string
+}
+
+func NewMoney(minorUnits int64, currency string) Money {
+	return Money{minorUnits: minorUnits, currency: currency}
+}
+
+func (m Money) Plus(other Money) (Money, error) {
+	if m.currency != other.currency {
+		return Money{}, fmt.Errorf("currency mismatch: %s vs %s", m.currency, other.currency)
+	}
+	return Money{minorUnits: m.minorUnits + other.minorUnits, currency: m.currency}, nil
+}
+
+func (m Money) Equal(other Money) bool {
+	return m.minorUnits == other.minorUnits && m.currency == other.currency
+}
+```
+
+```go
+// file: order.go
+package ddd
+
+import "errors"
+
+type status int
+
+const (
+	draft status = iota
+	submitted
+)
+
+type orderLine struct {
+	sku   string
+	price Money
+}
+
+type Order struct {
+	ID     string
+	status status
+	lines  []orderLine // unexported: only mutated through aggregate root methods
+}
+
+func NewOrder(id string) *Order {
+	return &Order{ID: id}
+}
+
+func (o *Order) AddLine(sku string, price Money) error {
+	if o.status == submitted {
+		return errors.New("cannot add lines to a submitted order")
+	}
+	o.lines = append(o.lines, orderLine{sku: sku, price: price})
+	return nil
+}
+
+func (o *Order) Submit()       { o.status = submitted }
+func (o *Order) LineCount() int { return len(o.lines) }
+```
+
+```go
+// file: repository.go
+package ddd
+
+type OrderRepository interface {
+	Find(id string) (*Order, bool)
+	Save(order *Order)
+}
+```
+
+```go
+// file: main.go
+package main
+
+import (
+	"ddd"
+	"fmt"
+)
+
+func main() {
+	order := ddd.NewOrder("ord-1")
+	_ = order.AddLine("SKU-1", ddd.NewMoney(1299, "USD"))
+	order.Submit()
+	if err := order.AddLine("SKU-2", ddd.NewMoney(500, "USD")); err != nil {
+		fmt.Println("rejected:", err) // rejected: cannot add lines to a submitted order
+	}
+}
+```
+
+## Common Mistake
+
+Exporting `Lines` as a public field lets callers `append` after `Submit()`, bypassing the aggregate root.
+
+```go
+type Order struct {
+	ID     string
+	Lines []OrderLine // ✗ callers can append after Submit() — invariant broken (OrderLine also exported)
+}
+```
--- a/skills/principle-ddd/examples/order.java.md
+++ b/skills/principle-ddd/examples/order.java.md
@@ -1,0 +1,99 @@
+# Light DDD — Java — Order Aggregate
+
+## Problem
+
+Java records give value-object semantics with zero boilerplate: `equals`, `hashCode`,
+and accessors are generated, so `Money` is compared by value without a custom `equals`.
+The `Order` aggregate root keeps its `lines` list private; the only mutation paths are
+`addLine` and `submit`, which enforce the invariant that lines cannot be added once
+an order is submitted.
+
+## Implementation
+
+```java
+// file: Money.java
+public record Money(long minorUnits, String currency) {
+    public Money plus(Money other) {
+        if (!this.currency.equals(other.currency)) {
+            throw new IllegalArgumentException(
+                "currency mismatch: " + this.currency + " vs " + other.currency);
+        }
+        return new Money(this.minorUnits + other.minorUnits, this.currency);
+    }
+}
+```
+
+```java
+// file: Order.java
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class Order {
+
+    public record OrderLine(String sku, Money price) {}
+
+    private enum Status { DRAFT, SUBMITTED }
+
+    private final String id;
+    private Status status = Status.DRAFT;
+    private final List<OrderLine> lines = new ArrayList<>();
+
+    public Order(String id) { this.id = id; }
+
+    public void addLine(String sku, Money price) {
+        if (status == Status.SUBMITTED) {
+            throw new IllegalStateException("cannot add lines to a submitted order");
+        }
+        lines.add(new OrderLine(sku, price));
+    }
+
+    public void submit() { this.status = Status.SUBMITTED; }
+
+    public int lineCount() { return lines.size(); }
+
+    public List<OrderLine> getLines() {
+        return Collections.unmodifiableList(lines); // defensive copy — callers cannot mutate
+    }
+
+    public String getId() { return id; }
+}
+```
+
+```java
+// file: OrderRepository.java
+import java.util.Optional;
+
+public interface OrderRepository {
+    Optional<Order> find(String id);
+    void save(Order order);
+}
+```
+
+```java
+// file: Main.java
+public class Main {
+    public static void main(String[] args) {
+        Order order = new Order("ord-1");
+        order.addLine("SKU-1", new Money(1299, "USD"));
+        order.submit();
+        try {
+            order.addLine("SKU-2", new Money(500, "USD"));
+        } catch (IllegalStateException e) {
+            System.out.println("rejected: " + e.getMessage());
+            // rejected: cannot add lines to a submitted order
+        }
+    }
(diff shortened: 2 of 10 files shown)


skills/principle-ddd/examples/order.rs.md
```diff
@@ -0,0 +1,97 @@
...
+    pub fn line_count(&self) -> usize { self.lines.len() }
+}
+```
+
+```rust
+// file: repository.rs  (domain-layer port — no implementation)
+use crate::order::Order;
+
+pub trait OrderRepository {
+    fn find(&self, id: &str) -> Option<Order>;
```
lugassawan, line 69: **High** — `fn find(&self, id: &str) -> Option<Order>` returns an owned `Order`, but `Order` does not derive `Clone`. Any concrete implementor must construct a fresh `Order` from scratch each call — it cannot return a stored instance. A reader copying this snippet into a real project will get a compile error.

**Suggested fix:** Add `#[derive(Clone)]` to `Order`, `OrderLine`, and `Status`:

```rust
#[derive(Debug, Clone, PartialEq)]
enum Status { Draft, Submitted }
#[derive(Debug, Clone)]
pub struct OrderLine { ... }
#[derive(Debug, Clone)]
pub struct Order { ... }
```

Or change the trait signature to `fn find<'a>(&'a self, id: &str) -> Option<&'a Order>`.

skills/principle-ddd/examples/order.rs.md
```diff
@@ -0,0 +1,97 @@
...
+        Ok(Money::new(self.minor_units + other.minor_units, &self.currency))
+    }
+}
+```
+
+```rust
+// file: order.rs
+use crate::money::Money;
+
+#[derive(PartialEq)]
```
lugassawan: **Medium** — `OrderLine.sku` and `OrderLine.price` are `pub`, which is at odds with the aggregate-root-as-sole-entry-point teaching. Also `enum Status` is missing `#[derive(Debug)]`; adding `#[derive(Debug)]` to `Order` later will cause a compile error.

**Suggested fix:** Make `OrderLine` fields private (with a constructor) or annotate them as illustrative-only. Add `#[derive(Debug)]` to `Status`.

skills/principle-ddd/examples/order.go.md
```diff
@@ -0,0 +1,120 @@
...
+unexported fields, value receivers, and the absence of setters. Callers cannot
+mutate `Money` directly; they go through constructors or methods that return new
+values. The `Order` aggregate root hides `lines []orderLine` behind an unexported
+field so only `AddLine` and `Submit` can mutate it, keeping the draft-only
+invariant enforceable. `AddLine` returns `error` (Go's idiomatic flow — no
+exceptions) and `OrderRepository` is a plain `interface` — a domain port with
+zero infrastructure coupling.
+
+## Implementation
+```go
```
lugassawan, line 16: **Medium** — Missing blank line between the `## Implementation` heading and the first ` ```go ` fence. Every other file in this PR includes this separator; its absence here is a raw-file inconsistency that future authors using these as templates will notice.

**Suggested fix:** Add one blank line after `## Implementation` before the opening fence.

skills/principle-ddd/examples/order.swift.md
```diff
@@ -0,0 +1,94 @@
...
+enum OrderError: Error {
+    case submittedOrderCannotBeModified
+}
+
+struct OrderLine {
+    let sku: String
+    let price: Money
+}
+
+private enum Status { case draft, submitted }
```
lugassawan: **Medium** — `private enum Status` is declared at file scope rather than nested inside `Order`. Functionally correct, but the idiomatic Swift placement for an aggregate-root implementation detail is as a nested type:

```swift
class Order {
    private enum Status { case draft, submitted }
    // ...
}
```

This reinforces the DDD lesson that `Status` is an aggregate-root detail, not a stand-alone domain concept.

skills/principle-ddd/examples/order.rb.md
```diff
@@ -0,0 +1,106 @@
...
+end
+```
+
+```ruby
+# file: order.rb
+require_relative 'money'
+
+class OrderError < StandardError; end
+
+OrderLine = Struct.new(:sku, :price)
```
lugassawan: **Low** — `OrderLine = Struct.new(:sku, :price)` creates a mutable value — callers can write `line.sku = "tampered"` after construction. `Money` correctly calls `freeze`, but `OrderLine` does not.

**Suggested fix:**
```ruby
OrderLine = Struct.new(:sku, :price) { def initialize(...) = super.freeze }
# or Ruby 3.2+:
OrderLine = Data.define(:sku, :price)
```

skills/principle-ddd/examples/order.py.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+@dataclass(frozen=True)
+class OrderLine:
+    sku: str
+    price: Money
+
+
+class Order:
+    def __init__(self, id: str) -> None:
```
lugassawan, line 50: **Low** — Using `id` as both the parameter name and instance attribute shadows the Python built-in `id()`. The code runs correctly, but readers may copy this pattern uncritically.

**Suggested fix:** Rename to `order_id` throughout, or add an inline comment acknowledging the intentional shadowing if consistency with the other 8 files is the priority.

skills/principle-ddd/examples/order.rs.md
lugassawan: Addressed in a60438dcadc6a69038e8cc1a824b524b994c1c2c: added `#[derive(Debug, Clone, PartialEq)]` to `Status` and `#[derive(Debug, Clone)]` to `OrderLine` and `Order` — the repository trait now has a compilable owned-value signature.

skills/principle-ddd/examples/order.rs.md
lugassawan: Addressed in a60438dcadc6a69038e8cc1a824b524b994c1c2c: added `#[derive(Debug, Clone)]` to `Status` and `OrderLine`; `OrderLine` fields remain `pub` (readable after retrieval is intentional in this teaching example).

skills/principle-ddd/examples/order.go.md
lugassawan: Addressed in a60438dcadc6a69038e8cc1a824b524b994c1c2c: added blank line between `## Implementation` heading and first fence — now consistent with all 8 other files in the PR.

skills/principle-ddd/examples/order.swift.md
lugassawan: Addressed in a60438dcadc6a69038e8cc1a824b524b994c1c2c: moved `private enum Status` inside `Order` class body — signals it as an aggregate-root implementation detail, not a stand-alone domain concept.

skills/principle-ddd/examples/order.rb.md
lugassawan: Addressed in a60438dcadc6a69038e8cc1a824b524b994c1c2c: added `def initialize(sku, price) = super.freeze` to `OrderLine`'s struct block — consistent with `Money`'s `freeze` call, ensuring value-object immutability throughout.

skills/principle-ddd/examples/order.py.md
lugassawan: Keeping `id` for consistency with the other 8 language files in this PR — all use `id` as the attribute name so examples read uniformly across languages. This is an intentional consistency trade-off; noted as a learning point for production code.

skills/principle-ddd/examples/order.java.md
```diff
@@ -0,0 +1,99 @@
...
+        }
+        lines.add(new OrderLine(sku, price));
+    }
+
+    public void submit() { this.status = Status.SUBMITTED; }
+
+    public int lineCount() { return lines.size(); }
+
+    public List<OrderLine> getLines() {
+        return Collections.unmodifiableList(lines); // defensive copy — callers cannot mutate
```
lugassawan: **Low** — `Collections.unmodifiableList(lines)` returns a **read-only view** of the backing list, not a defensive copy. The comment "defensive copy — callers cannot mutate" is misleading: it correctly describes the access control (callers cannot mutate via this reference), but misrepresents the mechanism. Readers who copy this pattern and later expose the backing list will find the "copy" reflects live changes.

**Suggested fix:**
```java
return Collections.unmodifiableList(lines); // read-only view — callers cannot mutate via this reference
```

skills/principle-ddd/examples/order.ts.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+```typescript
+// file: order.ts
+import { Money } from "./money";
+
+type Status = "draft" | "submitted";
+
+export interface OrderLine { sku: string; price: Money }
```
lugassawan: **Low** — `export interface OrderLine { sku: string; price: Money }` — fields are mutable by default in TypeScript interfaces. Every other language in this PR uses `val`, `readonly`, `let`, `freeze`, or `frozen=True` for line fields; this is the only exception. While the aggregate-root invariant is not broken (callers still cannot add lines after submit), the missing `readonly` is a teaching inconsistency.

**Suggested fix:**
```typescript
export interface OrderLine { readonly sku: string; readonly price: Money }
```

skills/principle-ddd/examples/order.java.md
lugassawan: Addressed in 28ab929: corrected comment from 'defensive copy' to 'read-only view' to accurately describe `Collections.unmodifiableList` semantics.

skills/principle-ddd/examples/order.ts.md
lugassawan: Addressed in 28ab929: added `readonly` to both `OrderLine` fields, consistent with all 8 other language files.

PR #445, opened 2026-06-12
[feat] Enhance principle-resiliency: idempotency & rate-limiting (#356)

### LINKED ISSUE #356: [feat] Enhance principle-resiliency: add idempotency & rate-limiting sections
## Product framing

- **Problem.** `principle-resiliency` covers circuit breakers, retries, and bulkheads but lacks idempotency (safe retry semantics, idempotency keys) and rate-limiting (throttling, quotas, backpressure). These are the two most common reliability gaps in service-to-service calls.
- **Value.** Engineers building APIs or consuming third-party services get explicit guidance on making operations safely retryable and on protecting services under load.
- **Acceptance criteria.** `principle-resiliency` gains idempotency and rate-limiting sections; parent stays ≤150 lines; `scripts/validate.py` passes.
- **Impact / Effort.** Impact: M — idempotency bugs cause duplicate charges and double-sends; rate-limit failures cause cascading overload. Effort: S — two focused sections.

## Goal
Enhance `principle-resiliency` with idempotency (keys, safe-retry semantics) and rate-limiting (throttling, quotas, backpressure) sections.

## Motivation
Circuit breakers and retries are already covered, but retry-safety (idempotency) is the complementary missing piece — retrying a non-idempotent operation is worse than not retrying at all. Rate-limiting is the producer-side complement to circuit-breaking.

## Surface
- [x] Skill — `principle-*`

## Sketch (optional)
Add to `skills/principle-resiliency/SKILL.md`:
- **Idempotency section:** idempotency keys, safe vs unsafe HTTP methods, at-least-once vs exactly-once semantics, dedup stores.
- **Rate-limiting section:** token bucket/leaky bucket algorithms, quota headers, client-side backpressure, jitter in retry delays.

## Acceptance (optional)
- [ ] Idempotency section added (keys, safe-retry semantics, dedup)
- [ ] Rate-limiting section added (algorithms, backpressure, jitter)
- [ ] `skills/principle-resiliency/SKILL.md` stays ≤150 lines
- [ ] `scripts/validate.py` passes

lugassawan:
## Summary

- Add `## Idempotency` section to `principle-resiliency` SKILL.md: idempotency keys, safe HTTP methods, at-least-once vs exactly-once, dedup store with `pending`/`completed` states, and the ordering trap (reserve key *before* side effect)
- Add `## Rate Limiting` section: token bucket vs leaky bucket vs fixed/sliding window, quota headers (`RateLimit-*`, `Retry-After`), client-side backpressure, and jitter to avoid thundering herd
- Add 9 idempotency example files (`examples/idempotency.{cs,go,java,kt,py,rb,rs,swift,ts}.md`) — dedup wrapper around a non-idempotent charge with error-path key cleanup
- Add 9 rate-limit example files (`examples/rate-limit.{cs,go,java,kt,py,rb,rs,swift,ts}.md`) — token-bucket limiter with jitter; common mistake: fixed-window boundary burst
- Update frontmatter keywords and auto-load triggers; add 4 new Red Flags rows; extend `triggers.txt` with 4 idempotency/rate-limiting prompts
- Add `principle-resiliency, principle-distributed-systems` sibling-set entry (intentional idempotency overlap)

## Test Plan

- [x] `bash scripts/validate.sh` — all checks pass (SKILL.md 121 lines < 150; all 18 examples ≤ 112 lines < 120)
- [x] Python idempotency: same-key retry returns cached result, charge called exactly once; error-path key cleanup releases PENDING correctly
- [x] TypeScript rate-limit: 3-token burst acquired, 4th rejected, refill allows re-acquire
- [x] Trigger ranking tests: `python -m pytest tests/test_skill_triggers.py -k principle-resiliency` — 9 passed

Closes #356


skills/principle-resiliency/examples/idempotency.rs.md
```diff
@@ -0,0 +1,88 @@
...
+                    // Reserve BEFORE executing — concurrent retry sees Pending and stops.
+                    guard.insert(key.to_string(), Entry::Pending);
+                }
+            }
+        }
+
+        let result = operation();
+        self.store.lock().unwrap().insert(key.to_string(), Entry::Completed(result.clone()));
+        Ok(result)
+    }
```
lugassawan, line 58: **High** — Rust idempotency `execute` has no error-path cleanup: if `operation()` panics, the key remains `Pending` forever, permanently blocking retries with `already in-flight`. All 8 other language examples delete the key on failure. The closure bound `FnOnce() -> T` (infallible) prevents adding a `Result` branch without changing the signature.

**Suggested fix:** Change the closure bound to `FnOnce() -> Result<T, String>` and add a cleanup arm:

```rust
pub fn execute<F, E>(&self, key: &str, operation: F) -> Result<T, E>
where
    F: FnOnce() -> Result<T, E>,
{
    // ... reserve key as Pending ...
    match operation() {
        Ok(result) => {
            self.store.lock().unwrap().insert(key.to_string(), Entry::Completed(result.clone()));
            Ok(result)
        }
        Err(e) => {
            self.store.lock().unwrap().remove(key); // release — allows retry
            Err(e)
        }
    }
}
```

skills/principle-resiliency/examples/rate-limit.swift.md
```diff
@@ -0,0 +1,99 @@
...
+func callWithRateLimit<T>(
+    _ op: () async -> T,
+    maxAttempts: Int = 5
+) async throws -> T {
+    for attempt in 0..<maxAttempts {
+        if await bucket.tryAcquire() { return await op() }
+        // Jitter prevents synchronized retries (thundering herd).
+        let baseMs = UInt64(1 << attempt)
+        let jitterMs = baseMs / 2 + UInt64.random(in: 0...(baseMs / 2))
+        try await Task.sleep(nanoseconds: jitterMs * 1_000_000)
```
lugassawan, line 58: **High** — Zero-ms jitter on first retry: `baseMs = 1`, then `1 / 2 = 0` (integer division), so `jitterMs = 0` guaranteed. `Task.sleep(nanoseconds: 0)` is effectively a no-op. An example that teaches 'jitter prevents thundering herd' but sleeps for zero on the most common case (first retry) actively mis-teaches the reader.

**Suggested fix:**
```swift
let baseMs = max(1, UInt64(1 << attempt))
let jitterMs = baseMs / 2 + UInt64.random(in: 0...(baseMs / 2))
try await Task.sleep(nanoseconds: jitterMs * 1_000_000)
```

skills/principle-resiliency/examples/rate-limit.rs.md
```diff
@@ -0,0 +1,102 @@
...
+    F: Fn() -> T,
+{
+    for attempt in 0..max_attempts {
+        if bucket.try_acquire() {
+            return Some(op());
+        }
+        // Jitter prevents synchronized retries (thundering herd).
+        let base_ms = (1u64 << attempt).min(1000);
+        let jitter_ms = base_ms / 2 + (rand_jitter() % (base_ms / 2 + 1));
+        std::thread::sleep(Duration::from_millis(jitter_ms));
```
lugassawan, line 64: **High** — Zero-ms jitter on first retry: `base_ms = 1`, `base_ms / 2 = 0` (integer division), `rand_jitter() % (0 + 1) = 0`. `Duration::from_millis(0)` is always zero on attempt 0. An example teaching jitter-prevents-thundering-herd that produces guaranteed no-sleep on the first attempt actively mis-teaches.

**Suggested fix:** Use `2u64 << attempt` so the minimum base is 2, ensuring a non-zero divisor:
```rust
let base_ms = (2u64 << attempt).min(1000);
let jitter_ms = base_ms / 2 + (rand_jitter() % (base_ms / 2 + 1));
std::thread::sleep(Duration::from_millis(jitter_ms));
```

skills/principle-resiliency/examples/rate-limit.java.md
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.Callable;
+
+public class RateLimitedCaller {
+    private static final Random RNG = new Random();
+
+    public static <T> T call(TokenBucket bucket, Callable<T> op, int maxAttempts) throws Exception {
+        for (int attempt = 0; attempt < maxAttempts; attempt++) {
+            if (bucket.tryAcquire()) return op.call();
+            // Jitter prevents synchronized retries (thundering herd).
+            long backoff = (long) ((1 << attempt) * (0.5 + RNG.nextDouble()));
```
lugassawan: **Medium** — `(long) ((1 << attempt) * (0.5 + RNG.nextDouble()))` on attempt 0 produces a double in `[0.5, 1.5)`. Truncation via `(long)` yields 0 roughly 50% of the time (`0.5 ≤ x < 1.0`). `Thread.sleep(0)` is a yield with no delay. The same truncation pattern affects `rate-limit.kt.md` (via `.toLong()`) and `rate-limit.cs.md` (via `(int)`).

**Suggested fix:** Add a minimum floor of 1 ms:
```java
long backoff = Math.max(1L, (long) ((1 << attempt) * (0.5 + RNG.nextDouble())));
Thread.sleep(backoff);
```
Apply equivalently in Kotlin (`max(1L, ...)`) and C# (`Math.Max(1, ...)`).

skills/principle-resiliency/SKILL.md
```diff
@@ -69,6 +69,33 @@ Cascading-check trap: if every instance polls every dependency on each health in
...
+
+> See `examples/` for a worked idempotency-key dedup store — prevents double-charge on concurrent retries (read on demand — not auto-loaded).
+
+## Rate Limiting
+
+Rate limiting controls request volume to protect producers from overload and consumers from accidental burst.
+
+- **Token bucket** — bucket holds up to N tokens; each request consumes one; tokens refill at a steady rate. Allows short bursts up to bucket capacity, then enforces the steady-state rate.
+- **Leaky bucket** — requests enter a queue and drain at a fixed rate regardless of burst, producing constant-rate output at the cost of higher latency.
+- **Fixed/sliding window** — simpler to implement; fixed window has a boundary-burst caveat: a client can send up to 2× the limit across a window boundary.
```
lugassawan: **Medium** — The bullet lists token bucket, leaky bucket, and fixed/sliding window but the sliding window's distinguishing advantage (eliminates the boundary-burst problem by counting requests over the preceding N seconds rather than resetting at a fixed point) goes unstated. A reader choosing between algorithms has no signal for when to prefer sliding over fixed.

**Suggested addition:** Extend the bullet: "…fixed window has a boundary-burst caveat: up to 2× the limit is possible across a window boundary. A sliding window counts requests over the preceding N seconds, eliminating the spike at the cost of per-request timestamp storage."

skills/principle-resiliency/examples/rate-limit.ts.md
```diff
@@ -0,0 +1,89 @@
...
+import { TokenBucket, callWithRateLimit } from "./rateLimiter";
+
+// 3-token burst, refills 1 token/second
+const limiter = new TokenBucket(3, 1 / 1000);
+
+for (let i = 0; i < 5; i++) {
+  callWithRateLimit(limiter, async () => `ok-${i}`)
+    .then((r) => console.log(r))
+    .catch(() => console.log("rejected"));
+}
```
lugassawan, line 69: **Low** — All 5 `callWithRateLimit` calls are fired concurrently (no `await` in the loop), making the demo output non-deterministic. The comment "first 3 succeed immediately; later calls depend on refill timing" is true but vague. Sequential `await` calls would produce a deterministic, readable demo that still exercises the limiter correctly.

```typescript
for (let i = 0; i < 5; i++) {
  try {
    const r = await callWithRateLimit(limiter, async () => `ok-${i}`);
    console.log(r);
  } catch { console.log("rejected"); }
}
```

tests/skill_sibling_sets.txt
```diff
@@ -54,3 +54,12 @@ language-python, language-rust
...
 principle-performance, workflow-performance-investigation
+
+# principle-resiliency and principle-distributed-systems share idempotency trigger
+# surface: principle-distributed-systems covers at-least-once delivery and idempotency
+# across the network layer, while principle-resiliency covers the dedup-store
+# implementation that makes retries safe. A prompt like "my payment service double-
+# charges on retries — how do I implement exactly-once semantics with idempotency"
+# legitimately belongs to either skill — the delivery-layer vs dedup-store distinction
+# cannot be resolved by BM25 alone.
+principle-resiliency, principle-distributed-systems
```
lugassawan, line 65: **Low** — The sibling-set comment says "principle-distributed-systems covers at-least-once delivery and idempotency across the network layer" — the exact H2 section heading in `skills/principle-distributed-systems/SKILL.md` is `## Idempotency Across the Network`. Consider quoting the heading verbatim (backtick-fenced) so future maintainers can `grep` for it:

```
# surface: principle-distributed-systems covers at-least-once delivery via
# `## Idempotency Across the Network`; principle-resiliency covers the dedup-store
```

This keeps the comment self-documenting and grep-linkable to the actual section.

skills/principle-resiliency/examples/idempotency.rs.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: changed closure bound from `FnOnce() -> T` to `FnOnce() -> Result<T, String>`; error-path cleanup now removes the Pending key on failure.

skills/principle-resiliency/examples/rate-limit.swift.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: changed to `max(2, UInt64(1 << attempt))` so `baseMs / 2 >= 1` on attempt 0, guaranteeing at least 1 ms jitter.

skills/principle-resiliency/examples/rate-limit.rs.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: changed to `2u64 << attempt` so `base_ms / 2 >= 1` on attempt 0, guaranteeing at least 1 ms jitter.

skills/principle-resiliency/examples/rate-limit.java.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: added `Math.max(1L, (long)(...))` floor in Java, `maxOf(1L, ...)` in Kotlin, and `Math.Max(1, (int)(...))` in C# — ensures at least 1 ms delay regardless of truncation on attempt 0.

skills/principle-resiliency/SKILL.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: extended the fixed/sliding window bullet to explain that sliding window tracks per-slot counts and weights across the boundary, avoiding the 2× burst at the cost of more state per client.

skills/principle-resiliency/examples/rate-limit.ts.md
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: replaced fire-and-forget .then()/.catch() calls with a sequential `for...await` loop for deterministic output.

tests/skill_sibling_sets.txt
lugassawan: Addressed in 07f5871000f3ae79e951df9a08a3e6d8c1a0bb42: updated comment to quote `## Idempotency Across the Network` verbatim for grep-ability.

skills/principle-resiliency/examples/rate-limit.ts.md
```diff
@@ -0,0 +1,92 @@
...
+
+export async function callWithRateLimit<T>(
+  limiter: TokenBucket,
+  operation: () => Promise<T>,
+  maxAttempts = 5,
+): Promise<T> {
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
+    if (limiter.tryAcquire()) return operation();
+    // Jitter prevents synchronized retries (thundering herd).
+    const backoff = 2 ** attempt * (0.5 + Math.random());
```
lugassawan: **Critical** — `callWithRateLimit` backoff on attempt 0 produces a sub-ms delay: `2 ** 0 * (0.5 + Math.random())` is in `[0.5, 1.5)`. `setTimeout` rounds sub-ms values to 0 ms in browsers and Node.js — effectively a zero-delay retry, which defeats the jitter entirely. This is the same class of bug that was fixed in Java, Kotlin, C#, Rust, and Swift in this PR.

Suggested fix (floor the base, matching the pattern used in other languages):
```typescript
const backoff = Math.max(1, 2 ** attempt) * (0.5 + Math.random());
```
This ensures attempt 0 sleeps `[0.5, 1.5)` ms minimum and scales correctly from there.

lugassawan: Addressed in d70f6f9c91c9e3a40de8c9794cfbc693d1ab12e7: added `Math.max(1, 2 ** attempt)` floor so attempt 0 sleeps at least [0.5, 1.5) ms — same floor pattern applied to Java/Kotlin/C#/Rust/Swift in the prior commit.

PR #446, opened 2026-06-12
[feat] Enhance principle-performance: add a caching section (#357)

### LINKED ISSUE #357: [feat] Enhance principle-performance: add a caching section
## Product framing

- **Problem.** `principle-performance` covers profiling-first and algorithmic complexity but lacks a caching section — cache layering (L1/L2/CDN), invalidation strategies, TTL choices, and cache-stampede prevention are among the highest-impact performance tools.
- **Value.** Engineers working on read-heavy systems get explicit guidance on cache topology, invalidation tradeoffs, and stampede protection without reaching for a separate observability skill.
- **Acceptance criteria.** `principle-performance` gains a caching section; parent stays ≤150 lines; `scripts/validate.py` passes.
- **Impact / Effort.** Impact: M — caching is the most common way to achieve a 10× read-path improvement. Effort: S — one focused section.

## Goal
Enhance `principle-performance` with a caching section covering layering, invalidation, TTL, and stampede prevention.

## Motivation
Caching is the highest-leverage performance tool in most systems, yet it is absent from the current skill. Cache invalidation ("the two hard things") and stampede on cold start are real failure modes that benefit from explicit guidance.

## Surface
- [x] Skill — `principle-*`

## Sketch (optional)
Add to `skills/principle-performance/SKILL.md`:
- **Caching section:** cache layers (in-process, distributed, CDN), write-through vs write-behind vs cache-aside, TTL selection, stampede prevention (probabilistic early expiry, mutex/lock), invalidation patterns.

## Acceptance (optional)
- [ ] Caching section added (layers, invalidation strategies, TTL, stampede prevention)
- [ ] `skills/principle-performance/SKILL.md` stays ≤150 lines
- [ ] `scripts/validate.py` passes

lugassawan:
## Summary
- Add `## Caching` section to `skills/principle-performance/SKILL.md` covering layers (in-process/L1, distributed/L2, CDN/edge), invalidation strategies (cache-aside, write-through, write-behind), TTL sizing, and stampede prevention with single-flight — reconciled with the existing "We'll cache it later" skeptical framing
- Add 3 new Red Flags rows: cache invalidation as afterthought, unbounded cache growth, no stampede guard
- Add `skills/principle-performance/examples/cache-aside.{cs,go,java,kt,py,rb,rs,swift,ts}.md` — 9 worked examples showing cache-aside read-through with single-flight/mutex guard (thundering-herd prevention), with a Common Mistake section in each
- Append 2 caching trigger prompts to `triggers.txt`; enrich frontmatter description with discriminating caching terms

## Test Plan

- [ ] `bash scripts/validate.sh` passes — SKILL.md ≤150 lines, each example ≤120 lines, frontmatter `name: principle-performance` intact, triggers ≥2 non-comment lines
- [ ] `python -m pytest tests/test_skill_triggers.py -k "principle-performance"` — all 7 trigger prompts rank #1
- [ ] Python `cache-aside.py.md` snippet exercised manually: cache hit returns cached value; 3 concurrent cold misses call loader exactly once (single-flight confirmed)
- [ ] TypeScript `cache-aside.ts.md` snippet exercised manually: same hit + single-flight verification
- [ ] Concurrency correctness review addressed: Kotlin uses `computeIfAbsent` (not `getOrPut`), Ruby guards `@locks` hash under `@global`, Java uses `compute` (atomic), Swift doc comment corrected to actor isolation

Closes #357


skills/principle-performance/examples/cache-aside.java.md
```diff
@@ -0,0 +1,63 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.*;
+import java.util.function.Function;
+
+public class CacheAside<V> {
+    private record Entry<V>(V value, Instant expiresAt) {}
+
+    private final ConcurrentHashMap<String, Entry<V>> store = new ConcurrentHashMap<>();
+    // computeIfAbsent on ConcurrentHashMap serialises concurrent calls for the same key,
```
lugassawan: **Medium:** The comment above references `computeIfAbsent` but the implementation uses `compute`. These have meaningfully different semantics: `computeIfAbsent` skips the lambda when the key is present; `compute` always runs it — which is exactly why `compute` was chosen here (to handle stale-entry replacement atomically). A reader looking up the wrong API will build an incorrect mental model.

Suggested fix:
```
// compute on ConcurrentHashMap serialises concurrent calls for the same key,
// providing single-flight semantics without an explicit per-key lock map.
```

skills/principle-performance/examples/cache-aside.rb.md
```diff
@@ -0,0 +1,59 @@
...
+caller recomputes while all others wait and share the result. The cache stores the value with a TTL;
+on a hit within the TTL, the value is returned immediately without touching the origin.
+
+## Implementation
+
+```ruby
+# file: cache_aside.rb
+require 'monitor'
+
+# Cache-aside with per-key MonitorMixin for single-flight on cold/expired keys.
```
lugassawan: **Medium:** The comment says "per-key MonitorMixin" but the implementation uses `Monitor` (a standalone class). `MonitorMixin` is a module you `include` into a class; `Monitor` is a standalone object you instantiate. Using the wrong term in a teaching example sends readers to the wrong API.

Suggested fix:
```
# Cache-aside with per-key Monitor for single-flight on cold/expired keys.
```

skills/principle-performance/SKILL.md
```diff
@@ -58,6 +58,17 @@ One query per iteration is the most common silent performance killer.
 - Measure query count at the boundary (query logs, explain plans) — not only wall-clock time.
 - "We'll cache it later" is not a query strategy; the cache miss path is still N+1.
 
+## Caching
+
+Caching trades freshness and memory for latency — earn it with a profile and a clear invalidation plan, not a hunch.
+- Add a cache only after measuring; a cache that hides an N+1 query still runs N+1 queries on a miss (see `N+1 and the Database Boundary` above).
+- Layers — in-process (L1, fastest, per-node, risks drift between replicas) → distributed (L2, shared, costs a network hop) → CDN/edge (static content and cacheable GETs). See `principle-api-design#REST vs RPC vs Event-Driven` for HTTP cacheability.
+- Invalidation strategies — cache-aside (lazy, app-managed, the default: read from cache; on miss, load from origin, populate, return) vs write-through (write populates cache synchronously, strong consistency, higher write cost) vs write-behind (async flush, fast writes, loss window on crash). Prefer TTL + explicit bust on known writes; see `principle-event-driven#Event Sourcing` for cache/projection drift.
```
lugassawan: **Low:** The write-behind bullet names the downside ("loss window on crash") but gives no pointer on how to bound it. Consider adding a brief note: "ensure a WAL or durable queue backs the async flush to bound the loss window" — or a cross-ref to `principle-resiliency` where durability patterns are covered.

skills/principle-performance/examples/cache-aside.rs.md
```diff
@@ -0,0 +1,71 @@
...
+    pub fn new(ttl: Duration) -> Arc<Self> {
+        Arc::new(Self {
+            store: Mutex::new(HashMap::new()),
+            ttl,
+        })
+    }
+
+    /// Returns the cached value, or calls `loader` under the lock (single-flight).
+    pub fn get_or_load(&self, key: K, loader: impl FnOnce(&K) -> V) -> V {
+        let mut store = self.store.lock().unwrap();
```
lugassawan, line 47: **Low:** `.unwrap()` on `Mutex::lock()` panics if the mutex is poisoned (a thread panicked while holding it). Since this example holds the lock while calling a user-supplied `loader` (which can panic), this is a realistic scenario. Consider a brief inline comment explaining the panic path, or use `.unwrap_or_else(|p| p.into_inner())` to recover from a poisoned mutex gracefully.

lugassawan:
Addressed all 4 review findings in 71b6093:
- `cache-aside.java.md:23` — fixed stale comment (`computeIfAbsent` → `compute`)
- `cache-aside.rb.md:16` — fixed comment (`MonitorMixin` → `Monitor`)
- `SKILL.md:66` — added WAL/durable-queue note to write-behind bullet
- `cache-aside.rs.md:45` — documented `.unwrap()` poison-mutex panic path

skills/principle-performance/examples/cache-aside.rs.md
```diff
@@ -0,0 +1,73 @@
...
+
+struct Entry<V> {
+    value: V,
+    expires_at: Instant,
+}
+
+pub struct CacheAside<K, V> {
+    // A single Mutex over the whole store is simple and correct for moderate contention.
+    // Trade-off: the lock is held while the loader runs, so slow loaders (e.g. HTTP calls)
+    // block all concurrent reads for any key. For high-throughput use, prefer a per-key
```
lugassawan, line 29: **Medium:** The trade-off comment (lines 27–31) warns that holding the global lock across `loader` serialises all concurrent reads for the duration of every call. The existing `## Common Mistake` section shows only the lack of single-flight, not this held-lock-across-IO problem. A reader copying this into a networked context will create a convoy without any visual warning.

Consider adding a second anti-pattern block or a note: "Do not use with async/networked loaders — the held lock will block all concurrent reads for the duration of every origin call. Prefer a per-key lock map (e.g. `DashMap`) for IO-bound loaders."

skills/principle-performance/examples/cache-aside.cs.md
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public async Task<T> GetAsync(string key)
+    {
+        if (_store.TryGetValue(key, out var entry) && entry.ExpiresAt > DateTime.UtcNow)
+            return entry.Value;
+
+        // Single-flight: acquire a per-key lock so only one caller recomputes.
+        // GetOrAdd may invoke the factory on multiple threads; only one SemaphoreSlim wins.
+        // Extra instances are discarded (not Disposed) — acceptable here, use a pool in production.
```
lugassawan: **Low:** The production guidance says "use a pool in production" which gives no concrete direction. A reader will be unsure whether to use `ConcurrentDictionary`-with-cleanup or a different pattern.

Consider replacing with: "In production, remove stale entries from `_locks` after `Release()`, or replace the per-key `SemaphoreSlim` map with a channel-based or `Lazy<SemaphoreSlim>` approach."

skills/principle-performance/examples/cache-aside.py.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+class CacheAside:
+    """Cache-aside with per-key asyncio.Lock for single-flight on cold/expired keys."""
+
+    def __init__(self, ttl: float, loader: Callable[[str], Awaitable[object]]) -> None:
+        self._ttl = ttl
+        self._loader = loader
+        self._store: dict[str, tuple[object, float]] = {}
+        self._locks: defaultdict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
```
lugassawan, line 31: **Low:** `defaultdict(asyncio.Lock)` is only safe in a single-threaded asyncio context. If this class is used alongside `run_in_executor` or `ThreadPoolExecutor`, two threads can each trigger `__missing__` for the same key and get different `Lock` instances, breaking single-flight.

Consider adding a comment above the `_locks` field:
```python
# Only safe in a single-threaded asyncio event loop; use threading.Lock + a plain dict if combining with threads.
```

skills/principle-performance/examples/cache-aside.kt.md
```diff
@@ -0,0 +1,57 @@
...
+
+private data class Entry<V>(val value: V, val expiresAt: Instant)
+
+class CacheAside<V>(
+    private val ttl: Duration,
+    private val loader: suspend (String) -> V,
+) {
+    private val store = ConcurrentHashMap<String, Entry<V>>()
+    // Per-key Mutex: only one coroutine recomputes a cold/expired entry.
+    // computeIfAbsent is atomic per-key; getOrPut is NOT — two coroutines could get different instances.
```
lugassawan, line 28: **Low:** The comment correctly notes `computeIfAbsent` is atomic, but doesn't mention that `kotlinx.coroutines.sync.Mutex` is non-reentrant. A loader that calls `get()` for the same key will deadlock.

Consider adding a note below line 28:
```kotlin
// Non-reentrant: a loader that calls get() for the same key will deadlock.
```

lugassawan:
Addressed round-2 review feedback in d19e3ef:

- **`cache-aside.rs.md`** — Added a second Common Mistake: "Global lock held across a slow loader blocks concurrent reads for *any* key." Points to DashMap for per-key isolation. All 4 threads resolved.
- **`cache-aside.cs.md`** — Replaced vague "use a pool in production" with actionable guidance: call `_locks.TryRemove(key, out _)` after `sem.Release()` once the entry is populated, preventing `SemaphoreSlim` accumulation on high-cardinality keys.
- **`cache-aside.py.md`** — Added comment on `_locks` explaining that `defaultdict(asyncio.Lock)` is safe only in a single-threaded event loop; concurrent threads require a `threading.Lock` guard on creation.
- **`cache-aside.kt.md`** — Added note that `kotlinx.coroutines.sync.Mutex` is NOT reentrant — calling `withLock` inside `withLock` on the same mutex deadlocks.

skills/principle-performance/examples/cache-aside.go.md
```diff
@@ -0,0 +1,96 @@
...
+		cl.wg.Wait()
+		return cl.val, cl.err
+	}
+	cl := &call[V]{}
+	cl.wg.Add(1)
+	c.inflight[key] = cl
+	c.mu.Unlock()
+
+	cl.val, cl.err = c.loader(key)
+	cl.wg.Done()
```
lugassawan: **High:** If `c.loader(key)` panics (not returns an error), `cl.wg.Done()` at line 69 is never called — every goroutine blocked on `cl.wg.Wait()` at line 60 leaks permanently. The successful-error path is handled correctly (`defer` is absent only for the panic path).

Protect with a deferred Done:
```go
cl.wg.Add(1)
c.inflight[key] = cl
c.mu.Unlock()

func() {
    defer cl.wg.Done()
    cl.val, cl.err = c.loader(key)
}()
```
Or add a comment: "This example assumes non-panicking loaders; add `recover` in production."

skills/principle-performance/examples/cache-aside.swift.md
```diff
@@ -0,0 +1,68 @@
...
+
+    func get(_ key: String) async throws -> V {
+        if let entry = store[key], Date.now < entry.expiresAt {
+            return entry.value
+        }
+        // Single-flight via actor isolation: reuse an in-progress Task for the same key.
+        if let task = inflight[key] {
+            return try await task.value
+        }
+        let task = Task { try await self.loader(key) }
```
lugassawan, line 46: **High:** `Task { try await self.loader(key) }` is an unstructured task not tied to the caller's cancellation scope. If the calling task is cancelled before awaiting `task.value`, the loader continues running and writes its result to the actor store. The doc comment says "the loader runs exactly once" (correct) but omits the cancellation gap — a common Swift concurrency footgun.

Add a note:
```swift
// Note: cancelling the caller's task does not cancel this shared loader Task —
// it runs to completion and populates the cache regardless.
```
For strict structured concurrency, use `withTaskCancellationHandler` or a `TaskGroup`.

skills/principle-performance/examples/cache-aside.java.md
```diff
@@ -0,0 +1,63 @@
...
+
+    public V get(String key) {
+        Entry<V> entry = store.get(key);
+        if (entry != null && Instant.now().isBefore(entry.expiresAt())) {
+            return entry.value();
+        }
+        // compute is atomic per-key (bin-level lock in ConcurrentHashMap since Java 8):
+        // eviction of a stale entry and recomputation happen as a single step, so only
+        // one thread calls loader for a given key at a time.
+        return store.compute(key, (k, existing) -> {
```
lugassawan, line 44: **Medium:** `ConcurrentHashMap.compute` holds a bin-level lock for the entire duration of `loader.apply(k)`. The existing comment says "one thread calls loader for a given key at a time" (correct), but omits that threads computing on *other keys* in the same hash bin are also blocked.

For blocking I/O loaders with high key cardinality, consider a note:
```java
// Note: compute holds the bin lock while loader runs. For blocking I/O loaders,
// concurrent compute calls on keys sharing a bin will stall. Prefer an explicit
// per-key CompletableFuture single-flight to avoid cross-key bin contention.
```

skills/principle-performance/examples/cache-aside.kt.md
```diff
@@ -0,0 +1,58 @@
...
+    // Note: kotlinx Mutex is NOT reentrant — calling withLock inside withLock on the same mutex deadlocks.
+    private val locks = ConcurrentHashMap<String, Mutex>()
+
+    suspend fun get(key: String): V {
+        store[key]?.let { if (Instant.now().isBefore(it.expiresAt)) return it.value }
+
+        val mutex = locks.computeIfAbsent(key) { Mutex() }
+        return mutex.withLock {
+            // Re-check after acquiring: a concurrent caller may have already populated the entry.
+            store[key]?.let { if (Instant.now().isBefore(it.expiresAt)) return it.value }
```
lugassawan: **Medium:** `return it.value` inside `mutex.withLock { }` is a *non-local* return from `get()`, not from the lambda. This is valid because `withLock` is `inline` — but readers adapting this to a non-inline higher-order function would see a compile error with no obvious explanation.

Add a brief comment to make the semantics explicit:
```kotlin
store[key]?.let { if (Instant.now().isBefore(it.expiresAt)) return it.value } // non-local return (withLock is inline)
```

skills/principle-performance/examples/cache-aside.py.md
```diff
@@ -0,0 +1,63 @@
...
+## Implementation
+
+```python
+# file: cache_aside.py
+import asyncio
+import time
+from collections import defaultdict
+from typing import Awaitable, Callable, TypeVar
+
+V = TypeVar("V")
```
lugassawan: **Low:** `V = TypeVar("V")` is defined but never used — the class annotates its value type as `object` throughout. The `TypeVar` declaration and the related `Callable`/`Awaitable` imports are partially dead in the example.

Either use `V` generically (`class CacheAside(Generic[V])`) or remove the unused `TypeVar`.

skills/principle-performance/examples/cache-aside.go.md
lugassawan: Addressed in 73324c4: wrapped loader in anonymous func with defer wg.Done() — a panicking loader no longer blocks all waiters forever.

skills/principle-performance/examples/cache-aside.swift.md
lugassawan: Addressed in 73324c4: added doc-comment note that cancelling the caller's Task does not cancel the shared unstructured loader Task.

skills/principle-performance/examples/cache-aside.java.md
lugassawan: Addressed in 73324c4: added note that compute holds the bin lock for the entire loader call; suggests per-key CompletableFuture for blocking I/O.

skills/principle-performance/examples/cache-aside.kt.md
lugassawan: Addressed in 73324c4: annotated the inner return as non-local (withLock is inline, so it returns from get(), not the lambda).

skills/principle-performance/examples/cache-aside.py.md
lugassawan: Addressed in 73324c4: removed unused TypeVar declaration and import.

skills/principle-performance/examples/cache-aside.go.md
```diff
@@ -0,0 +1,100 @@
...
+	cl.wg.Add(1)
+	c.inflight[key] = cl
+	c.mu.Unlock()
+
+	// Wrap in an anonymous func so defer fires on panic too — without this,
+	// a panicking loader would leave wg.Done() uncalled and all waiters blocked forever.
+	func() {
+		defer cl.wg.Done()
+		cl.val, cl.err = c.loader(key)
+	}()
```
lugassawan, line 78: **High (documentation):** If `c.loader(key)` panics inside the anonymous func, `defer cl.wg.Done()` correctly unblocks all waiters — but the panic propagates out of `func(){}()`, bypassing the `c.mu.Lock()` / `delete(c.inflight, key)` block (lines 75-81). The inflight entry is never cleaned up: all subsequent `Get(key)` calls read the stale `cl` (WaitGroup already at 0, so `Wait()` returns instantly) and get a zero-value + nil-error result silently.

Add a comment documenting the production gap:
```go
// Wrap in an anonymous func so defer fires on panic too — without this,
// a panicking loader would leave wg.Done() uncalled and all waiters blocked forever.
// Note: if the loader panics, the panic still propagates and the inflight entry is not
// cleaned up — future callers silently get the zero value. Use recover() in the defer
// to convert panics to errors and ensure inflight cleanup in production.
```

skills/principle-performance/examples/cache-aside.ts.md
```diff
@@ -0,0 +1,66 @@
...
+    if (existing) return existing;
+
+    const promise = this.loader(key).then((value) => {
+      this.store.set(key, { value, expiresAt: Date.now() + this.ttlMs });
+      this.inflight.delete(key);
+      return value;
+    });
+    // Attach a no-op catch so Node.js doesn't report an unhandled rejection on the shared
+    // inflight promise itself; callers still receive the error via their own await.
+    promise.catch(() => this.inflight.delete(key));
```
lugassawan: **Medium:** The current dual-path cleanup (`.then()` deletes on success, `.catch()` deletes on failure) is correct but fragile — a future edit that accidentally stores `promise.catch(...)` instead of `promise` in `inflight` would coerce rejections to `undefined` for all callers. `.finally()` eliminates the two-path cleanup and makes the intent explicit:

```typescript
const promise = this.loader(key)
  .then((value) => {
    this.store.set(key, { value, expiresAt: Date.now() + this.ttlMs });
    return value;
  })
  .finally(() => this.inflight.delete(key));
// No separate .catch() needed — .finally() runs on both success and failure,
// and does not swallow the rejection.
this.inflight.set(key, promise);
return promise;
```

skills/principle-performance/examples/cache-aside.swift.md
```diff
@@ -0,0 +1,70 @@
...
+        self.loader = loader
+    }
+
+    func get(_ key: String) async throws -> V {
+        if let entry = store[key], Date.now < entry.expiresAt {
+            return entry.value
+        }
+        // Single-flight via actor isolation: reuse an in-progress Task for the same key.
+        if let task = inflight[key] {
+            return try await task.value
```
lugassawan, line 42: **Low:** `Task { try await self.loader(key) }` captures `self` strongly. Because `CacheAside` is an actor, the inflight `Task` holds a strong reference to the actor — the actor is not deallocated while any loader task is in flight. Fine for long-lived singletons, but unexpected for short-lived actors.

Add a comment:
```swift
// `self` is captured strongly — the actor lives at least as long as any inflight Task.
// For short-lived actors, use `[weak self]` and guard-let before calling loader.
let task = Task { try await self.loader(key) }
```

skills/principle-performance/examples/cache-aside.go.md
lugassawan: Addressed in 45464b7: added comment documenting panic-inflight-poison gap and recover() guidance.

skills/principle-performance/examples/cache-aside.ts.md
lugassawan: Addressed in 45464b7: replaced dual-path .then(delete)+.catch(delete) with .finally(delete).

skills/principle-performance/examples/cache-aside.swift.md
lugassawan: Addressed in 45464b7: added inline comment noting strong self-capture and [weak self] guidance for short-lived actors.

skills/principle-performance/examples/cache-aside.rs.md
```diff
@@ -0,0 +1,84 @@
...
+// ✗ release lock before calling loader — other threads race to recompute the same key
+let mut store = self.store.lock().unwrap();
+if let Some(e) = store.get(&key) {
+    if Instant::now() < e.expires_at { return e.value.clone(); }
+}
+drop(store); // ✗ multiple threads now call loader concurrently for the same key
+let value = loader(&key);
+```
+
+Global lock held across a slow loader: concurrent reads for *any* key block until the loader returns — not just reads on the cold key.
```
lugassawan: **Medium (advisory):** The second Common Mistake block warns that "a slow loader blocks all concurrent reads for any key" — but the implementation above does exactly that (holds the global `Mutex` across the loader call). A reader sees the implementation and the Common Mistake as contradictory rather than complementary.

Clarify that the second mistake is a *different scenario* (high-cardinality + blocking I/O) distinct from the example which is scoped to moderate contention:

```markdown
Global lock held across a slow loader on a **high-cardinality** store: concurrent reads
for *any* key block until the loader returns. This is distinct from the implementation
above (sized for moderate contention). For high-cardinality or high-latency loaders,
prefer a per-key lock map (e.g. DashMap) to limit contention to one key at a time.
```

skills/principle-performance/examples/cache-aside.go.md
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Cache is a generic cache-aside store with single-flight on cold/expired keys.
+type Cache[V any] struct {
+	mu     sync.Mutex
+	store  map[string]entry[V]
+	// inflight gates concurrent misses: only one goroutine recomputes per key.
+	inflight map[string]*call[V]
+	ttl    time.Duration
+	loader func(key string) (V, error)
```
lugassawan, line 35: **Medium (advisory):** The `loader` signature is `func(key string) (V, error)` — standard Go convention is to pass `context.Context` as the first argument to any blocking or I/O function. A reader who copies this pattern builds a cache that cannot propagate deadlines or cancellation to the origin.

Consider threading `ctx` through:
```go
loader func(ctx context.Context, key string) (V, error)
```
And in `Get`: `cl.val, cl.err = c.loader(ctx, key)` (adding `ctx context.Context` to `Get`'s signature). This is advisory — the file's scope is single-flight, not Go idioms — but a note acknowledging the omission would prevent readers from copying it naively.

skills/principle-performance/examples/cache-aside.kt.md
```diff
@@ -0,0 +1,58 @@
...
+
+class CacheAside<V>(
+    private val ttl: Duration,
+    private val loader: suspend (String) -> V,
+) {
+    private val store = ConcurrentHashMap<String, Entry<V>>()
+    // Per-key Mutex: only one coroutine recomputes a cold/expired entry.
+    // computeIfAbsent is atomic per-key; getOrPut is NOT — two coroutines could get different instances.
+    // Note: kotlinx Mutex is NOT reentrant — calling withLock inside withLock on the same mutex deadlocks.
+    private val locks = ConcurrentHashMap<String, Mutex>()
```
lugassawan, line 32: **Medium (advisory):** The `locks` map (`ConcurrentHashMap<String, Mutex>`) accumulates one entry per distinct key and is never pruned — the same unbounded-growth risk documented in `cache-aside.cs.md` (which gives a mitigation). Kotlin's example is the only per-key lock file missing this note.

Add a comment parallel to C#:
```kotlin
// Note: locks accumulates one Mutex per distinct key and is never pruned.
// For high-cardinality keys (e.g. user IDs), remove the entry after withLock
// completes: locks.remove(key) — or accept the growth for bounded key spaces.
```

skills/principle-performance/examples/cache-aside.rs.md
lugassawan: Addressed in 774f2ce: clarified the second Common Mistake applies to high-cardinality stores, not the example's moderate-contention scope — removes the apparent contradiction.

skills/principle-performance/examples/cache-aside.go.md
lugassawan: Addressed in 774f2ce: added note that production loaders should accept context.Context for deadline/cancellation propagation.

skills/principle-performance/examples/cache-aside.kt.md
lugassawan: Addressed in 774f2ce: added note that locks map accumulates one Mutex per distinct key; suggest locks.remove(key) for high-cardinality key spaces.

PR #447, opened 2026-06-12
[fix] Fall back to cd when EnterWorktree/ExitWorktree reject rimba worktrees

lugassawan:
## Summary
- Add try-then-fallback contract for `EnterWorktree`/`ExitWorktree` across 4 worktree skill files
- Correct the overstatement that "`cd` silently does nothing" — Bash cwd does persist across calls; only session-level caches (plans dir, memory dir) are not re-anchored
- Add `cd <abs-path>` as sanctioned fallback when `EnterWorktree(path=…)` is rejected (nested-worktree or pinned-subagent context with rimba's `../<repo>-worktrees/` path outside `.claude/worktrees/`)
- Add `cd "$(git rev-parse --git-common-dir | sed 's|/\.git$||')"` as exit fallback when `ExitWorktree` is a no-op (cd-entered session); also fixes a pre-existing bug in `workflow-cleanup-merged` shell-fallback strategy that used `--show-toplevel` (returns worktree root) instead of `--git-common-dir` (returns main repo root)

## Test Plan

- [x] Changed skills/commands/agents load without errors (`./scripts/validate.sh` — all passed)
- [x] Examples in the diff were exercised manually (rimba round-trip: `rimba add`, `EnterWorktree` accepted from main session; `git rev-parse --git-common-dir | sed 's|/\.git$||'` confirmed to land at main repo root)
- [x] `EnterWorktree` was tried with rimba path — succeeded from main session, confirming try-first is correct; rejection context (nested worktree / pinned agent) documented accurately
- [x] 1332 pre-push tests pass

- [x] Reproduce: old text said "`cd` only affects that single shell subprocess … silently does nothing" — confirmed inaccurate (Bash cwd persists across calls; harness resets it only when an active `EnterWorktree` session is held)
- [x] Verify: reframed Forbidden-pattern section unambiguously forbids `cd` as *primary* switch while sanctioning it as fallback; no internal contradiction remains

Issue: N/A — skill behavior fix surfaced via /debug


skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -54,7 +54,15 @@ If the session is currently inside a worktree (e.g. entered via `EnterWorktree p
 - Releases the harness's session lock on the worktree so the rimba post-merge hook (fired by `git pull` in 3c) can remove it cleanly.
 - Ensures rimba's binary `remove` strategy (if reached) won't fire `git branch -D` from a deleted cwd.
 
-If `EnterWorktree` was never called this session (or the `ExitWorktree` tool is unavailable), this step is a no-op — proceed to 3b without aborting.
+If the worktree was entered via the `cd` fallback (no active `EnterWorktree` session), `ExitWorktree` is a no-op — instead, `cd` to the main repo root before deriving `$MAIN_REPO` and running `git pull`:
+
+```bash
+cd "$(git rev-parse --git-common-dir | sed 's|/\.git$||')"
```
lugassawan: **Critical — `git-common-dir` pipeline is broken when called from the main repo**

```bash
cd "$(git rev-parse --git-common-dir | sed 's|/\.git$||')"
```

`git rev-parse --git-common-dir` returns the **relative** string `.git` when executed from the main worktree — not an absolute path. The `sed` pattern `s|/\.git$||` requires a leading `/` to match, so it silently does nothing, and the command becomes `cd .git`, navigating into the git metadata directory.

The command is only safe when confirmed to be inside a **linked worktree** (where `--git-common-dir` returns an absolute path like `/path/to/main/.git`).

The prose comment on line 63 compounds this by stating the output is always an absolute path — that claim is incorrect.

**Suggested fix** — use `git worktree list` (always absolute) or add a guard:
```bash
_GCD=$(git rev-parse --git-common-dir)
# relative (.git) means we're already at main root — nothing to do
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```

This same issue exists in `skills/workflow-worktree-session/SKILL.md:49` and `skills/workflow-cleanup-merged/SKILL.md:202`.

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -54,7 +54,15 @@ If the session is currently inside a worktree (e.g. entered via `EnterWorktree p
...
+```
+
+`git rev-parse --git-common-dir` returns the absolute path to the common `.git` directory (e.g. `/path/to/main/.git`); stripping `/.git` yields the main repo root. This is reliable from any linked worktree regardless of how git output is filtered.
```
lugassawan: **High — Prose description is factually wrong**

> `git rev-parse --git-common-dir` returns the absolute path to the common `.git` directory

This claim is incorrect. From the **main worktree** the output is the relative string `.git`, not an absolute path. This misleads any agent reading the skill into trusting the command unconditionally without adding a guard.

Suggested correction: "`git rev-parse --git-common-dir` returns the path to the common `.git` directory — absolute from a linked worktree (e.g. `/path/to/main/.git`), relative (`.git`) from the main worktree. Only call this after confirming the cwd is inside a linked worktree."

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -190,8 +198,8 @@ eval "$("$_SCRIPTS/probe-worktree.sh" "<headRefName>")"
 *[Optional] cwd-fix*
 
 If `cwd` is a subdirectory of `$WORKTREE`, cd to the main repo root before removal:
-```
-cd "$(git rev-parse --show-toplevel)"
+```bash
+cd "$(git rev-parse --git-common-dir | sed 's|/\.git$||')"
```
lugassawan: **High — Wrong semantics: `git-common-dir` replaces `show-toplevel` in the wrong context**

This hunk replaces `git rev-parse --show-toplevel` with `git rev-parse --git-common-dir | sed ...` in the *cwd-fix* step, whose purpose is to escape a **subdirectory of the worktree** back to the worktree's root before calling `git worktree remove`.

From inside a linked worktree:
- `git rev-parse --show-toplevel` → the **linked worktree root** (e.g. `/path/to/worktrees/feat-x`) ← correct here
- `git rev-parse --git-common-dir | sed ...` → the **main repo root** (e.g. `/path/to/main`) ← wrong here

The original `--show-toplevel` was correct for the cwd-fix step: it brings the cwd up to the top of the worktree, then `git worktree remove` runs cleanly. Replacing it with `--git-common-dir` silently changes the destination to the main repo root — a different level entirely.

**Suggested fix:** revert this hunk. The `--git-common-dir` form belongs only in Step 3a (ExitWorktree fallback) where navigating to the main root is intentional.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,19 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
...
 
-**Active remedy:** If you notice you have already been prepending `cd <worktree>` to commands this session, that is the signal — stop and call `EnterWorktree(path=<worktree-path>)` now. The cd-prefix is not a workaround; it leaves the session anchored elsewhere and bypasses the `ExitWorktree action=keep` lock contract.
+**Active remedy:** If you notice you have already been prepending `cd <worktree>` to commands this session, that is the signal — stop and try `EnterWorktree(path=<worktree-path>)` now. If it succeeds, the session is properly anchored. If it is rejected (session is already in a worktree + rimba path is outside `.claude/worktrees/`), the `cd`-prefix is the correct sanctioned fallback — ensure you are consistently `cd`-ing to the same path and do not also call `ExitWorktree` (it is a no-op for `cd`-entered worktrees; use the `cd`-to-main-root fallback in Mode C instead).
```
lugassawan: **Medium — "rimba path" is narrower than the actual rejection condition**

The Forbidden pattern section says: "session is already in a worktree + **rimba path** is outside `.claude/worktrees/`".

But Mode A step 2 correctly says: "session is already inside another worktree AND the **path** is outside `.claude/worktrees/`".

The phrase "rimba path" implies the fallback only applies to rimba-managed worktrees. An agent reading only this section might not apply the `cd` fallback when the worktree was created via plain `git worktree add`.

Suggested: change "rimba path is outside" → "path is outside" to match the formulation in Mode A.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,19 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
 
 Triggers: "exit the worktree", "go back", "leave this worktree", "return to main".
 
-Call `ExitWorktree(action: "keep")` by default.
+Call `ExitWorktree(action: "keep")` by default. If the worktree was entered via the `cd` fallback (no active `EnterWorktree` session), `ExitWorktree` is a no-op — run the following instead to return to the main repo root:
```
lugassawan: **Medium — No mechanism to detect whether the session was entered via `cd` vs `EnterWorktree`**

Mode C says "if the worktree was entered via the `cd` fallback, `ExitWorktree` is a no-op" — but neither this file nor `workflow-cleanup-merged` explains how to make that determination at runtime, especially after a session resume when in-session state is gone.

An agent that cannot tell which entry method was used will either:
- Always call `ExitWorktree` (wrong for `cd`-entered sessions — it's a no-op, but the agent may still be confused when it gets no acknowledgement), or
- Always use the `cd` command (skips proper harness lock release for `EnterWorktree`-anchored sessions)

**Suggested addition:** "To determine which exit path to use, attempt `ExitWorktree(action=keep)`. If it reports success, the session was `EnterWorktree`-anchored. If it reports it is unavailable or a no-op, use the `cd`-to-main-root command."

skills/workflow-development/SKILL.md
```diff
@@ -148,7 +148,7 @@ Use the service scope whenever the work is clearly contained within one module 
...
 
-**Enter the worktree:** After `Path: <abs-path>` appears, call `EnterWorktree(path=<abs-path>)` to anchor the session — `cd` only affects a single Bash subprocess and does **not** move the session. On any resumed or continued session, call `EnterWorktree(path=<worktree-path>)` before running commands; if you catch yourself cd-prefixing, that is the signal to re-anchor.
+**Enter the worktree:** After `Path: <abs-path>` appears, try `EnterWorktree(path=<abs-path>)`. From the main session this works for any git-registered worktree (including rimba's `../<repo>-worktrees/` layout). **If rejected** — which happens when the session is already inside another worktree AND the path is outside `.claude/worktrees/` (e.g. a pinned subagent) — fall back to `cd <abs-path>` via Bash; the persistent Bash cwd anchors subsequent commands in the worktree, though session-level caches are not re-anchored. On any resumed or continued session, try `EnterWorktree(path=<worktree-path>)` first; if rejected, re-`cd <worktree-path>` (the Bash cwd does not persist across session resume, so `cd` must be re-issued each time — unlike `EnterWorktree`, which restores it automatically).
```
lugassawan: **Low — Rejection condition enumeration may miss non-rimba rejection causes**

The phrase "**If rejected** — which happens when the session is already inside another worktree AND the path is outside `.claude/worktrees/`" enumerates a specific two-clause conjunction. If `EnterWorktree` is rejected for any other reason (tool unavailable, harness version mismatch, etc.), agents may not apply the `cd` fallback.

Suggested: "**If rejected for any reason** (most commonly: session already inside a worktree with the target path outside `.claude/worktrees/`) — fall back to `cd <abs-path>` via Bash."

This wording applies to all 4 files changed in this PR.

skills/workflow-development/templates/plan-workflow-section.md
```diff
@@ -24,8 +24,8 @@ Copy this `## Workflow` section into your plan and substitute `[[detect:KEY]]` m
...
+- **Enter worktree:** Try `EnterWorktree path=<rimba-output-path>` first — this works from the main session for any git-registered worktree (including rimba's `../<repo>-worktrees/` layout). If rejected (session is already inside a worktree AND the path is outside `.claude/worktrees/`), fall back to `cd <rimba-output-path>` via Bash; note that `cd` only anchors the Bash persistent cwd and does not re-anchor session-level caches the way `EnterWorktree` does.
+- **Resume note:** record the worktree path (`Path:` line from `rimba add`) in the plan. On any continued or resumed session, try `EnterWorktree path=<that-path>` first; if rejected, re-`cd <that-path>` (Bash cwd resets on session resume). If you catch yourself cd-prefixing commands, that is the signal to stop and try `EnterWorktree path=<worktree-path>` — if it is rejected, the cd-prefix is the sanctioned fallback.
```
lugassawan, line 28: **Low — Plan template covers re-entry but not exit for `cd`-fallback sessions**

The resume note explains how to re-enter the worktree on session resume (try `EnterWorktree`, else re-`cd`). It doesn't mention exit — specifically, that agents following this plan who later run `workflow-cleanup-merged` with a `cd`-entered worktree must use the `cd`-to-main-root command instead of `ExitWorktree`.

An agent following only the plan template will have to discover this contract by chance when it reaches Step 3a of `workflow-cleanup-merged`. A one-line cross-reference closes the gap without adding bulk.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: guard the git-common-dir pipeline with [[ "$_GCD" != /* ]] so it is a no-op from the main repo and avoids cd .git.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: corrected the prose — git-common-dir returns a relative path (.git) from the main worktree, not an absolute one; updated the description to reflect both cases.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: reverted git-common-dir back to show-toplevel in the cwd-fix context — show-toplevel targets the linked worktree root (correct), git-common-dir targets the main repo root (wrong for this step).

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: changed 'rimba path is outside' to 'path is outside' to match Mode A and not imply the fallback is rimba-only.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: added exit-detection guidance — attempt ExitWorktree(action=keep) first; if no-op/unavailable, use the cd-to-main-root command. Also fixed the same broken git-common-dir pipeline in Mode C.

skills/workflow-development/SKILL.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: changed 'If rejected — which happens when...' to 'If rejected for any reason (most commonly: ...)' across all 4 files.

skills/workflow-development/templates/plan-workflow-section.md
lugassawan: Addressed in 1b70175085db9ec36754fcd57330479070d80d2a: added an Exit (cd-fallback sessions) bullet cross-referencing workflow-cleanup-merged Step 3a so agents know how to exit without discovering it by chance.

skills/workflow-development/templates/plan-workflow-section.md
```diff
@@ -24,8 +24,9 @@ Copy this `## Workflow` section into your plan and substitute `[[detect:KEY]]` m
...
+- **Exit (cd-fallback sessions):** if the worktree was entered via `cd` (not `EnterWorktree`), skip `ExitWorktree` — instead, attempt `ExitWorktree(action=keep)`; if it reports a no-op, return to main with `_GCD=$(git rev-parse --git-common-dir); [[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"`. See `workflow-cleanup-merged` Step 3a for the full contract.
```
lugassawan: **Medium — Contradictory "skip / attempt" instruction**

The bullet opens with "skip \`ExitWorktree\`" then immediately says "attempt \`ExitWorktree(action=keep)\`" — two directly opposed instructions in the same sentence. A model following this literally will either execute the first clause (skip it) and ignore the second, or stall trying to reconcile them.

The canonical contract in `workflow-worktree-session` Mode C uses attempt-first without any "skip" preamble. This bullet diverges from its own authority.

**Suggested fix** — remove the "skip" clause and lead with the attempt-first protocol:
```markdown
- **Exit (cd-fallback sessions):** if uncertain whether the worktree was entered via `EnterWorktree` or `cd`, attempt `ExitWorktree(action=keep)` first; if it reports a no-op, return to main with `_GCD=$(git rev-parse --git-common-dir); [[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"`. See `workflow-worktree-session` Mode C for the full exit contract.
```

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -54,7 +54,17 @@ If the session is currently inside a worktree (e.g. entered via `EnterWorktree p
 - Releases the harness's session lock on the worktree so the rimba post-merge hook (fired by `git pull` in 3c) can remove it cleanly.
 - Ensures rimba's binary `remove` strategy (if reached) won't fire `git branch -D` from a deleted cwd.
 
-If `EnterWorktree` was never called this session (or the `ExitWorktree` tool is unavailable), this step is a no-op — proceed to 3b without aborting.
+If the worktree was entered via the `cd` fallback (no active `EnterWorktree` session), `ExitWorktree` is a no-op — instead, `cd` to the main repo root before deriving `$MAIN_REPO` and running `git pull`:
+
+```bash
+_GCD=$(git rev-parse --git-common-dir)
```
lugassawan, line 60: **Medium — `${_GCD%/.git}` produces wrong destination for non-standard git setups**

`${_GCD%/.git}` strips exactly the literal suffix `/.git`. When the common dir does not end in `/.git` (e.g. `--separate-git-dir`, bare repos, or submodule common dirs like `.git/modules/sub`), the strip is a no-op and `cd "${_GCD%/.git}"` navigates into the git object store rather than the working tree root.

Concrete failure: a repo created with `git clone --separate-git-dir /data/git/myrepo.git` will have `--git-common-dir` return `/data/git/myrepo.git`. The `%/.git` strip does nothing, and the cd lands inside the bare object store. Subsequent `git` commands run from the wrong directory and produce confusing errors. Realistic in CI environments.

Same issue exists identically in `skills/workflow-worktree-session/SKILL.md:49` and `skills/workflow-development/templates/plan-workflow-section.md:29`.

**Suggested fix** — add an explicit suffix check, or at minimum a prose caveat:
```bash
_GCD=$(git rev-parse --git-common-dir)
if [[ "$_GCD" == /* ]]; then
  # Standard embedded .git layout — strip /.git to get main repo root
  if [[ "$_GCD" == */.git ]]; then
    cd "${_GCD%/.git}"
  else
    # Non-standard layout (--separate-git-dir, etc.) — fall back to show-toplevel on main
    cd "$(git -C "$_GCD" rev-parse --show-toplevel 2>/dev/null || dirname "$_GCD")"
  fi
fi
```

Or add a prose note: "Assumes a standard embedded `.git` directory; `--separate-git-dir` setups require a different approach."

skills/workflow-worktree-session/SKILL.md
```diff
@@ -27,7 +27,7 @@ Examples: "open the `feat-login` worktree", "switch to `test-enter`", "move into
 
    Match the user's name against branch names or directory basenames. If no match is found, tell the user no worktree matched their description and stop — do not call `EnterWorktree` with an empty path.
 
-2. Call `EnterWorktree(path=<absolute-path>)`.
+2. Enter the worktree. **Try `EnterWorktree(path=<absolute-path>)` first.** From the main session this works for any registered worktree (including rimba's `../<repo>-worktrees/` layout). If it is rejected for any reason (most commonly: session already inside a worktree with the target path outside `.claude/worktrees/`) — fall back to `cd <absolute-path>` via Bash. The Bash tool's working directory persists across calls, so subsequent git/build/test commands run from the worktree; note that this fallback does not re-anchor session-level caches (plans dir, memory dir) the way `EnterWorktree` does.
 
 3. Confirm: run `git rev-parse --git-dir --git-common-dir` and verify the two paths differ (linked worktree, not main). Report the new CWD.
```
lugassawan, line 32: **Low — Step 3 confirmation is not mentioned for the cd-fallback path**

Step 3 instructs "run `git rev-parse --git-dir --git-common-dir` and verify the two paths differ" — this appears under the Mode A flow without noting it applies equally when the cd fallback is used. An agent that took the cd path may skip the confirmation, silently running commands in the wrong directory with no signal.

**Suggested addition:** Append a sentence to Step 2's fallback clause: "After `cd <path>`, run the Step 3 confirmation — the `--git-dir ≠ --git-common-dir` check works identically regardless of entry method."

skills/workflow-development/templates/plan-workflow-section.md
lugassawan: Addressed in 97a808e2144db7ac9e42aeda56d394c3b28fa98c: removed the contradictory 'skip ExitWorktree' clause; the bullet now leads with attempt-first matching workflow-worktree-session Mode C. Also corrected the cross-reference from workflow-cleanup-merged Step 3a → workflow-worktree-session Mode C.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 97a808e2144db7ac9e42aeda56d394c3b28fa98c: added a prose caveat to all three git-common-dir command blocks noting that --separate-git-dir and submodule common dirs may not end in /.git, making the %/.git strip a no-op and requiring a different navigation strategy.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in 97a808e2144db7ac9e42aeda56d394c3b28fa98c: extended the Mode A step 2 fallback clause to explicitly note that the step 3 --git-dir vs --git-common-dir confirmation applies equally after the cd fallback, regardless of entry method.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -60,12 +69,12 @@ Do not call `ExitWorktree` proactively. Only call it when the user explicitly re
...
+| "open the `feat-login` worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "switch to `.worktrees/auth`" | A | `EnterWorktree(path=<abs-path>)` → else `cd <path>` |
+| "continue work in the worktree I made" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "resume in the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "I've been cd-ing into the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
 | "in a fresh worktree" | B | defer to `superpowers:using-git-worktrees` |
-| "exit the worktree" | C | `ExitWorktree(action: "keep")` |
-| "delete this worktree and go back" | C | `ExitWorktree(action: "remove")` |
+| "exit the worktree" | C | `ExitWorktree(action: "keep")` → else `cd <main-root>` |
+| "delete this worktree and go back" | C | `ExitWorktree(action: "remove")` → else `git worktree remove <path> && cd <main-root>` |
```
lugassawan: **Medium: Quick-reference row for "delete this worktree and go back" provides a fallback absent from Mode C prose**

The table at line 79 shows:
```
ExitWorktree(action: "remove") → else git worktree remove <path> && cd <main-root>
```
But Mode C body text only says "`ExitWorktree(action: "remove")` **only** when the user explicitly says *remove*, *delete*, or *clean up*" — there is no procedure for what to do when `ExitWorktree(action: "remove")` is rejected in a `cd`-entered session. A reader following Mode C step-by-step has no fallback procedure, while the table implies one.

**Suggested fix:** Add to Mode C after the `ExitWorktree(action: "remove")` line:
```bash
# If ExitWorktree(action: "remove") is rejected or unavailable and the session is cd-entered:
git worktree remove "$(git rev-parse --show-toplevel)"
_GCD=$(git rev-parse --git-common-dir)
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```
(Remove the worktree first, then navigate to the main repo root.)

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,23 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
 
 Triggers: "exit the worktree", "go back", "leave this worktree", "return to main".
 
-Call `ExitWorktree(action: "keep")` by default.
+Call `ExitWorktree(action: "keep")` by default. If uncertain whether the session was entered via `EnterWorktree` or the `cd` fallback, attempt `ExitWorktree(action=keep)` first — if it succeeds the session was `EnterWorktree`-anchored and exit is complete; if it reports a no-op or is unavailable, the session was `cd`-entered — run the following instead to return to the main repo root:
```
lugassawan: **Low: "if it reports a no-op or is unavailable" conflates tool-absent with `cd`-entered**

Line 46: "if it reports a no-op or is unavailable, the session was `cd`-entered"

`ExitWorktree` being *unavailable* (tool not registered in the harness) does not necessarily mean the session was `cd`-entered — it could be an `EnterWorktree`-anchored session on a harness version that lacks the tool. The guarded `cd` is mechanically safe in either case, but the prose makes a diagnostic claim ("the session was `cd`-entered") that is not always accurate.

**Suggested fix:** Change to: "if it reports a no-op (or if the tool is unavailable — treat the following as a best-effort fallback)"

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -54,7 +54,17 @@ If the session is currently inside a worktree (e.g. entered via `EnterWorktree p
...
+# relative (.git) means we're already at main root — nothing to do
+[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
+```
+
+`git rev-parse --git-common-dir` returns the path to the common `.git` directory — absolute from a linked worktree (e.g. `/path/to/main/.git`), relative (`.git`) from the main worktree itself. The guard `[[ "$_GCD" != /* ]]` skips the `cd` when already at main root. **Assumes a standard embedded `.git` directory**; repos created with `--separate-git-dir` or submodule common dirs (e.g. `.git/modules/sub`) may return a path that does not end in `/.git`, in which case `${_GCD%/.git}` is a no-op and a different navigation strategy is needed.
+
+If `EnterWorktree` was never called this session and the cwd is not inside a worktree, this step is a no-op — proceed to 3b without aborting.
```
lugassawan: **Low: Third no-op paragraph in Step 3a is redundant with the guarded command**

Line 67: "If `EnterWorktree` was never called this session and the cwd is not inside a worktree, this step is a no-op — proceed to 3b without aborting."

When "cwd is not inside a worktree," `git rev-parse --git-common-dir` returns the relative `.git` path, which triggers `[[ "$_GCD" != /* ]]` to short-circuit — the `cd` is already skipped. The standalone third paragraph implies a distinct third scenario requiring different handling, which could mislead readers into thinking the guarded command does not apply to the "not in a worktree at all" case.

**Suggested fix:** Fold into the cd-fallback paragraph as a parenthetical: "(if the cwd is not inside any worktree, the relative `.git` path causes the guard to short-circuit — this is already a no-op)" and remove the standalone paragraph.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in c1ef79fb47b75a9af3da274aa4d0715b28c7c911: added cd-fallback remove procedure to Mode C prose after the ExitWorktree(action: "remove") line.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in c1ef79fb47b75a9af3da274aa4d0715b28c7c911: split the diagnostic claim — no-op confirms cd-entry; tool-unavailable now labeled as best-effort fallback with unknown entry method.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in c1ef79fb47b75a9af3da274aa4d0715b28c7c911: folded the standalone no-op paragraph into the Step 3a explanatory prose — the relative .git short-circuit already handles this case and the new parenthetical makes it explicit.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,29 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
...
+# relative (.git) means we're already at main root — nothing to do
+[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
+```
+
+> **Note:** assumes a standard embedded `.git` directory; `--separate-git-dir` setups and submodule common dirs may return a path that does not end in `/.git`, requiring a different navigation strategy.
+
+Call `ExitWorktree(action: "remove")` **only** when the user explicitly says *remove*, *delete*, or *clean up* the worktree. If it is rejected or unavailable and the session is `cd`-entered, remove the worktree and then navigate to the main repo root:
+
+```bash
+git worktree remove "$(git rev-parse --show-toplevel)"
```
lugassawan: **High: Mode C remove fallback calls `git rev-parse --git-common-dir` after the worktree is deleted — will fail with "fatal: not a git repository"**

```bash
git worktree remove "$(git rev-parse --show-toplevel)"
_GCD=$(git rev-parse --git-common-dir)   # ← cwd is gone at this point
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```

`git worktree remove` deletes the worktree directory. The shell's `cwd` is now a dangling path. Git walks up from cwd to find `.git`, which fails when the directory no longer exists — making `_GCD` empty and leaving the session stranded in a deleted directory.

**Suggested fix:** capture `_GCD` *before* the remove:

```bash
_GCD=$(git rev-parse --git-common-dir)
git worktree remove "$(git rev-parse --show-toplevel)"
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,29 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
...
-Call `ExitWorktree(action: "remove")` **only** when the user explicitly says *remove*, *delete*, or *clean up* the worktree.
+Call `ExitWorktree(action: "keep")` by default. If uncertain whether the session was entered via `EnterWorktree` or the `cd` fallback, attempt `ExitWorktree(action=keep)` first — if it succeeds the session was `EnterWorktree`-anchored and exit is complete; if it reports a no-op (confirms `cd`-entry) or is unavailable (treat the following as a best-effort fallback — entry method is unknown) — run the following to return to the main repo root:
+
+```bash
+_GCD=$(git rev-parse --git-common-dir)
+# relative (.git) means we're already at main root — nothing to do
+[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
+```
+
+> **Note:** assumes a standard embedded `.git` directory; `--separate-git-dir` setups and submodule common dirs may return a path that does not end in `/.git`, requiring a different navigation strategy.
```
lugassawan, line 54: **Medium: `--separate-git-dir` caveat Note block is absent from Mode C remove path**

The keep path (line 54) carries a Note warning that `--separate-git-dir` setups may not produce a `/.git`-suffixed path. The remove path uses the identical `${_GCD%/.git}` expansion with no caveat. Since the remove path is destructive (worktree is gone before the `cd` is attempted), the caveat is arguably more important here.

**Suggested fix:** Add the same Note block after the remove fallback code block.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -60,12 +75,12 @@ Do not call `ExitWorktree` proactively. Only call it when the user explicitly re
...
+| "open the `feat-login` worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "switch to `.worktrees/auth`" | A | `EnterWorktree(path=<abs-path>)` → else `cd <path>` |
+| "continue work in the worktree I made" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "resume in the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "I've been cd-ing into the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
 | "in a fresh worktree" | B | defer to `superpowers:using-git-worktrees` |
-| "exit the worktree" | C | `ExitWorktree(action: "keep")` |
-| "delete this worktree and go back" | C | `ExitWorktree(action: "remove")` |
+| "exit the worktree" | C | `ExitWorktree(action: "keep")` → else `cd <main-root>` |
+| "delete this worktree and go back" | C | `ExitWorktree(action: "remove")` → else `git worktree remove <path> && cd <main-root>` |
```
lugassawan: **Medium: Quick-reference table Mode C remove row shows dangerous pre-remove ordering**

The table cell shows `git worktree remove <path> && cd <main-root>` — removal then navigation. After the High fix above is applied (capture GCD before remove), the table should match. A reader skimming the table without reading the body will implement it in the wrong order.

**Suggested fix:** Update table cell to reflect capture-first ordering, e.g. "capture GCD → remove worktree → cd to main-root".

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -54,7 +54,15 @@ If the session is currently inside a worktree (e.g. entered via `EnterWorktree p
 - Releases the harness's session lock on the worktree so the rimba post-merge hook (fired by `git pull` in 3c) can remove it cleanly.
 - Ensures rimba's binary `remove` strategy (if reached) won't fire `git branch -D` from a deleted cwd.
 
-If `EnterWorktree` was never called this session (or the `ExitWorktree` tool is unavailable), this step is a no-op — proceed to 3b without aborting.
+If the worktree was entered via the `cd` fallback (no active `EnterWorktree` session), `ExitWorktree` is a no-op — instead, `cd` to the main repo root before deriving `$MAIN_REPO` and running `git pull`:
```
lugassawan: **Low: "ExitWorktree is a no-op" overstates certainty for cd-entered sessions**

"If the worktree was entered via the `cd` fallback… `ExitWorktree` is a no-op" states the outcome as a given. Mode C uses attempt-first precisely because the session cannot know in advance. The nearby Step 3a parenthetical handles this correctly; this sentence subtly contradicts the attempt-first protocol.

**Suggested fix:** Change to "ExitWorktree will report a no-op (confirming cd-entry)".

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in c5ac846ed74b024add5a517db67436fcf25ec7f7: reordered to capture _GCD before git worktree remove — cwd is gone after remove so git rev-parse would fail.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in c5ac846ed74b024add5a517db67436fcf25ec7f7: added the same --separate-git-dir Note block after the remove fallback code block.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in c5ac846ed74b024add5a517db67436fcf25ec7f7: updated table row to show capture-first ordering (capture GCD → remove worktree → cd to main-root).

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in c5ac846ed74b024add5a517db67436fcf25ec7f7: changed 'ExitWorktree is a no-op' to 'will report a no-op (confirming cd-entry)' to match the attempt-first protocol.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,31 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
...
+```
+
+> **Note:** assumes a standard embedded `.git` directory; `--separate-git-dir` setups and submodule common dirs may return a path that does not end in `/.git`, requiring a different navigation strategy.
+
+Call `ExitWorktree(action: "remove")` **only** when the user explicitly says *remove*, *delete*, or *clean up* the worktree. If it is rejected or unavailable and the session is `cd`-entered, remove the worktree and then navigate to the main repo root:
+
+```bash
+_GCD=$(git rev-parse --git-common-dir)
+git worktree remove "$(git rev-parse --show-toplevel)"
+[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```
lugassawan, line 63: **Important:** The remove-variant snippet lacks the inline comment present in the keep-variant (`# relative (.git) means we're already at main root — nothing to do`), making the guard harder to reason about. More critically, there is no pre-condition warning: if `ExitWorktree(action=remove)` is rejected while already at main root (not inside a linked worktree), `git rev-parse --show-toplevel` returns the main repo root and `git worktree remove <main-repo-root>` fails with "fatal: is a main worktree". The failure is safe (no data loss) but undocumented.

Suggested fix — add the inline comment and a brief pre-condition note:

```bash
# Only run this snippet when confirmed to be inside a linked worktree
_GCD=$(git rev-parse --git-common-dir)
# relative (.git) means we're already at main root — nothing to do
git worktree remove "$(git rev-parse --show-toplevel)"
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```

skills/workflow-worktree-session/SKILL.md
```diff
@@ -60,12 +77,12 @@ Do not call `ExitWorktree` proactively. Only call it when the user explicitly re
...
-| "I've been cd-ing into the worktree" | A | `EnterWorktree(path=<resolved>)` |
+| "open the `feat-login` worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "switch to `.worktrees/auth`" | A | `EnterWorktree(path=<abs-path>)` → else `cd <path>` |
+| "continue work in the worktree I made" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "resume in the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
+| "I've been cd-ing into the worktree" | A | `EnterWorktree(path=<resolved>)` → else `cd <path>` |
 | "in a fresh worktree" | B | defer to `superpowers:using-git-worktrees` |
-| "exit the worktree" | C | `ExitWorktree(action: "keep")` |
-| "delete this worktree and go back" | C | `ExitWorktree(action: "remove")` |
+| "exit the worktree" | C | `ExitWorktree(action: "keep")` → else `cd <main-root>` |
```
lugassawan: **Important:** The Quick Reference entry for "exit the worktree" shows `ExitWorktree(action: "keep") → else cd <main-root>` with no qualification. When `ExitWorktree` is **unavailable** (not just a no-op), the `cd` fallback runs as best-effort — but for a session entered via `EnterWorktree`, the harness session lock is never released. The prose in Mode C correctly labels this case "treat the following as a best-effort fallback — entry method is unknown," but the Quick Reference omits that caveat.

Suggested fix:
```
| "exit the worktree" | C | `ExitWorktree(action: "keep")` → else `cd <main-root>` (best-effort if tool unavailable; lock may remain) |
```

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in 74bdde9bc26abbd84609ff42c92830a77ff25462: added pre-condition note (only run when in a linked worktree) and inline comment matching the keep-variant guard.

skills/workflow-worktree-session/SKILL.md
lugassawan: Addressed in 74bdde9bc26abbd84609ff42c92830a77ff25462: Quick Reference row for 'exit the worktree' now reads '→ else `cd <main-root>` (best-effort if tool unavailable; lock may remain)'.

skills/workflow-worktree-session/SKILL.md
```diff
@@ -43,14 +43,32 @@ Defer entirely to `superpowers:using-git-worktrees`. That skill handles consent,
...
+[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
+```
+
+> **Note:** assumes a standard embedded `.git` directory; `--separate-git-dir` setups and submodule common dirs may return a path that does not end in `/.git`, requiring a different navigation strategy.
+
+Call `ExitWorktree(action: "remove")` **only** when the user explicitly says *remove*, *delete*, or *clean up* the worktree. If it is rejected or unavailable and the session is `cd`-entered, remove the worktree and then navigate to the main repo root. Only run this snippet when confirmed to be inside a linked worktree (if cwd is already at main root, `git worktree remove` will fail with "fatal: is a main worktree"):
+
+```bash
+_GCD=$(git rev-parse --git-common-dir)
+# relative (.git) means we're already at main root — nothing to do
```
lugassawan: **Important:** The comment `# relative (.git) means we're already at main root — nothing to do` is placed between `_GCD=…` and `git worktree remove`, making it look like a guard over the removal. In the keep-variant this comment is accurate — the sole operation is the `cd`, so "nothing to do" describes the whole block's outcome when the guard fires. In the remove-variant, `git worktree remove` executes **unconditionally** on the next line — the comment only protects the final `cd`. An agent reading this snippet in isolation may incorrectly infer the guard makes the whole snippet safe to run from main root, despite the pre-condition note in the prose.

Suggested fix — move the comment to sit immediately above the `cd` it governs, and add an ordering comment:

```bash
# capture before remove — path is gone after git worktree remove
_GCD=$(git rev-parse --git-common-dir)
git worktree remove "$(git rev-parse --show-toplevel)"
# relative (.git) means we're already at main root — cd is a no-op
[[ "$_GCD" != /* ]] || cd "${_GCD%/.git}"
```

lugassawan: Addressed in 8d744bd637bfc8fc9e52aa1a78d8c1975d85f78d: moved the guard comment to sit immediately above the cd it governs; added ordering comment above _GCD explaining why capture must precede git worktree remove.

PR #464, opened 2026-06-22
[fix] Mode A embeds full ## Workflow template, not a summary

### LINKED ISSUE #455: [bug] workflow-development Mode A embeds a summarized ## Workflow section instead of the full template body
## Product framing

- **Problem.** When a plan-producing command finalizes a plan, the embedded `## Workflow` section
  comes out as a condensed "short version" — the executor loses the detailed Branch/Verify/Deliver
  guidance the template carries (rimba flag taxonomy, worktree enter/exit fallbacks, the
  Imports→Format→Quality→Lint→Test order, PR-template handling).
- **Value.** Anyone running `/debug`, `/implement`, `/migrate`, etc. gets the *full, correct*
  lifecycle steps in their plan. The workflow guidance is only useful if reproduced faithfully;
  a summary silently drops the exact instructions that prevent common mistakes.
- **Acceptance criteria.**
  - Mode A wording (`SKILL.md:237`) explicitly requires the `## Workflow` section to be reproduced
    **in full / verbatim from the template** — only `[[detect:KEY]]` markers may be substituted; no
    summarizing, paraphrasing, or step-collapsing.
  - The template header (`plan-workflow-section.md:3`) reinforces "copy in full, do not abridge."
  - Running any plan-producing command (`/debug` as the repro) yields a plan whose `## Workflow`
    section matches the template body line-for-line (modulo detected values).
  - (Optional / nice-to-have) a `scripts/validate.py` check that the rendered section isn't a
    truncated form of the template.
- **Impact / Effort.** `Impact: M` — affects every plan from 8 commands; degrades execution
  fidelity but doesn't break the pipeline. `Effort: S` — primarily a wording/guard change in one
  skill file plus the template header.

## Surface affected

- [x] Skill — `workflow-*`  (`workflow-development`, Mode A)
- [x] Slash command  (repro surface: `/swe-workbench:debug`; also implement/migrate/refactor/test/design/architect/document)

**Specific surface:** `skills/workflow-development` (Mode A plan embedding) — repro via `/swe-workbench:debug`

## Claude Code version

2.1.179

## Plugin version

0.1.20

## Reproduction steps

1. Run `/swe-workbench:debug <some bug>` and let it author a plan (Mode A is always active for
   `/debug` since it produces a fix).
2. Inspect the `## Workflow` section embedded in the generated plan.
3. Compare it against `skills/workflow-development/templates/plan-workflow-section.md`.

## Expected behavior

The embedded `## Workflow` section reproduces the template body in full (Phases 1–5 with all rimba
flag/worktree/verify/deliver detail), substituting only `[[detect:KEY]]` markers.

## Actual behavior

The section is rendered as a summarized / short version — phases are collapsed to brief steps and
most of the template's detailed guidance is dropped. Root cause: `SKILL.md:237` mandates
`[[detect:]]` substitution but never mandates full-fidelity reproduction, so the orchestrator
condenses the 96-line template. The command file (`debug.md:94`) is correct and is **not** the
defect.

## Environment (optional)

macOS, zsh, Opus 4.8

lugassawan:
## Summary
- Add explicit full-fidelity mandate to `skills/workflow-development/SKILL.md` Mode A paragraph: "Reproduce the template's `## Workflow` body **in full and verbatim**" — the only permitted edit is substituting `[[detect:KEY]]` markers.
- Add two guardrail table rows to SKILL.md reinforcing the mandate (Common Mistakes + If You Catch Yourself Thinking…).
- Strengthen `skills/workflow-development/templates/plan-workflow-section.md` header with "do not abridge, summarize, or collapse steps" so the instruction travels with the template itself.
- Add `check_workflow_full_fidelity_mandate()` to `scripts/validate.py` + register in `main()` — static guard that fails if either mandate phrase is ever stripped, locking the fix against regression.

## Test Plan
- [x] `bash scripts/validate.sh` → `All checks passed.`
- [x] `pytest tests/test_validate.py::TestCheckWorkflowFullFidelityMandate -v` → 4/4 pass
- [x] `pytest tests/test_plan_workflow_section.py::TestFullFidelityHeader -v` → 1/1 pass
- [x] `pytest tests/ -v` → 1362/1362 pass
- [x] Acceptance spot-check: strip mandate phrase from SKILL.md → `validate.sh` FAILS with `#455` message → restore → passes

- [x] Reproduce bug: before the fix, Mode A paragraph contained no "in full"/"verbatim" → `check_workflow_full_fidelity_mandate` would have reported a failure (confirmed by RED phase — all new tests failed before implementation).
- [x] Verify fix: after the fix, all 5 new tests pass and the guard bites on stripped files.

Closes #455

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -856,6 +856,65 @@ def check_plan_mode_workflow_embedding():
             )
 
 
+def check_workflow_full_fidelity_mandate():
+    """SKILL.md Mode A must mandate verbatim reproduction of the Workflow template, and
+    the template header must carry 'do not abridge' — locks fix for #455.
+
+    Two invariants checked:
+    1. The '## Plan-Time Behavior (Mode A)' section of workflow-development/SKILL.md
+       must contain both 'in full' and 'verbatim' (the explicit no-summarize mandate).
+    2. The header of templates/plan-workflow-section.md (before the first ```markdown
+       fence) must contain 'do not abridge' (the instruction travels with the template).
+    """
+    skill_md = ROOT / "skills" / "workflow-development" / "SKILL.md"
+    if not skill_md.is_file():
+        fail(Path("skills/workflow-development/SKILL.md"),
+             "missing — cannot check Mode A full-fidelity mandate (#455)")
+        skill_text = None
+    else:
+        try:
+            skill_text = skill_md.read_text(encoding="utf-8")
+        except OSError:
+            fail(skill_md.relative_to(ROOT), "could not read file")
+            skill_text = None
+    if skill_text is not None:
+        mode_a_marker = "## Plan-Time Behavior (Mode A)"
+        idx = skill_text.find(mode_a_marker)
+        if idx >= 0:
+            next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
+            section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
+        else:
+            section = ""
+        if "in full" not in section or "verbatim" not in section:
+            fail(
+                skill_md.relative_to(ROOT),
+                "Mode A paragraph is missing the full-fidelity mandate — the section must "
+                "contain both 'in full' and 'verbatim' to prevent the orchestrator from "
+                "condensing the ## Workflow template (#455).",
+            )
+
+    template_md = ROOT / "skills" / "workflow-development" / "templates" / "plan-workflow-section.md"
+    if not template_md.is_file():
+        fail(Path("skills/workflow-development/templates/plan-workflow-section.md"),
+             "missing — cannot check template full-fidelity header (#455)")
+        tmpl_text = None
+    else:
+        try:
+            tmpl_text = template_md.read_text(encoding="utf-8")
+        except OSError:
+            fail(template_md.relative_to(ROOT), "could not read file")
+            tmpl_text = None
+    if tmpl_text is not None:
+        fence_idx = tmpl_text.find("````markdown")
+        header = tmpl_text[:fence_idx] if fence_idx >= 0 else tmpl_text
+        if "do not abridge" not in header:
+            fail(
+                template_md.relative_to(ROOT),
+                "template header is missing 'do not abridge' — the no-summarize instruction "
+                "must travel with the template so the mandate is visible at the point of use (#455).",
+            )
+
+
 def check_browser_tool_gate(cache=None):
     """Any agent or command referencing browser MCP tools must carry a BLOCKED: sentinel
     and a per-backend install hint — enforces the hard gate from #364.
@@ -929,6 +988,7 @@ def main():
     check_skill_skill_refs(cache=cache)
     check_workflow_development_activation_contract()
     check_plan_mode_workflow_embedding()
+    check_workflow_full_fidelity_mandate()
     check_catalog_completeness(cache=cache)
     check_shared_includes_not_blockquoted(cache=cache)
     check_template_placeholders(cache=cache)
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -889,6 +889,7 @@ def check_workflow_full_fidelity_mandate():
         else:
             next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
             section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
+            # Fail if either token is absent — both are required.
             if "in full" not in section or "verbatim" not in section:
                 fail(
                     skill_md.relative_to(ROOT),
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -863,7 +863,7 @@ def check_workflow_full_fidelity_mandate():
     Two invariants checked:
     1. The '## Plan-Time Behavior (Mode A)' section of workflow-development/SKILL.md
        must contain both 'in full' and 'verbatim' (the explicit no-summarize mandate).
-    2. The header of templates/plan-workflow-section.md (before the first ```markdown
+    2. The header of templates/plan-workflow-section.md (before the first ````markdown
        fence) must contain 'do not abridge' (the instruction travels with the template).
     """
     skill_md = ROOT / "skills" / "workflow-development" / "SKILL.md"
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -868,7 +868,7 @@ def check_workflow_full_fidelity_mandate():
     """
     skill_md = ROOT / "skills" / "workflow-development" / "SKILL.md"
     if not skill_md.is_file():
-        fail(Path("skills/workflow-development/SKILL.md"),
+        fail(skill_md.relative_to(ROOT),
              "missing — cannot check Mode A full-fidelity mandate (#455)")
         skill_text = None
     else:
@@ -880,22 +880,26 @@ def check_workflow_full_fidelity_mandate():
     if skill_text is not None:
         mode_a_marker = "## Plan-Time Behavior (Mode A)"
         idx = skill_text.find(mode_a_marker)
-        if idx >= 0:
-            next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
-            section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
-        else:
-            section = ""
-        if "in full" not in section or "verbatim" not in section:
+        if idx < 0:
             fail(
                 skill_md.relative_to(ROOT),
-                "Mode A paragraph is missing the full-fidelity mandate — the section must "
-                "contain both 'in full' and 'verbatim' to prevent the orchestrator from "
-                "condensing the ## Workflow template (#455).",
+                "section '## Plan-Time Behavior (Mode A)' not found — "
+                "cannot locate Mode A full-fidelity mandate (#455).",
             )
+        else:
+            next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
+            section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
+            if "in full" not in section or "verbatim" not in section:
+                fail(
+                    skill_md.relative_to(ROOT),
+                    "Mode A paragraph is missing the full-fidelity mandate — the section must "
+                    "contain both 'in full' and 'verbatim' to prevent the orchestrator from "
+                    "condensing the ## Workflow template (#455).",
+                )
 
     template_md = ROOT / "skills" / "workflow-development" / "templates" / "plan-workflow-section.md"
     if not template_md.is_file():
-        fail(Path("skills/workflow-development/templates/plan-workflow-section.md"),
+        fail(template_md.relative_to(ROOT),
              "missing — cannot check template full-fidelity header (#455)")
         tmpl_text = None
     else:
--- a/tests/test_plan_workflow_section.py
+++ b/tests/test_plan_workflow_section.py
@@ -125,7 +125,7 @@ class TestFullFidelityHeader:
     def test_header_no_abridge_phrase(self):
         """Header region must carry 'do not abridge' before the fenced markdown block."""
         text = _text()
-        # Slice header: everything before the first ```markdown fence
+        # Slice header: everything before the first ````markdown fence
         fence_idx = text.find("````markdown")
         header = text[:fence_idx] if fence_idx >= 0 else text
         assert "do not abridge" in header, (
--- a/tests/test_plan_workflow_section.py
+++ b/tests/test_plan_workflow_section.py
@@ -6,6 +6,8 @@ Tests assert that:
    cd-prefix anti-pattern).
 2. The template includes a re-anchor/resume note so a continued session knows
    to call EnterWorktree before running commands.
+3. The template header carries a 'do not abridge' instruction so the mandate
+   travels with the template itself (#455).
 """
 
 from pathlib import Path
@@ -110,4 +112,23 @@ class TestReanchorNote:
         assert any(p in text for p in signal_phrases), (
             "Template must state that catching yourself cd-prefixing is the signal "
             "to stop and call EnterWorktree(path=…)."
+        )
+
+
+# ---------------------------------------------------------------------------
+# Full-fidelity header mandate (#455)
+# ---------------------------------------------------------------------------
+
+class TestFullFidelityHeader:
+    """Template header must instruct the orchestrator not to abridge."""
+
+    def test_header_no_abridge_phrase(self):
+        """Header region must carry 'do not abridge' before the fenced markdown block."""
+        text = _text()
+        # Slice header: everything before the first ```markdown fence
+        fence_idx = text.find("````markdown")
+        header = text[:fence_idx] if fence_idx >= 0 else text
+        assert "do not abridge" in header, (
+            "Template header must carry 'do not abridge' so the no-summarize "
+            "instruction travels with the template itself (#455)."
         )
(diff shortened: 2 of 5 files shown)


scripts/validate.py
```diff
@@ -856,6 +856,65 @@ def check_plan_mode_workflow_embedding():
...
+            fail(skill_md.relative_to(ROOT), "could not read file")
+            skill_text = None
+    if skill_text is not None:
+        mode_a_marker = "## Plan-Time Behavior (Mode A)"
+        idx = skill_text.find(mode_a_marker)
+        if idx >= 0:
+            next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
+            section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
+        else:
+            section = ""
```
lugassawan: **Medium** — When the `## Plan-Time Behavior (Mode A)` heading is absent from SKILL.md, `section = ""` and the validator emits the generic "missing the full-fidelity mandate" message without indicating that the heading itself is missing. A deleted or renamed heading looks identical to a mandate that was never written.

**Suggested fix:**
```python
if idx < 0:
    fail(
        skill_md.relative_to(ROOT),
        "section '## Plan-Time Behavior (Mode A)' not found in SKILL.md — "
        "cannot locate Mode A mandate (#455).",
    )
else:
    next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
    section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
    if "in full" not in section or "verbatim" not in section:
        fail(...)
```

tests/test_validate.py
```diff
@@ -2358,3 +2358,64 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
...
+        )
+        (skill_dir / "SKILL.md").write_text(content, encoding="utf-8")
+
+    def _write_template(self, root, header):
+        tmpl_dir = root / "skills" / "workflow-development" / "templates"
+        tmpl_dir.mkdir(parents=True, exist_ok=True)
+        content = header + "\n\n````markdown\n## Workflow\n````\n"
+        (tmpl_dir / "plan-workflow-section.md").write_text(content, encoding="utf-8")
+
+    def test_skill_md_missing_in_full_and_verbatim_fails(self, reset_validate):
```
lugassawan, line 2401: **Medium** — No test exercises the `not skill_md.is_file()` branch at line 870. A typo in the hardcoded `Path("skills/workflow-development/SKILL.md")` string would produce a non-relative failure message and no test would catch it.

**Suggested fix:** add a test that creates neither file and asserts a failure containing `"missing"` or the expected path fragment, mirroring the pattern used by other missing-directory checks in the suite.

scripts/validate.py
```diff
@@ -856,6 +856,65 @@ def check_plan_mode_workflow_embedding():
...
+
+    Two invariants checked:
+    1. The '## Plan-Time Behavior (Mode A)' section of workflow-development/SKILL.md
+       must contain both 'in full' and 'verbatim' (the explicit no-summarize mandate).
+    2. The header of templates/plan-workflow-section.md (before the first ```markdown
+       fence) must contain 'do not abridge' (the instruction travels with the template).
+    """
+    skill_md = ROOT / "skills" / "workflow-development" / "SKILL.md"
+    if not skill_md.is_file():
+        fail(Path("skills/workflow-development/SKILL.md"),
```
lugassawan: **Low** — The missing-file `fail()` call uses `Path("skills/workflow-development/SKILL.md")` (a hardcoded string) rather than `skill_md.relative_to(ROOT)`. The OSError branch directly below uses `skill_md.relative_to(ROOT)`, creating a minor style inconsistency. If ROOT ever changes, the hardcoded path could drift.

**Suggested fix:** `fail(skill_md.relative_to(ROOT), "missing — ...")`

tests/test_validate.py
```diff
@@ -2358,3 +2358,64 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
...
+            "Expected a failure naming 'in full' or 'verbatim' when Mode A mandate is absent"
+        )
+
+    def test_skill_md_missing_verbatim_fails(self, reset_validate):
+        root = reset_validate
+        # Has "in full" but NOT "verbatim"
+        self._write_skill(root, "Reproduce the template in full — substitute [[detect:KEY]] markers.\n")
+        self._write_template(root, "Copy this section — do not abridge.")
+        validate.check_workflow_full_fidelity_mandate()
+        assert any("verbatim" in f for f in validate.FAILURES), (
```
lugassawan: **Low** — The assertion `any("in full" in f or "verbatim" in f ...)` matches any failure message that happens to contain those words, not specifically the mandate-missing failure. Within a single `check_workflow_full_fidelity_mandate()` call this is unlikely to cause false passes, but tightening to the actual message string is more robust.

**Suggested fix:** `assert any("full-fidelity mandate" in f for f in validate.FAILURES)`

scripts/validate.py
lugassawan: Addressed in 903b32331279b1e10cf31ef0e9a56c05522486bd: added explicit `if idx < 0: fail(...)` guard that emits a specific 'section not found' message before falling through to the mandate check.

tests/test_validate.py
lugassawan: Addressed in 903b32331279b1e10cf31ef0e9a56c05522486bd: added `test_skill_md_file_absent_fails` which calls `check_workflow_full_fidelity_mandate()` with no files written and asserts a failure containing 'missing'.

scripts/validate.py
lugassawan: Addressed in 903b32331279b1e10cf31ef0e9a56c05522486bd: both missing-file `fail()` calls now use `skill_md.relative_to(ROOT)` / `template_md.relative_to(ROOT)` — consistent with the OSError branch.

tests/test_validate.py
lugassawan: Addressed in 903b32331279b1e10cf31ef0e9a56c05522486bd: assertion now checks `any("full-fidelity mandate" in f for f in validate.FAILURES)` to match the actual error string.

tests/test_validate.py
```diff
@@ -2358,3 +2358,72 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
...
+            "Expected a failure naming 'verbatim' when that token is missing from Mode A"
```
lugassawan: **Low** — `test_skill_md_missing_in_full_and_verbatim_fails` and `test_skill_md_missing_verbatim_fails` both exercise the same branch (one or both tokens missing); neither tests the case where `verbatim` is present but `in full` is absent. A regression replacing `or` with `and` in the guard condition would not be caught.

**Suggested fix:** add a test for the mirror case, e.g.:
```python
def test_skill_md_missing_in_full_fails(self, reset_validate):
    root = reset_validate
    # Has "verbatim" but NOT "in full"
    self._write_skill(root, "Reproduce verbatim — substitute [[detect:KEY]] markers.\n")
    self._write_template(root, "Copy this section — do not abridge.")
    validate.check_workflow_full_fidelity_mandate()
    assert any("full-fidelity mandate" in f for f in validate.FAILURES)
```

lugassawan: Addressed in 9f77a0ac500df6360c94dc2bca06028213d94f3d: added `test_skill_md_missing_in_full_fails` — exercises the mirror case (has 'verbatim' but not 'in full') so both halves of the `or` guard are independently covered.

scripts/validate.py
```diff
@@ -856,6 +856,69 @@ def check_plan_mode_workflow_embedding():
...
 
 
+def check_workflow_full_fidelity_mandate():
+    """SKILL.md Mode A must mandate verbatim reproduction of the Workflow template, and
+    the template header must carry 'do not abridge' — locks fix for #455.
+
+    Two invariants checked:
+    1. The '## Plan-Time Behavior (Mode A)' section of workflow-development/SKILL.md
+       must contain both 'in full' and 'verbatim' (the explicit no-summarize mandate).
+    2. The header of templates/plan-workflow-section.md (before the first ```markdown
```
lugassawan: **Low** — Docstring says ` ```markdown ` (three backticks) but the code correctly searches for ` ````markdown ` (four backticks). A reader following the docstring to understand the slicing boundary will look for the wrong marker.

**Suggested fix:** update the docstring line to:
```
    2. The header of templates/plan-workflow-section.md (before the first ````markdown
```

tests/test_plan_workflow_section.py
```diff
@@ -111,3 +113,22 @@ def test_cd_prefix_signal_mentioned(self):
...
+# Full-fidelity header mandate (#455)
+# ---------------------------------------------------------------------------
+
+class TestFullFidelityHeader:
+    """Template header must instruct the orchestrator not to abridge."""
+
+    def test_header_no_abridge_phrase(self):
+        """Header region must carry 'do not abridge' before the fenced markdown block."""
+        text = _text()
+        # Slice header: everything before the first ```markdown fence
```
lugassawan: **Low** — Inline comment says `# Slice header: everything before the first` ` ```markdown ` `fence` (three backticks) but the template uses four-backtick fences and the code searches for ` ````markdown `. Creates a false paper trail for anyone debugging a header-slicing failure.

**Suggested fix:**
```python
        # Slice header: everything before the first ````markdown fence
```

tests/test_validate.py
```diff
@@ -2358,3 +2358,83 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
         val.check_agents(cache=cache)
         val.check_agent_skill_refs(cache=cache)
         assert val.FAILURES == [], f"validate.py failures: {val.FAILURES}"
+
+
+# ──────────────────────────────────────────────
+# check_workflow_full_fidelity_mandate
+# ──────────────────────────────────────────────
+
+class TestCheckWorkflowFullFidelityMandate:
```
lugassawan, line 2367: **Low** — `TestCheckWorkflowFullFidelityMandate` has no synthetic clean-pass test (both files correct → `FAILURES == []`). The only clean-path coverage is `test_real_repo_passes`, which couples the unit-test outcome to the real repo's file state. If a future change accidentally adds a spurious `fail()` call on the happy path, no synthetic test would catch it.

**Suggested fix:** add one test:
```python
def test_both_files_correct_passes(self, reset_validate):
    root = reset_validate
    self._write_skill(root, "Reproduce in full and verbatim — substitute [[detect:KEY]] markers.\n")
    self._write_template(root, "Copy this section — do not abridge.")
    validate.check_workflow_full_fidelity_mandate()
    assert validate.FAILURES == [], f"Expected no failures but got: {validate.FAILURES}"
```

scripts/validate.py
lugassawan: Addressed in 417b56a00f96b70c4d128278a2b51a1288004c70: corrected ```markdown to ````markdown in the docstring to match the actual four-backtick fence used in the template.

tests/test_plan_workflow_section.py
lugassawan: Addressed in 417b56a00f96b70c4d128278a2b51a1288004c70: corrected the inline comment from ```markdown to ````markdown to match the template's actual four-backtick fence.

tests/test_validate.py
lugassawan: Addressed in 417b56a00f96b70c4d128278a2b51a1288004c70: added test_both_files_correct_passes as a synthetic clean-pass test before the failure tests — happy path now covered by a fixture-based test, not only by test_real_repo_passes.

tests/test_validate.py
```diff
@@ -2358,3 +2358,90 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
...
+            "Expected a failure containing 'full-fidelity mandate' when Mode A mandate tokens are absent"
+        )
+
+    def test_skill_md_missing_verbatim_fails(self, reset_validate):
+        root = reset_validate
+        # Has "in full" but NOT "verbatim"
+        self._write_skill(root, "Reproduce the template in full — substitute [[detect:KEY]] markers.\n")
+        self._write_template(root, "Copy this section — do not abridge.")
+        validate.check_workflow_full_fidelity_mandate()
+        assert any("verbatim" in f for f in validate.FAILURES), (
```
lugassawan: **Medium** — `test_skill_md_missing_verbatim_fails` asserts `any("verbatim" in f ...)` but the failure message always contains the word "verbatim" regardless of which token triggered it (the shared message reads: "the section must contain both 'in full' and 'verbatim'"). If the guard were accidentally narrowed to check only `"in full"`, this test would still pass — it does not independently verify the verbatim arm.

**Suggested fix:** use the same phrase the other tests assert on, so both missing-token tests are symmetric:
```python
assert any("full-fidelity mandate" in f for f in validate.FAILURES), (
    "Expected a failure containing 'full-fidelity mandate' when 'verbatim' token is absent"
)
```

tests/test_validate.py
```diff
@@ -2358,3 +2358,90 @@ def test_agent_and_skill_ref_checks_pass(self, reset_validate, monkeypatch):
...
+    def test_template_missing_no_abridge_fails(self, reset_validate):
+        root = reset_validate
+        self._write_skill(root, "Reproduce in full and verbatim — substitute [[detect:KEY]] markers.\n")
+        self._write_template(root, "Copy this section into your plan.")
+        validate.check_workflow_full_fidelity_mandate()
+        assert any("do not abridge" in f for f in validate.FAILURES), (
+            "Expected a failure naming 'do not abridge' when that phrase is missing from template header"
+        )
+
+    def test_real_repo_passes(self, reset_validate, monkeypatch):
```
lugassawan, line 2440: **Low** — `val.FAILURES.clear()` is redundant here. The autouse `reset_validate` fixture already clears `FAILURES` before every test body runs — this explicit clear adds noise and may mislead a reader into thinking the monkeypatch leaves dirty state that needs manual cleanup.

**Suggested fix:** remove the `val.FAILURES.clear()` line; the autouse fixture handles it.

scripts/validate.py
```diff
@@ -856,6 +856,69 @@ def check_plan_mode_workflow_embedding():
...
+        if idx < 0:
+            fail(
+                skill_md.relative_to(ROOT),
+                "section '## Plan-Time Behavior (Mode A)' not found — "
+                "cannot locate Mode A full-fidelity mandate (#455).",
+            )
+        else:
+            next_h2 = skill_text.find("\n## ", idx + len(mode_a_marker))
+            section = skill_text[idx:next_h2] if next_h2 >= 0 else skill_text[idx:]
+            if "in full" not in section or "verbatim" not in section:
```
lugassawan, line 893: **Low** — The `or` guard is semantically correct (fail if either token is absent), but a reader expecting a "must contain A and B" check may momentarily expect `and`. A one-line comment makes the intent unambiguous.

**Suggested fix:**
```python
# Fail if either token is absent — both are required.
if "in full" not in section or "verbatim" not in section:
```

tests/test_validate.py
lugassawan: Addressed in 5134b91d267daab7443d70de3c52fe6ee75838d9: changed assertion from `"verbatim" in f` to `"full-fidelity mandate" in f`, matching the actual distinguishing phrase — symmetric with the other missing-token test.

tests/test_validate.py
lugassawan: Addressed in 5134b91d267daab7443d70de3c52fe6ee75838d9: removed the redundant `val.FAILURES.clear()` — the autouse `reset_validate` fixture clears `FAILURES` before each test body runs.

scripts/validate.py
lugassawan: Addressed in 5134b91d267daab7443d70de3c52fe6ee75838d9: added `# Fail if either token is absent — both are required.` above the guard condition.

PR #465, opened 2026-06-23
[chore] Harden CI workflows: SHA-pin actions/*, lockfile guard, release concurrency

### LINKED ISSUE #456: [audit] workflows: pin actions/* to commit SHAs, guard lockfile auto-push, add release concurrency
## Audit findings — workflows

> Scope: `security`, `reliability` · Severity range: `Medium–Low`
> Source: codebase audit (deep · all-domain · 2026-06-17)

These three findings all live under `.github/workflows/` but are independent fixes — split
into sections below; each can be its own PR.

### security

| Severity | File:Line | Symptom | Suggested fix |
|----------|-----------|---------|---------------|
| Medium | `.github/workflows/pr.yml:104` (+`:114,128,129,141,142,155`; `release.yml:18,26,75`; `skill-triggers.yml:17,18`; `dependabot-lockfile.yml:33,41`) | First-party `actions/checkout` & `actions/setup-python` float on mutable `@v6` while third-party actions are full-SHA-pinned | Pin each `@v6` to a commit SHA + `# v6.x.y` comment; add a `github-actions` Dependabot ecosystem so the SHAs receive bump PRs |

**Root cause:** The repo adopted SHA-pinning as a supply-chain control (vendor actions are
pinned with version comments) but exempted GitHub-owned `actions/*`, leaving `@v6` mutable. A
force-moved `v6` tag would execute in the privileged `contents: write` jobs (release publish,
Dependabot auto-push).

**Reasoning chain:** 1. Vendor actions are deliberately full-SHA-pinned with tag comments → the
maintainer knows the pattern. 2. `release.yml:75` and `dependabot-lockfile.yml:33` both checkout
via the floating tag under `contents: write`. 3. Tag mutability is the documented attack vector
(tj-actions/changed-files, 2025). 4. ∴ privileged jobs depend on a mutable reference.

**Counter-evidence considered:** "GitHub-owned actions are lower risk" — true, but the repo's own
standard is SHA-pinning, so the asymmetry is more likely oversight than policy; Dependabot covers
`github-actions`, so pins cost nothing in maintenance.

**security-auditor deep-dive:** Confirmed at **Medium — not downgraded**. Decisive mitigating
fact: **no workflow uses `pull_request_target`** (all use `pull_request`/`push`/`schedule`), so
the catastrophic "fork PR mounts secrets into attacker code" scenario does not exist. Every
privileged step uses the ephemeral repo-scoped `GITHUB_TOKEN` — **no PATs, no registry/signing
secrets** mounted anywhere → max loss capped at repo-scoped write, not org-wide credential theft.
Exploitation needs an upstream `actions/*` namespace compromise (not attacker-reachable via normal
repo interaction) → not High; two `contents: write` jobs make `actions/checkout@v6` the single
highest-leverage CI point → not Low. **Enforcement test for this fix is filed in the companion
`tests` issue.**

### reliability

| Severity | File:Line | Symptom | Suggested fix |
|----------|-----------|---------|---------------|
| Medium | `.github/workflows/dependabot-lockfile.yml:45-58` | Auto-pushes the regenerated lockfile (`contents: write`, `persist-credentials: true`) with no diff / unexpected-package guard | Extract a testable `scripts/check-lockfile-additions.sh` that fails-closed when the diff introduces a previously-absent top-level package |
| Low | `.github/workflows/release.yml:3-6` | Tag-triggered `contents: write` publish path has no `concurrency:` block | Add `concurrency: { group: release-${{ github.ref }}, cancel-in-progress: false }` plus an idempotent `gh release view … || gh release create …` |

**Root cause (dependabot-lockfile):** Dependency resolution is network-fetching and partly
attacker-influenceable (the bumped package's metadata drives what `pip-compile` pulls); its output
is auto-committed with write credentials and no post-resolution integrity gate beyond
`--generate-hashes`.

**Reasoning chain:** 1. The `if` guard (`:29`) correctly restricts to `dependabot[bot]` authors.
2. `--generate-hashes` bakes cryptographic pins for the full transitive closure as resolved on the
runner. 3. The result is auto-pushed under a `github-actions[bot]` signature → lends a malicious pin
false "routine bot housekeeping" provenance that reviewers skim past. 4. With `contents: write`, the
commit lands with no human gate.

**Counter-evidence considered:** "`--generate-hashes` makes it safe" — hashes guarantee
reproducibility of what resolved, not that resolution was benign; they don't detect an unexpected new
package. `pull_request` (not `pull_request_target`) caps escalation, and deps are test-only → Medium,
not High.

**debugger deep-dive (dependabot-lockfile):** REAL, narrower than first stated; confidence HIGH.
Genuine residual risk: the `user.login` (not `github.actor`) gate at `:29` is intentional (#277/#278)
to let humans push fixups — combined with `persist-credentials: true` + `contents: write`, any
collaborator who can push to an open Dependabot branch gets their `tests/*.txt` edits auto-resolved
and auto-committed by the privileged bot job. Not externally exploitable → stays Medium. Recommended
guard: compare top-level package *name* sets pre/post-regen, fail-closed on any previously-absent
name; `bats-core` test the 3 cases (version bump → pass, new package → block, identical → pass).

**Root cause (release concurrency):** Unlike `pr.yml` / `dependabot-lockfile.yml`, the release
workflow has no single-flight guard; `release.sh`'s documented resume paths can re-trigger it.

**Reasoning chain:** 1. `release.yml` and `skill-triggers.yml` lack `concurrency:`. 2.
`skill-triggers.yml` is cron/dispatch → single-flight by nature. 3. `release.yml` is reachable by
any `v*.*.*` push and `release.sh` has re-run semantics. 4. `gh release create` on an existing
release errors → noisy/half-failed release.

**Counter-evidence considered:** Tag pushes are mostly serial → Low not Medium; but the idempotent
re-run design is exactly the scenario that double-triggers, and a guard costs one block.

**debugger deep-dive (release):** REAL but effectively benign; confidence HIGH. `release.sh:348-357`
aborts on a moved tag, and a true same-tag double-fire fails `HTTP 422 already_exists` — `gh release
create` **refuses, doesn't corrupt** → worst case is a spuriously red run, correctly Low. The
idempotency discipline already applied to the *script* should extend to the *workflow*.

## Provenance

Filed by `workflow-audit-emit-issues` from an in-session codebase audit (2026-06-17).
Re-running re-files — check for duplicates before confirming.

lugassawan:
## Summary
- **SHA-pin `actions/*`** — replaced floating `actions/checkout@v7` and `actions/setup-python@v6` tags with full 40-char SHAs + version comments across all four CI workflows (`pr.yml`, `release.yml`, `skill-triggers.yml`, `dependabot-lockfile.yml`). The `github-actions` Dependabot ecosystem (already configured in `.github/dependabot.yml`) will bump the pins automatically going forward.
- **Lockfile-additions guard** — new `scripts/check-lockfile-additions.sh` fails-closed (exit 1 + `::error::` annotation) if a Dependabot lockfile regen introduces a previously-absent top-level package (detected via pip-compile's `# via -r *.txt` blocks). Wired into `dependabot-lockfile.yml` between regen and commit steps. 5 pytest behavioral cases added.
- **Release concurrency + idempotent publish** — `release.yml` now has a `concurrency: cancel-in-progress: false` block (prevents duplicate tag runs from racing) and a `gh release view` idempotency guard on the "Create GitHub Release" step (safe to re-trigger).

## Test Plan
- [x] `shellcheck -S warning scripts/check-lockfile-additions.sh` — 0 issues
- [x] `actionlint` on all 4 edited workflows — 0 issues
- [x] `pytest tests/ -v` — 1370/1370 pass (5 new guard cases: version bump → 0, new top-level → 1, identical → 0, transitive-only → 0, absent-from-HEAD → 1)
- [x] Pre-push hook ran full suite on push — all green

- [x] SHA SHAs verified to resolve to the commented tags before pinning
- [x] Dependabot `github-actions` ecosystem already present — no config change needed
- [x] Guard step positioned after regen but before commit — `HEAD:<lock>` is always the pre-regen baseline

Closes #456

### DIFF
--- a/.github/workflows/dependabot-lockfile.yml
+++ b/.github/workflows/dependabot-lockfile.yml
@@ -46,6 +46,7 @@ jobs:
         run: bash scripts/lock-pip-requirements.sh
 
       - name: Guard against new top-level packages
+        # Keep this lockfile list in sync with the "Regenerate pip lockfiles" step above.
         run: bash scripts/check-lockfile-additions.sh tests/requirements.lock tests/release-requirements.lock tests/build-requirements.lock
 
       - name: Commit and push updated lockfiles if changed
--- a/.github/workflows/dependabot-lockfile.yml
+++ b/.github/workflows/dependabot-lockfile.yml
@@ -30,7 +30,7 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
         with:
           # Checkout the branch tip (not a detached SHA) so git push is a fast-forward.
           # github.head_ref is safe here: action inputs are not shell-evaluated by the runner.
@@ -38,12 +38,15 @@ jobs:
           ref: ${{ github.head_ref }}
           persist-credentials: true
 
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: '3.12'
 
       - name: Regenerate pip lockfiles
         run: bash scripts/lock-pip-requirements.sh
+
+      - name: Guard against new top-level packages
+        run: bash scripts/check-lockfile-additions.sh tests/requirements.lock tests/release-requirements.lock tests/build-requirements.lock
 
       - name: Commit and push updated lockfiles if changed
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   validate-schemas:
     runs-on: ubuntu-latest
@@ -15,7 +19,7 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Validate JSON well-formedness
         run: |
@@ -23,7 +27,7 @@ jobs:
           jq empty .claude-plugin/marketplace.json
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: '3.12'
 
@@ -72,12 +76,14 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Create GitHub Release
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          gh release create "$GITHUB_REF_NAME" \
-            --title "$GITHUB_REF_NAME" \
-            --generate-notes
+          if gh release view "$GITHUB_REF_NAME" >/dev/null 2>&1; then
+            echo "Release $GITHUB_REF_NAME already exists — skipping."
+          else
+            gh release create "$GITHUB_REF_NAME" --title "$GITHUB_REF_NAME" --generate-notes
+          fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          if gh release view "$GITHUB_REF_NAME" >/dev/null 2>&1; then
+          if gh release view "$GITHUB_REF_NAME" >/dev/null; then
             echo "Release $GITHUB_REF_NAME already exists — skipping."
           else
             gh release create "$GITHUB_REF_NAME" --title "$GITHUB_REF_NAME" --generate-notes
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,6 +82,8 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          # gh release view exits 1 when the release doesn't exist; stderr is
+          # intentionally preserved so auth/network errors remain visible.
           if gh release view "$GITHUB_REF_NAME" >/dev/null; then
             echo "Release $GITHUB_REF_NAME already exists — skipping."
           else
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# ── Usage ──────────────────────────────────────────────────────────────────
+
+if [[ $# -eq 0 ]]; then
+  echo "Error: no lockfile paths given" >&2
+  echo "Usage: $0 <lockfile> [<lockfile> ...]" >&2
+  exit 1
+fi
+
+# ── Top-level extractor ────────────────────────────────────────────────────
+# Reads pip-compile lock content on stdin.
+# A package is top-level iff any line in its # via block matches -r *.txt.
+# Outputs sorted package names, one per line.
+
+extract_top_level() {
+  awk '
+    /^[A-Za-z0-9._-]+==/ {
+      name = $1
+      sub(/==.*/, "", name)
+      invia = 0
+      next
+    }
+    /^[[:space:]]*#[[:space:]]*via/ {
+      invia = 1
+      if ($0 ~ /-r[[:space:]].*\.txt/) top[name] = 1
+      next
+    }
+    invia && /^[[:space:]]*#/ {
+      if ($0 ~ /-r[[:space:]].*\.txt/) top[name] = 1
+      next
+    }
+    { invia = 0 }
+    END { for (p in top) if (p != "") print p }
+  ' | sort -u
+}
+
+# ── Per-lockfile check ─────────────────────────────────────────────────────
+
+failed=0
+
+for lock in "$@"; do
+  # Base set: HEAD content (pre-regen, pre-commit).
+  # Empty if the lockfile is new (git cat-file -e exits non-zero).
+  if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
+    base=$(git show "HEAD:${lock}" | extract_top_level)
+  else
+    base=""
+  fi
+
+  # New set: working-tree content (post-regen).
+  new=$(extract_top_level < "${lock}")
+
+  # Additions: packages in new that are absent from base.
+  # Filter empty lines so an empty base doesn't false-positive.
+  added=$(comm -13 \
+    <(printf '%s\n' "${base}" | grep -v '^$' | sort -u) \
+    <(printf '%s\n' "${new}"  | grep -v '^$' | sort -u))
+
+  if [[ -n "${added}" ]]; then
+    echo "::error::${lock}: new top-level package(s) added — review before merging:" >&2
+    while IFS= read -r pkg; do
+      echo "  ${pkg}" >&2
+    done <<< "${added}"
+    failed=1
+  fi
+done
+
+if [[ "${failed}" -eq 1 ]]; then
+  exit 1
+fi
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -61,8 +61,8 @@ for lock in "$@"; do
   # Additions: packages in new that are absent from base.
   # Filter empty lines so an empty base doesn't false-positive.
   added=$(comm -13 \
-    <(printf '%s\n' "${base}" | grep -v '^$' | sort -u) \
-    <(printf '%s\n' "${new}"  | grep -v '^$' | sort -u))
+    <(printf '%s\n' "${base}" | grep -v '^$') \
+    <(printf '%s\n' "${new}"  | grep -v '^$'))
 
   if [[ -n "${added}" ]]; then
     echo "::error::${lock}: new top-level package(s) added — review before merging:" >&2
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -73,6 +73,6 @@ for lock in "$@"; do
   fi
 done
 
-if [[ "${failed}" -eq 1 ]]; then
+if [[ "${failed}" -ne 0 ]]; then
   exit 1
 fi
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -51,6 +51,11 @@ for lock in "$@"; do
   fi
 
   # New set: working-tree content (post-regen).
+  if [[ ! -f "${lock}" ]]; then
+    echo "::error::${lock}: file not found in working tree" >&2
+    failed=1
+    continue
+  fi
   new=$(extract_top_level < "${lock}")
 
   # Additions: packages in new that are absent from base.
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -42,7 +42,8 @@ failed=0
 
 for lock in "$@"; do
   # Base set: HEAD content (pre-regen, pre-commit).
-  # Empty if the lockfile is new (git cat-file -e exits non-zero).
+  # When the lockfile is new (absent from HEAD), every package is treated as an
+  # addition — this is intentional; a new lockfile always requires human review.
   if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
     base=$(git show "HEAD:${lock}" | extract_top_level)
   else
--- a/tests/test_check_lockfile_additions.py
+++ b/tests/test_check_lockfile_additions.py
@@ -137,6 +137,20 @@ class TestLockfileAdditionsGuard:
         assert result.returncode == 1
         assert "file not found" in result.stderr
 
+    def test_all_top_level_removed_exits_zero(self, tmp_path):
+        """Top-level package in HEAD but absent after regen → removal is safe → exit 0."""
+        lock_file = _make_git_repo(tmp_path, _BASE_LOCK)
+        lock_file.write_text("")  # post-regen: empty lock (package removed)
+
+        result = subprocess.run(
+            ["bash", str(GUARD_SCRIPT), "tests/requirements.lock"],
+            cwd=tmp_path,
+            capture_output=True,
+            text=True,
+            env=_CLEAN_ENV,
+        )
+        assert result.returncode == 0, result.stderr
+
     def test_transitive_only_addition_exits_zero(self, tmp_path):
         """A new package with # via <dep> (no -r) is transitive-only → exit 0."""
         lock_file = _make_git_repo(tmp_path, _BASE_LOCK)
--- a/tests/test_check_lockfile_additions.py
+++ b/tests/test_check_lockfile_additions.py
@@ -1,0 +1,127 @@
+"""Behavioral tests for scripts/check-lockfile-additions.sh.
+
+Each case builds a hermetic temp git repo (git init + initial commit of the
+base lock), overwrites the working-tree lock to simulate a post-regen state,
+then runs the script via subprocess and asserts the exit code and output.
+
+The script reads HEAD:<lock> as the pre-regen baseline and compares top-level
+sets (packages whose # via block references -r *.txt).
+"""
+
+import subprocess
+from pathlib import Path
+
+from conftest import _CLEAN_ENV
+
+ROOT = Path(__file__).parent.parent
+GUARD_SCRIPT = ROOT / "scripts" / "check-lockfile-additions.sh"
+
+# Minimal pip-compile lock with one top-level package.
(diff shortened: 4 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -72,12 +76,14 @@ jobs:
...
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Create GitHub Release
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          gh release create "$GITHUB_REF_NAME" \
-            --title "$GITHUB_REF_NAME" \
-            --generate-notes
+          if gh release view "$GITHUB_REF_NAME" >/dev/null 2>&1; then
```
lugassawan: **Medium:** `gh release view … >/dev/null 2>&1` suppresses all stderr, including auth and network failures. If the runner temporarily lacks `contents:write` or GitHub is briefly unreachable, the condition falls through to `gh release create`, which will also fail — but the original root cause (permission/network) is no longer visible in the log.

Consider redirecting only stdout to suppress the "not found" noise while preserving stderr:
```bash
if gh release view "$GITHUB_REF_NAME" --json tagName >/dev/null 2>&1; then
```
Or accept the current behavior — `gh release create` will surface its own error clearly. This is informational; APPROVE stands.

scripts/check-lockfile-additions.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+# ── Per-lockfile check ─────────────────────────────────────────────────────
+
+failed=0
+
+for lock in "$@"; do
+  # Base set: HEAD content (pre-regen, pre-commit).
+  # Empty if the lockfile is new (git cat-file -e exits non-zero).
+  if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
```
lugassawan, line 47: **Medium (design note):** When a brand-new lockfile is committed for the first time (absent from `HEAD`), the guard treats every package in it as a new top-level addition and exits 1 — even if no `-r requirements.txt` line changed. This is a conservative but intentional stance (the test `test_new_lockfile_absent_from_head_exits_one` explicitly covers it).

Consider adding a comment here to make the intent explicit for future maintainers:
```bash
# When the lockfile is new (not yet in HEAD), every package is treated as an
# addition — this is intentional. A human must review the first lockfile commit.
if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
```

.github/workflows/dependabot-lockfile.yml
```diff
@@ -30,21 +30,24 @@ jobs:
...
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: '3.12'
 
       - name: Regenerate pip lockfiles
         run: bash scripts/lock-pip-requirements.sh
 
+      - name: Guard against new top-level packages
+        run: bash scripts/check-lockfile-additions.sh tests/requirements.lock tests/release-requirements.lock tests/build-requirements.lock
```
lugassawan, line 50: **Low:** The guard step hardcodes exactly three lockfile paths (`tests/requirements.lock`, `tests/release-requirements.lock`, `tests/build-requirements.lock`), matching the paths in the "Regenerate pip lockfiles" step above. If a fourth lockfile is ever added to `scripts/lock-pip-requirements.sh`, both steps must be updated independently — easy to miss.

Consider adding a comment:
```yaml
      - name: Guard against new top-level packages
        # Keep lockfile list in sync with the "Regenerate pip lockfiles" step above.
        run: bash scripts/check-lockfile-additions.sh ...
```

.github/workflows/release.yml
lugassawan: Addressed in c598d72c82cfee8de68f1a70ac9d142b5c123bb1: removed `2>&1` so stderr (auth/network errors) now surfaces in the runner log while stdout is still suppressed.

scripts/check-lockfile-additions.sh
lugassawan: Addressed in c598d72c82cfee8de68f1a70ac9d142b5c123bb1: added a two-line comment above the `git cat-file` block explaining that a new lockfile (absent from HEAD) intentionally exits 1 and requires human review.

.github/workflows/dependabot-lockfile.yml
lugassawan: Addressed in c598d72c82cfee8de68f1a70ac9d142b5c123bb1: added a `# Keep this lockfile list in sync` comment above the guard step so the coupling is explicit.

scripts/check-lockfile-additions.sh
```diff
@@ -0,0 +1,73 @@
...
+  # When the lockfile is new (absent from HEAD), every package is treated as an
+  # addition — this is intentional; a new lockfile always requires human review.
+  if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
+    base=$(git show "HEAD:${lock}" | extract_top_level)
+  else
+    base=""
+  fi
+
+  # New set: working-tree content (post-regen).
+  new=$(extract_top_level < "${lock}")
```
lugassawan, line 59: **Medium:** `extract_top_level < "${lock}"` will fail with a shell "No such file or directory" error (and exit 1 via `set -euo pipefail`) if the lockfile path does not exist in the working tree, but the error message does not name the specific file. If the workflow invocation and the regen step ever diverge on a filename, the failure is confusing to triage.

Consider adding an explicit pre-check:
```bash
if [[ ! -f "${lock}" ]]; then
  echo "::error::${lock}: lockfile not found in working tree" >&2
  exit 1
fi
new=$(extract_top_level < "${lock}")
```

tests/test_check_lockfile_additions.py
```diff
@@ -0,0 +1,127 @@
...
+
+    subprocess.run(["git", "init", "-b", "main"], cwd=tmp_path, check=True,
+                   capture_output=True, env=_CLEAN_ENV)
+    subprocess.run(["git", "add", "tests/requirements.lock"], cwd=tmp_path,
+                   check=True, capture_output=True, env=_CLEAN_ENV)
+    subprocess.run(["git", "commit", "-m", "init"], cwd=tmp_path, check=True,
+                   capture_output=True, env=_CLEAN_ENV)
+    return lock_file
+
+
```
lugassawan, line 41: **Low:** The no-arguments usage-error path (lines 7–9 of the script: `exit 1` + usage message) has no test. If that guard were accidentally removed the 5 behavioral tests would still pass. A quick case:
```python
def test_no_args_exits_one(self, tmp_path):
    result = subprocess.run(["bash", str(GUARD_SCRIPT)],
                            cwd=tmp_path, capture_output=True, text=True, env=_CLEAN_ENV)
    assert result.returncode == 1
    assert "Usage" in result.stderr
```

scripts/check-lockfile-additions.sh
lugassawan: Addressed in 2bb9029: added an explicit `[[ -f "${lock}" ]]` guard before the stdin redirect that emits a `::error::` annotation and sets `failed=1; continue` — matching the script's existing error-reporting style.

tests/test_check_lockfile_additions.py
lugassawan: Addressed in 2bb9029: added `test_no_args_exits_one` — invokes the script with no arguments and asserts exit code 1 and `Usage` in stderr. Suite is now 6 cases (was 5).

tests/test_check_lockfile_additions.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        result = subprocess.run(
+            ["bash", str(GUARD_SCRIPT), "tests/requirements.lock"],
+            cwd=tmp_path,
+            capture_output=True,
+            text=True,
+            env=_CLEAN_ENV,
+        )
+        assert result.returncode == 0, result.stderr
```
lugassawan, line 170: **Medium:** The `[[ ! -f "${lock}" ]]` branch (script lines 54–58) has no behavioral test. Passing a misspelled path silently hits this branch — the exit code is correct but the code path is untested. Add a case that passes a non-existent path and asserts `returncode == 1` and `"file not found" in stderr`:

```python
def test_missing_lockfile_exits_one(self, tmp_path):
    """Path given but file absent from working tree → exit 1 with error message."""
    subprocess.run(["git", "init", "-b", "main"], cwd=tmp_path, check=True,
                   capture_output=True, env=_CLEAN_ENV)
    subprocess.run(["git", "commit", "--allow-empty", "-m", "init"],
                   cwd=tmp_path, check=True, capture_output=True, env=_CLEAN_ENV)
    result = subprocess.run(
        ["bash", str(GUARD_SCRIPT), "tests/requirements.lock"],
        cwd=tmp_path, capture_output=True, text=True, env=_CLEAN_ENV,
    )
    assert result.returncode == 1
    assert "file not found" in result.stderr
```

scripts/check-lockfile-additions.sh
```diff
@@ -0,0 +1,78 @@
...
+    failed=1
+    continue
+  fi
+  new=$(extract_top_level < "${lock}")
+
+  # Additions: packages in new that are absent from base.
+  # Filter empty lines so an empty base doesn't false-positive.
+  added=$(comm -13 \
+    <(printf '%s\n' "${base}" | grep -v '^$' | sort -u) \
+    <(printf '%s\n' "${new}"  | grep -v '^$' | sort -u))
```
lugassawan: **Low:** `extract_top_level` already pipes through `sort -u` (line 36), so the `sort -u` calls inside the `comm` process substitutions on lines 64–65 are redundant — two extra sort processes per lockfile, every run. Consider simplifying:

```bash
  added=$(comm -13 \
    <(printf '%s\n' "${base}" | grep -v '^$') \
    <(printf '%s\n' "${new}"  | grep -v '^$'))
```

No behaviour change since `extract_top_level` guarantees sorted unique output.

tests/test_check_lockfile_additions.py
lugassawan: Addressed in b65b83a: added `test_missing_lockfile_exits_one` — scaffolds an empty git repo (no lockfile on disk), passes the path to the script, and asserts exit 1 + "file not found" in stderr. Suite is now 7 cases.

scripts/check-lockfile-additions.sh
lugassawan: Addressed in b65b83a: removed the redundant `sort -u` from both `comm` process substitutions — `extract_top_level` already guarantees sorted unique output via its `| sort -u` pipe, so `comm` receives correctly sorted input without the extra processes.

scripts/check-lockfile-additions.sh
```diff
@@ -0,0 +1,78 @@
...
+  if [[ -n "${added}" ]]; then
+    echo "::error::${lock}: new top-level package(s) added — review before merging:" >&2
+    while IFS= read -r pkg; do
+      echo "  ${pkg}" >&2
+    done <<< "${added}"
+    failed=1
+  fi
+done
+
+if [[ "${failed}" -eq 1 ]]; then
```
lugassawan: **Medium:** `failed -eq 1` is a fragile exit gate — if `failed` were ever set to anything other than exactly `1` (e.g. a future edit turns it into a counter incremented with `((failed++))`), the condition silently passes and the script exits 0 despite failures. Since `failed` is a boolean flag, use `-ne 0` to express intent:

```suggestion
if [[ "${failed}" -ne 0 ]]; then
```

tests/test_check_lockfile_additions.py
```diff
@@ -0,0 +1,156 @@
...
+    tests_dir = tmp_path / "tests"
+    tests_dir.mkdir()
+    lock_file = tests_dir / "requirements.lock"
+    lock_file.write_text(lock_content)
+
+    subprocess.run(["git", "init", "-b", "main"], cwd=tmp_path, check=True,
+                   capture_output=True, env=_CLEAN_ENV)
+    subprocess.run(["git", "add", "tests/requirements.lock"], cwd=tmp_path,
+                   check=True, capture_output=True, env=_CLEAN_ENV)
+    subprocess.run(["git", "commit", "-m", "init"], cwd=tmp_path, check=True,
```
lugassawan, line 37: **Medium:** `_CLEAN_ENV` sets `GIT_CONFIG_NOSYSTEM=1` (strips system gitconfig) but still reads `~/.gitconfig`. If a developer has GPG commit signing enabled globally (`commit.gpgsign=true`), every `git commit` spawned by these tests fails with a signing error — breaking local runs and any CI runner where the bot's git identity has signing configured. The fix belongs in `conftest.py` at the `_CLEAN_ENV` definition so all tests pick it up automatically:

```python
_CLEAN_ENV: Final[MappingProxyType[str, str]] = MappingProxyType(
    {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
    | {
        "GIT_CONFIG_NOSYSTEM": "1",
        "GIT_AUTHOR_NAME": "T",
        "GIT_AUTHOR_EMAIL": "t@t.com",
        "GIT_COMMITTER_NAME": "T",
        "GIT_COMMITTER_EMAIL": "t@t.com",
        "GIT_CONFIG_COUNT": "1",
        "GIT_CONFIG_KEY_0": "commit.gpgsign",
        "GIT_CONFIG_VALUE_0": "false",
    }
)
```

Alternatively pass `-c commit.gpgsign=false` in every `git commit` invocation, but patching `conftest.py` is the single-fix-point approach.

.github/workflows/release.yml
```diff
@@ -72,12 +76,14 @@ jobs:
...
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Create GitHub Release
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          gh release create "$GITHUB_REF_NAME" \
-            --title "$GITHUB_REF_NAME" \
-            --generate-notes
+          if gh release view "$GITHUB_REF_NAME" >/dev/null; then
```
lugassawan, line 87: **Low (informational):** When a release does not yet exist, `gh release view` prints an error to stderr (e.g. `release not found`) — stdout is redirected to `/dev/null` but stderr is intentionally preserved (correct, since suppressing stderr would also hide auth/network errors). This is the right tradeoff, but the stray error line on first-time releases may confuse readers of the run log since the `else` branch is the happy path. Consider adding a clarifying `echo` before the guard:

```yaml
        run: |
          # gh release view exits 1 when the release doesn't exist; stderr is
          # intentionally preserved so auth/network errors remain visible.
          if gh release view "$GITHUB_REF_NAME" >/dev/null; then
            echo "Release $GITHUB_REF_NAME already exists — skipping."
          else
            gh release create "$GITHUB_REF_NAME" --title "$GITHUB_REF_NAME" --generate-notes
          fi
```

No behaviour change needed — current implementation is correct.

tests/test_check_lockfile_additions.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+
+        result = subprocess.run(
+            ["bash", str(GUARD_SCRIPT), "tests/requirements.lock"],
+            cwd=tmp_path,
+            capture_output=True,
+            text=True,
+            env=_CLEAN_ENV,
+        )
+        assert result.returncode == 0, result.stderr
```
lugassawan, line 170: **Low:** No test covers the case where all top-level packages are removed between HEAD and the working tree (i.e. `base` is non-empty, `new` is empty → `added` is empty → exit 0). Removals are safe and the script correctly exits 0 here, but an explicit test would document this intent and catch a regression if the set-difference logic changes:

```python
def test_all_top_level_removed_exits_zero(self, tmp_path):
    """Top-level package present in HEAD but absent after regen → removal is safe → exit 0."""
    lock_file = _make_git_repo(tmp_path, _BASE_LOCK)
    lock_file.write_text("")  # post-regen: empty lock (package removed)

    result = subprocess.run(
        ["bash", str(GUARD_SCRIPT), "tests/requirements.lock"],
        cwd=tmp_path,
        capture_output=True,
        text=True,
        env=_CLEAN_ENV,
    )
    assert result.returncode == 0, result.stderr
```

scripts/check-lockfile-additions.sh
lugassawan: Addressed in de3b0e1d565691198226ff3e0666d16d116e9987: changed `-eq 1` to `-ne 0` — expresses boolean intent and is robust if `failed` ever becomes a counter.

tests/test_check_lockfile_additions.py
lugassawan: Addressed in de3b0e1d565691198226ff3e0666d16d116e9987: added `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0` to `_CLEAN_ENV` in `conftest.py` to suppress `commit.gpgsign` via git's transient-config mechanism — covers all tests from a single fix point without touching individual call sites.

.github/workflows/release.yml
lugassawan: Addressed in de3b0e1d565691198226ff3e0666d16d116e9987: added the clarifying comment above the `gh release view` guard explaining that stderr is intentionally preserved so auth/network errors remain visible.

tests/test_check_lockfile_additions.py
lugassawan: Addressed in de3b0e1d565691198226ff3e0666d16d116e9987: added `test_all_top_level_removed_exits_zero` — commits a lockfile with one top-level package, writes an empty post-regen lock (package removed), and asserts exit 0. Suite is now 8 cases.

PR #466, opened 2026-06-23
[fix] Phase 4 dispatches both reviewers in parallel (#458)

### LINKED ISSUE #458: [bug] Review phase sometimes dispatches only one reviewer instead of both
## Surface affected
- [x] Skill — `workflow-*`
- [x] Slash command

**Specific surface:** `workflow-development` Phase 4 (Review); `/swe-workbench:implement`

## Claude Code version
N/A — instruction-level defect, version-independent

## Plugin version
0.1.20

## Reproduction steps
1. Run `/swe-workbench:implement` (or any `workflow-development` flow) through to Phase 4: Review.
2. Observe the dispatch in Phase 4.
3. Sometimes only one reviewer runs instead of both.

## Expected behavior
Phase 4 dispatches BOTH reviewers, in parallel, as two distinct required invocations:
- `superpowers:requesting-code-review` (a Skill — plan-alignment), and
- `swe-workbench:reviewer` (a subagent — diff correctness/security/design).

## Actual behavior
Sometimes only one is dispatched. Suspected root causes (from code investigation):
1. The two "reviewers" are different units — one is a Skill, one is a subagent. "Dispatch both
   reviewers" (workflow-development/SKILL.md:202) can be satisfied by running the skill inline and
   dispatching only the one real subagent.
2. The word "parallel" never appears in Phase 4; nothing forces a single concurrent batch.
3. The skip-condition (workflow-development/SKILL.md:210) — "if Phase 2 sub-skill already ran
   two-stage review with evidence, proceed" — can swallow one or both dispatches.

## Suggested fix
- Rewrite workflow-development/SKILL.md:202 to state: dispatch BOTH reviewers IN PARALLEL, in a
  single batch, as two distinct required invocations (Skill + subagent) — neither optional.
- Mirror the same wording in commands/implement.md:36.
- Tighten the skip-condition at workflow-development/SKILL.md:210 so it cannot silently drop a
  reviewer (require explicit evidence that BOTH stages ran).

## Environment (optional)
N/A

lugassawan:
## Summary

- Fixes issue #458: Phase 4 review step now unambiguously dispatches **both** reviewers **in parallel**, as two distinct required invocations, **neither optional**
- Adds `TestPhase4DispatchesBothReviewers` regression guard in `tests/test_validate.py` — slices each file's Phase 4 section and asserts it contains `superpowers:requesting-code-review`, `swe-workbench:reviewer`, and `parallel` (TDD red → green)
- Updates all 4 dispatch sites with explicit Skill-vs-subagent type labels and parallel-dispatch wording: `skills/workflow-development/SKILL.md`, `commands/implement.md`, `skills/workflow-development/templates/plan-workflow-section.md`, `skills/workflow-extend/SKILL.md`
- Updates flow diagram in `workflow-development/SKILL.md` to show `BOTH in parallel (neither optional)` with `├─` / `└─` ASCII branches
- Tightens skip-condition to require named evidence from **both** stages; one-sided review no longer qualifies
- Fixes `check_command_skill_refs` and `check_agent_skill_refs` in `scripts/validate.py` to use `_artifact_exists()` (resolves skill dirs, agent files, and command files), consistent with how `check_skill_skill_refs` already works

## Test Plan

- [x] `bash scripts/validate.sh` — passes
- [x] `pytest tests/ -v` — 1366/1366 pass (including new `TestPhase4DispatchesBothReviewers`)
- [x] Confirmed test goes RED before doc edits, GREEN after (TDD discipline)

- [x] Root defect reproduced: no `parallel` token at any Phase 4 site before the fix
- [x] All 3 contributing defects addressed: ambiguous wording, missing `parallel`, over-broad skip-condition
- [x] No regression in `TestCheckAgentSkillRefs` / `TestCheckCommandSkillRefs` after validator fix

Closes #458

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -271,8 +271,8 @@ def check_agent_skill_refs(cache=None):
             if not _artifact_exists(artifact_id):
                 fail(
                     agent_md.relative_to(ROOT),
-                    f"references 'swe-workbench:{artifact_id}' but skills/{artifact_id}/ does not exist "
-                    f"(also checked agents/{artifact_id}.md and commands/{artifact_id}.md)",
+                    f"references 'swe-workbench:{artifact_id}' but no matching artifact found "
+                    f"(checked skills/{artifact_id}/, agents/{artifact_id}.md, commands/{artifact_id}.md)",
                 )
 
 
@@ -287,8 +287,8 @@ def check_command_skill_refs():
             if not _artifact_exists(artifact_id):
                 fail(
                     cmd_md.relative_to(ROOT),
-                    f"references 'swe-workbench:{artifact_id}' but skills/{artifact_id}/ does not exist "
-                    f"(also checked agents/{artifact_id}.md and commands/{artifact_id}.md)",
+                    f"references 'swe-workbench:{artifact_id}' but no matching artifact found "
+                    f"(checked skills/{artifact_id}/, agents/{artifact_id}.md, commands/{artifact_id}.md)",
                 )
 
 
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -254,9 +254,9 @@ def check_commands():
 
 
 def check_agent_skill_refs(cache=None):
-    """Every `swe-workbench:<id>` in agents/*.md must resolve to skills/<id>/ on disk."""
+    """Every `swe-workbench:<id>` in agents/*.md must resolve to a skill dir,
+    agent file, or command file on disk."""
     agents_dir = ROOT / "agents"
-    skills_dir = ROOT / "skills"
     agents_cache = cache[0] if cache is not None else None
     pattern = re.compile(r'`swe-workbench:([\w-]+)`')
     for agent_md in sorted(agents_dir.glob("*.md")):
@@ -267,26 +267,28 @@ def check_agent_skill_refs(cache=None):
                 continue
         else:
             text = agent_md.read_text(encoding="utf-8")
-        for skill_id in set(pattern.findall(text)):
-            if not (skills_dir / skill_id).is_dir():
+        for artifact_id in set(pattern.findall(text)):
+            if not _artifact_exists(artifact_id):
                 fail(
                     agent_md.relative_to(ROOT),
-                    f"references 'swe-workbench:{skill_id}' but skills/{skill_id}/ does not exist",
+                    f"references 'swe-workbench:{artifact_id}' but skills/{artifact_id}/ does not exist "
+                    f"(also checked agents/{artifact_id}.md and commands/{artifact_id}.md)",
                 )
 
 
 def check_command_skill_refs():
-    """Every `swe-workbench:<id>` in commands/*.md must resolve to skills/<id>/ on disk."""
+    """Every `swe-workbench:<id>` in commands/*.md must resolve to a skill dir,
+    agent file, or command file on disk."""
     commands_dir = ROOT / "commands"
-    skills_dir = ROOT / "skills"
     pattern = re.compile(r'`swe-workbench:([\w-]+)`')
     for cmd_md in sorted(commands_dir.glob("*.md")):
         text = cmd_md.read_text(encoding="utf-8")
-        for skill_id in set(pattern.findall(text)):
-            if not (skills_dir / skill_id).is_dir():
+        for artifact_id in set(pattern.findall(text)):
+            if not _artifact_exists(artifact_id):
                 fail(
                     cmd_md.relative_to(ROOT),
-                    f"references 'swe-workbench:{skill_id}' but skills/{skill_id}/ does not exist",
+                    f"references 'swe-workbench:{artifact_id}' but skills/{artifact_id}/ does not exist "
+                    f"(also checked agents/{artifact_id}.md and commands/{artifact_id}.md)",
                 )
 
 
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -662,7 +662,27 @@ class TestCheckAgentSkillRefs:
             encoding="utf-8",
         )
         validate.check_agent_skill_refs()
-        assert any("does not exist" in f for f in validate.FAILURES)
+        assert any("no matching artifact found" in f for f in validate.FAILURES)
+
+    def test_ref_to_existing_agent_file_passes(self, reset_validate):
+        """An agent referencing another agent via swe-workbench: must pass
+        when the target exists as agents/<id>.md (not just skills/<id>/)."""
+        root = reset_validate
+        make_plugin_tree(root)
+        agents_dir = root / "agents"
+        agents_dir.mkdir(parents=True, exist_ok=True)
+        (agents_dir / "bar.md").write_text(
+            "---\nname: bar\ndescription: d\ntools: Read\n---\n",
+            encoding="utf-8",
+        )
+        (agents_dir / "my-agent.md").write_text(
+            "---\nname: my-agent\ndescription: d\ntools: Read, Skill\n---\n"
+            "\nUse `swe-workbench:bar` subagent.\n"
+            "\n> See @./shared/principles.md\n",
+            encoding="utf-8",
+        )
+        validate.check_agent_skill_refs()
+        assert len(validate.FAILURES) == 0
 
 
 # ──────────────────────────────────────────────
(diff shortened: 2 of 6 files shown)


scripts/validate.py
```diff
@@ -267,26 +267,28 @@ def check_agent_skill_refs(cache=None):
                 continue
         else:
             text = agent_md.read_text(encoding="utf-8")
-        for skill_id in set(pattern.findall(text)):
-            if not (skills_dir / skill_id).is_dir():
+        for artifact_id in set(pattern.findall(text)):
+            if not _artifact_exists(artifact_id):
```
lugassawan, line 271: **Medium** — The new `_artifact_exists()` fallback path (agent file + command file) in `check_agent_skill_refs` and `check_command_skill_refs` is not exercised by the existing unit tests. `TestCheckAgentSkillRefs` and `TestCheckCommandSkillRefs` only create a `skills/<id>/` dir as the valid artifact.

**Why it matters:** If either function's agent-file or command-file branch regresses silently, the tests will still pass, defeating the purpose of the validator extension.

**Suggested fix:** Add one test to each class that creates `agents/<id>.md` (or `commands/<id>.md`) instead of a skill dir and asserts no failure is reported:

```python
def test_ref_to_existing_agent_file_passes(self, reset_validate):
    root = reset_validate
    make_plugin_tree(root)
    (root / 'agents' / 'bar.md').write_text(
        '---\nname: bar\ndescription: d\ntools: Read\n---\n', encoding='utf-8'
    )
    (root / 'agents' / 'my-agent.md').write_text(
        '---\nname: my-agent\ndescription: d\ntools: Read, Skill\n---\n'
        '\nUse `swe-workbench:bar` subagent.\n'
        '\n> See @./shared/principles.md\n',
        encoding='utf-8',
    )
    validate.check_agent_skill_refs()
    assert len(validate.FAILURES) == 0
```

scripts/validate.py
```diff
@@ -267,26 +267,28 @@ def check_agent_skill_refs(cache=None):
...
     for cmd_md in sorted(commands_dir.glob("*.md")):
         text = cmd_md.read_text(encoding="utf-8")
-        for skill_id in set(pattern.findall(text)):
-            if not (skills_dir / skill_id).is_dir():
+        for artifact_id in set(pattern.findall(text)):
+            if not _artifact_exists(artifact_id):
                 fail(
                     cmd_md.relative_to(ROOT),
-                    f"references 'swe-workbench:{skill_id}' but skills/{skill_id}/ does not exist",
+                    f"references 'swe-workbench:{artifact_id}' but skills/{artifact_id}/ does not exist "
```
lugassawan: **Low** — The error message primary clause says `skills/<id>/ does not exist` even when `_artifact_exists()` checked agent and command paths too. If this fires for a genuine agent or command mismatch, the author will look in `skills/` first.

**Suggested fix:**
```python
f"references 'swe-workbench:{artifact_id}' but no matching artifact found "
f"(checked skills/{artifact_id}/, agents/{artifact_id}.md, commands/{artifact_id}.md)",
```
(Same change needed at line 274 in `check_agent_skill_refs`.)

scripts/validate.py
lugassawan: Addressed in 5f0bcf2: added `test_ref_to_existing_agent_file_passes` to both `TestCheckAgentSkillRefs` and `TestCheckCommandSkillRefs`, each creating an `agents/<id>.md` target and asserting no failure.

scripts/validate.py
lugassawan: Addressed in 5f0bcf2: rephrased error message to `'no matching artifact found (checked skills/<id>/, agents/<id>.md, commands/<id>.md)'` — neutral phrasing that no longer implies only the skill dir was consulted. Three test assertions updated to match.

tests/test_validate.py
```diff
@@ -662,7 +662,27 @@ def test_ref_to_absent_skill_fails(self, reset_validate):
             encoding="utf-8",
         )
         validate.check_agent_skill_refs()
-        assert any("does not exist" in f for f in validate.FAILURES)
+        assert any("no matching artifact found" in f for f in validate.FAILURES)
+
+    def test_ref_to_existing_agent_file_passes(self, reset_validate):
```
lugassawan, line 667: **Low** — `agents_dir.mkdir(parents=True, exist_ok=True)` is redundant here: `make_plugin_tree(root)` already creates `agents/` unconditionally, so this call is a no-op. It is harmless but slightly misleading — implying the directory might not exist when it always does after `make_plugin_tree`.

**Suggested fix:** Remove the `agents_dir = ...` and `agents_dir.mkdir(...)` lines and reference `root / 'agents'` directly where needed. (Same pattern applies to the pre-existing instance at line 643 in `test_ref_to_absent_skill_fails`, though that is out of scope for this PR.)

tests/test_validate.py
```diff
@@ -2444,3 +2480,80 @@ def test_real_repo_passes(self, reset_validate, monkeypatch):
...
+        ),
+    ]
+
+    REQUIRED_TOKENS = [
+        "superpowers:requesting-code-review",
+        "swe-workbench:reviewer",
+        "parallel",
+    ]
+
+    def _extract_phase4_section(self, text: str, start_marker: str, terminator: str) -> str:
```
lugassawan, line 2527: **Low** — When a Phase 4 start marker is renamed or deleted, `_extract_phase4_section` raises a `ValueError` (test *error*) rather than a clean assertion failure. This makes the failure harder to triage — a test error looks like an infrastructure problem, not a content drift.

**Suggested fix:** Return an empty string when the marker is not found; the caller's `missing` list will then include all three required tokens and produce an informative assertion message:

```python
def _extract_phase4_section(self, text: str, start_marker: str, terminator: str) -> str:
    start = text.find(start_marker)
    if start == -1:
        return ""  # all tokens will register as missing → clear failure message
    end = text.find(terminator, start + len(start_marker))
    return text[start:] if end == -1 else text[start:end]
```

tests/test_validate.py
```diff
@@ -2444,3 +2480,80 @@ def test_real_repo_passes(self, reset_validate, monkeypatch):
...
+            return text[start:]
+        return text[start:end]
+
+    def test_all_dispatch_sites_name_both_reviewers_and_parallel(self):
+        """Each Phase 4 section must contain both reviewer identifiers and 'parallel'."""
+        failures = []
+        for rel_path, start_marker, terminator in self.SITES:
+            full_path = self.REAL_ROOT / rel_path
+            text = full_path.read_text(encoding="utf-8")
+            section = self._extract_phase4_section(text, start_marker, terminator)
```
lugassawan, line 2543: **Low** — When `_extract_phase4_section` raises `ValueError` (missing Phase-4 start marker), the exception propagates unhandled out of the `for` loop in `test_all_dispatch_sites_name_both_reviewers_and_parallel`. Instead of a clean assertion failure with the diagnostic `"Phase 4 dispatch sites are missing required tokens"` message, the test runner surfaces a raw `ValueError` traceback, making it harder to distinguish a "marker renamed" failure from a real Python error.

*(Distinct from the already-open thread at line 2529 about `_extract_phase4_section`'s design — this is about the call site not catching the exception.)*

**Suggested fix:**
```python
for rel_path, start_marker, terminator in self.SITES:
    full_path = self.REAL_ROOT / rel_path
    text = full_path.read_text(encoding="utf-8")
    try:
        section = self._extract_phase4_section(text, start_marker, terminator)
    except ValueError as exc:
        failures.append(f"{rel_path}: {exc}")
        continue
    section_lower = section.lower()
    ...
```

_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

tests/test_validate.py
lugassawan: Addressed in 11aab1e02096ce94ea64289e801bf4de6f7af16f: removed redundant `agents_dir` variable and `mkdir` call; writing directly to `root / 'agents' / '<file>.md'` instead.

tests/test_validate.py
lugassawan: Addressed in 11aab1e02096ce94ea64289e801bf4de6f7af16f: changed `_extract_phase4_section` to return `""` when the start marker is not found. An empty string causes all required tokens to register as missing, producing the clean assertion failure message instead of a raw `ValueError` traceback.

tests/test_validate.py
lugassawan: Resolved by the Thread #2 fix — `_extract_phase4_section` now returns `""` instead of raising `ValueError`, so there is no exception to catch at the call site. The `try/except` suggested here is no longer needed.

PR #467, opened 2026-06-23
[feat] Enforce tell-don't-ask and member-ordering conventions

### LINKED ISSUE #459: [feat] Enforce behaviour-method (tell-don't-ask) and member-ordering conventions in reviewer/implementer
## Goal
Generated/reviewed code should automatically (a) move logic onto the domain object that owns the
data (tell-don't-ask / no anemic domain) and (b) order members `public → protected → private`,
without the user having to prompt for either.

## Motivation
Today both conventions are missing or unenforced, so they require a manual follow-up prompt:
- Behaviour-method / anemic-domain: the only explicit, reviewer-reachable rule is
  principle-solid/SKILL.md:72 ("Anemic domain. Move behavior onto the entity."). principle-ddd is
  the conceptual owner but never names tell-don't-ask/anemic, AND principle-ddd is not on the
  reviewer's (agents/reviewer.md:88-104) or implementer's (agents/code-impl.md:60-72) invoke list.
- Member ordering `public → protected → private`: no normative rule exists anywhere — only
  incidental prose in two principle-clean-architecture example files
  (examples/mvc.ts.md:6, examples/mvc.java.md:6) that no review/implement agent reliably loads.
Root cause is wiring + missing normative text, not "wrong skill."

## Surface
- [x] Skill — `principle-*`
- [x] Subagent

## Sketch (optional)
- Add an explicit tell-don't-ask / "move behaviour onto the owning object; avoid anemic domain"
  rule to principle-ddd (keep principle-solid:72).
- Add a `public → protected → private` member-ordering rule to principle-clean-code.
- Wire principle-ddd (and the member-ordering rule's home) into the reviewer + code-impl invoke
  lists so they apply without manual prompting.

## Acceptance (optional)
- [ ] principle-ddd states the tell-don't-ask / anemic-domain rule explicitly.
- [ ] principle-clean-code states the member-ordering convention as a normative rule.
- [ ] agents/reviewer.md and agents/code-impl.md invoke-lists include the skills carrying both rules.
- [ ] A generated/reviewed change applies both without an extra user prompt.

lugassawan:
## Summary
- Add normative tell-don't-ask / anemic-domain rule to `skills/principle-ddd/SKILL.md` Entity subsection — names both **tell-don't-ask** and **anemic domain model** explicitly so reviewer and implementer agents can surface the anti-pattern without a manual follow-up prompt
- Add new `## Member ordering` section to `skills/principle-clean-code/SKILL.md` with a modifier-aware `public → protected → private` rule, plus a note covering Go, Rust, and Python (modifier-less languages)
- Wire `swe-workbench:principle-ddd` into the `## Principle consultation` list of both `agents/reviewer.md` and `agents/code-impl.md` so the skill is conditionally invoked when entity-related code is in scope
- Add `tests/test_behaviour_and_ordering_conventions.py` (TDD red→green): 11 tests asserting all 4 acceptance criteria

## Test Plan
- [x] `pytest tests/test_behaviour_and_ordering_conventions.py -v` → RED before changes, GREEN after (7 failures → 0)
- [x] `pytest tests/ -v` → 1376/1376 pass, no regression in `test_review_routing.py` / `test_skill_triggers.py`

- [x] New rule text verified against the plan's normative phrasing requirements (both `tell` and `anemic` present in DDD skill)
- [x] `grep -n "principle-ddd" agents/reviewer.md agents/code-impl.md` confirms both bullets present
- [x] `grep -n "Member ordering" skills/principle-clean-code/SKILL.md` confirms section added after `## Naming reveals intent`
- [x] Cross-reference to `principle-solid` § "If You Catch Yourself Thinking…" updated per code review (medium finding)
- [x] Word-boundary regex fix applied in test for modifier-less language check (low finding)

Closes #459

### DIFF
--- a/tests/test_behaviour_and_ordering_conventions.py
+++ b/tests/test_behaviour_and_ordering_conventions.py
@@ -53,7 +53,7 @@ def test_ddd_skill_file_exists():
 def test_ddd_skill_names_tell_dont_ask():
     """Entity subsection must explicitly use the tell-don't-ask idiom."""
     body = DDD_SKILL.read_text()
-    assert re.search(r"tell[,\-]?\s*don\W?t\s*ask", body, re.IGNORECASE), (
+    assert re.search(r"tell\W?\s*don\W?t\W?\s*ask", body, re.IGNORECASE), (
         "skills/principle-ddd/SKILL.md must mention the 'tell, don't ask' principle "
         "so reviewers and implementers can surface the anti-pattern"
     )
--- a/tests/test_behaviour_and_ordering_conventions.py
+++ b/tests/test_behaviour_and_ordering_conventions.py
@@ -1,0 +1,146 @@
+"""Structural tests for tell-don't-ask + member-ordering conventions (closes #459).
+
+Acceptance criteria:
+- skills/principle-ddd/SKILL.md body names both 'tell' (don't-ask) and 'anemic' domain.
+- skills/principle-clean-code/SKILL.md has a '## Member ordering' section that states
+  'public → protected → private' and mentions modifier-less languages (Go, Rust, or Python).
+- agents/reviewer.md Principle consultation list references swe-workbench:principle-ddd.
+- agents/code-impl.md Principle consultation list references swe-workbench:principle-ddd.
+"""
+
+import re
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+
+DDD_SKILL = ROOT / "skills" / "principle-ddd" / "SKILL.md"
+CLEAN_CODE_SKILL = ROOT / "skills" / "principle-clean-code" / "SKILL.md"
+REVIEWER_AGENT = ROOT / "agents" / "reviewer.md"
+CODE_IMPL_AGENT = ROOT / "agents" / "code-impl.md"
+
+
+def _section(body: str, heading: str) -> str:
+    """Extract body of a ## heading, stopping at the next ## heading (skips fenced blocks)."""
+    marker = f"## {heading}"
+    assert marker in body, f"Expected '{marker}' section not found"
+    start = body.index(marker) + len(marker)
+    rest = body[start:]
+    fence_open = False
+    lines = []
+    for line in rest.splitlines():
+        stripped = line.strip()
+        if stripped.startswith("```") or stripped.startswith("~~~~"):
+            fence_open = not fence_open
+        if not fence_open and line.startswith("## "):
+            break
+        lines.append(line)
+    return "\n".join(lines)
+
+
+# ---------------------------------------------------------------------------
+# AC 1 — principle-ddd names tell-don't-ask and anemic domain
+# ---------------------------------------------------------------------------
+
+
+def test_ddd_skill_file_exists():
+    assert DDD_SKILL.exists(), "skills/principle-ddd/SKILL.md must exist"
+
+
+def test_ddd_skill_names_tell_dont_ask():
+    """Entity subsection must explicitly use 'tell' (covering tell-don't-ask)."""
+    body = DDD_SKILL.read_text()
+    assert "tell" in body.lower(), (
+        "skills/principle-ddd/SKILL.md must mention 'tell' (tell-don't-ask principle) "
+        "so reviewers and implementers can surface the anti-pattern"
+    )
+
+
+def test_ddd_skill_names_anemic_domain():
+    """Skill body must explicitly name 'anemic' domain model as an anti-pattern."""
+    body = DDD_SKILL.read_text()
+    assert "anemic" in body.lower(), (
+        "skills/principle-ddd/SKILL.md must mention 'anemic' (anemic domain model) "
+        "so reviewers and implementers can surface the anti-pattern"
+    )
+
+
+# ---------------------------------------------------------------------------
+# AC 2 — principle-clean-code has a Member ordering section
+# ---------------------------------------------------------------------------
+
+
+def test_clean_code_skill_file_exists():
+    assert CLEAN_CODE_SKILL.exists(), "skills/principle-clean-code/SKILL.md must exist"
+
+
+def test_clean_code_has_member_ordering_section():
+    body = CLEAN_CODE_SKILL.read_text()
+    assert "## Member ordering" in body, (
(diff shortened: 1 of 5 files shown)


tests/test_behaviour_and_ordering_conventions.py
```diff
@@ -0,0 +1,146 @@
...
+DDD_SKILL = ROOT / "skills" / "principle-ddd" / "SKILL.md"
+CLEAN_CODE_SKILL = ROOT / "skills" / "principle-clean-code" / "SKILL.md"
+REVIEWER_AGENT = ROOT / "agents" / "reviewer.md"
+CODE_IMPL_AGENT = ROOT / "agents" / "code-impl.md"
+
+
+def _section(body: str, heading: str) -> str:
+    """Extract body of a ## heading, stopping at the next ## heading (skips fenced blocks)."""
+    marker = f"## {heading}"
+    assert marker in body, f"Expected '{marker}' section not found"
```
lugassawan: **Medium** — `_section` raises `AssertionError` instead of failing the calling test cleanly.

`assert marker in body` inside a shared helper points the traceback at `_section:25`, not the calling test, making CI failures harder to diagnose. Consider returning `""` when the heading is absent and letting each caller assert with its own message:

```python
def _section(body: str, heading: str) -> str:
    marker = f"## {heading}"
    if marker not in body:
        return ""  # caller asserts with its own message
    start = body.index(marker) + len(marker)
    ...
```

tests/test_behaviour_and_ordering_conventions.py
```diff
@@ -0,0 +1,146 @@
...
+# ---------------------------------------------------------------------------
+
+
+def test_ddd_skill_file_exists():
+    assert DDD_SKILL.exists(), "skills/principle-ddd/SKILL.md must exist"
+
+
+def test_ddd_skill_names_tell_dont_ask():
+    """Entity subsection must explicitly use 'tell' (covering tell-don't-ask)."""
+    body = DDD_SKILL.read_text()
```
lugassawan, line 55: **Low** — `"tell" in body.lower()` would pass if the word "tell" appeared anywhere in the file for any reason. Tighten to anchor on the full idiom:

```python
assert re.search(r"tell[,\-–]?\s*don't\s*ask", body, re.IGNORECASE), (
    "skills/principle-ddd/SKILL.md must mention the tell-don't-ask principle"
)

File: tests/test_behaviour_and_ordering_conventions.py [hunk 3]
```diff
@@ -0,0 +1,146 @@
...
+    body = CLEAN_CODE_SKILL.read_text()
+    assert "## Member ordering" in body, (
+        "skills/principle-clean-code/SKILL.md must contain a '## Member ordering' section "
+        "so the convention has a canonical normative home"
+    )
+
+
+def test_member_ordering_states_public_protected_private():
+    """The section must name the canonical order: public → protected → private."""
+    body = CLEAN_CODE_SKILL.read_text()
```
lugassawan, line 90: **Medium** — `test_member_ordering_states_public_protected_private` (and the sibling modifier-less-languages test) call `_section` but don't guard against an empty return before checking content.

If `_section` returns `""` (heading absent), the `has_order` check fails — but the message says `"must state the ordering…"` rather than `"section is missing"`, obscuring the real cause. Adding a non-empty guard makes failures self-diagnosing:

```python
def test_member_ordering_states_public_protected_private():
    body = CLEAN_CODE_SKILL.read_text()
    section = _section(body, "Member ordering")
    assert section, "'## Member ordering' section is empty or missing"
    # ... rest of assertions
```

tests/test_behaviour_and_ordering_conventions.py
lugassawan: Addressed in 94dff21: `_section` now returns `""` when heading is absent; each calling test asserts on the return value with its own message.

tests/test_behaviour_and_ordering_conventions.py
lugassawan: Addressed in 94dff21: replaced `"tell" in body.lower()` with `re.search(r"tell[,\-]?\s*don\W?t\s*ask", body, re.IGNORECASE)` — uses `\W?` to match either apostrophe style, and anchors on the full idiom rather than the bare word "tell".

tests/test_behaviour_and_ordering_conventions.py
lugassawan: Addressed in 94dff21: added `assert section, "'## Member ordering' section is empty or missing"` as the first guard in both `test_member_ordering_states_public_protected_private` and `test_member_ordering_mentions_modifier_less_languages`.

tests/test_behaviour_and_ordering_conventions.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_ddd_skill_file_exists():
+    assert DDD_SKILL.exists(), "skills/principle-ddd/SKILL.md must exist"
+
+
+def test_ddd_skill_names_tell_dont_ask():
+    """Entity subsection must explicitly use the tell-don't-ask idiom."""
+    body = DDD_SKILL.read_text()
+    assert re.search(r"tell[,\-]?\s*don\W?t\s*ask", body, re.IGNORECASE), (
```
lugassawan: **Medium** — Tell-don't-ask regex does not cover the hyphenated form `tell-don't-ask`.

`tell[,\-]?\s*don\W?t\s*ask` matches the current comma form `"tell, don't ask"` in the skill — but fails on `"tell-don't-ask"`, which appears verbatim in both agent files (`agents/reviewer.md:104`, `agents/code-impl.md:73`). If a future edit reformats the skill to the hyphen form the test silently breaks while the content remains correct.

Suggested fix — extend to be form-agnostic:
```python
re.search(r"tell\W?\s*don\W?t\W?\s*ask", body, re.IGNORECASE)
```
This matches `"tell, don't ask"`, `"tell-don't-ask"`, and `"tell don't ask"` without false positives.

lugassawan: Addressed in c52ce10: widened regex to `tell\W?\s*don\W?t\W?\s*ask` — now matches both `tell, don't ask` (comma) and `tell-don't-ask` (hyphenated) and any similar variant.

PR #469, opened 2026-06-23
[fix] Switch between worktrees via ExitWorktree→retry, not bare cd

lugassawan:
## Summary
- Route `EnterWorktree` rejection (session already in a different worktree) to `ExitWorktree(action=keep)` → return to main → retry, instead of bare `cd`
- Demote `cd` to a last resort only for non-rimba checkouts with no `.claude/worktrees` infrastructure
- Apply the two-branch remedy consistently across four instruction blocks: `workflow-worktree-session/SKILL.md` (step 2, active remedy, quick-ref table), `workflow-development/SKILL.md` (Phase 1 Enter), and `plan-workflow-section.md` (Enter-worktree bullet + Resume note)
- Regression tests written first (RED), skill edits second (GREEN); pre-existing `TestNoCdEquivalence` remains green

## Test Plan
- [x] `pytest tests/test_worktree_session_skill.py tests/test_plan_workflow_section.py -v` — 21/21 pass (4 new tests, 17 pre-existing)
- [x] Full suite `pytest` — 1393/1393 pass

- [x] Root cause identified: skill text routed switch-between-worktrees case to `cd` fallback instead of `ExitWorktree(keep)`+retry
- [x] Verified new tests fail against old text (RED) and pass after edits (GREEN)
- [x] Confirmed `Do not auto-exit` rule (SKILL.md lines 74-76) is unchanged and not contradicted by the new guidance

Issue: N/A — corrective guidance fix discovered during live orchestrator session; no open issue


skills/workflow-development/SKILL.md
```diff
@@ -151,7 +151,7 @@ Use the service scope whenever the work is clearly contained within one module 
...
+**Enter the worktree:** After `Path: <abs-path>` appears, try `EnterWorktree(path=<abs-path>)`. From the main session this works for any git-registered worktree (including rimba's `../<repo>-worktrees/` layout). If rejected because the session is already inside a **different worktree** (target path outside `.claude/worktrees/`), the primary remedy is `ExitWorktree(action=keep)` → return to main → retry `EnterWorktree(path=<abs-path>)` (re-anchors session caches). Fall back to `cd <abs-path>` via Bash only as a last resort for non-rimba checkouts with no `.claude/worktrees` infrastructure; `cd` only anchors the Bash persistent cwd and does not re-anchor session-level caches the way `EnterWorktree` does. On any resumed or continued session, try `EnterWorktree(path=<worktree-path>)` first; if rejected because the session is already inside a different worktree, call `ExitWorktree(action=keep)` → retry; otherwise re-`cd <worktree-path>` as a last resort (Bash cwd does not persist across session resume, so `cd` must be re-issued each time — unlike `EnterWorktree`, which restores it automatically).
```
lugassawan: **Medium — Resume-tail `cd` fallback is missing the "non-rimba only" qualifier**

The `ExitWorktree`→retry clause is correctly qualified, but the `otherwise re-\`cd\`` branch in the same sentence lacks the "for non-rimba checkouts only" constraint present in every other `cd` fallback sentence in this PR. An orchestrator in a rimba-managed repo could silently fall back to `cd` after an ExitWorktree retry failure without knowing it is contraindicated.

**Suggested fix:**
```
On any resumed or continued session, try `EnterWorktree(path=<worktree-path>)` first; if rejected because the session is already inside a different worktree, call `ExitWorktree(action=keep)` → retry; otherwise re-`cd <worktree-path>` as a last resort for non-rimba checkouts only (Bash cwd does not persist across session resume, so `cd` must be re-issued each time — unlike `EnterWorktree`, which restores it automatically).
```

tests/test_worktree_session_skill.py
```diff
@@ -102,3 +102,27 @@ def test_forbidden_pattern_mentions_lock_contract(self):
...
+    not a bare cd, when EnterWorktree is rejected because the session is already
+    inside a different worktree."""
+
+    def test_exit_then_retry_is_primary_remedy(self):
+        text = _text()
+        assert "ExitWorktree(action=keep)" in text or 'ExitWorktree(action: "keep")' in text
+
+    def test_retry_enter_worktree_after_exit(self):
+        text = _text().lower()
+        assert "retry" in text and "enterworktree" in text
```
lugassawan: **Low — `test_retry_enter_worktree_after_exit` passes vacuously on the word "enterworktree"**

The test asserts `"retry" in text and "enterworktree" in text.lower()`. The second condition is always true as long as `EnterWorktree` appears *anywhere* in the file — not specifically in the switch remedy. If the retry instruction were removed while `EnterWorktree` remained elsewhere, this test would still pass.

**Suggested tighter assertion:**
```python
def test_retry_enter_worktree_after_exit(self):
    text = _text()
    switch_section = text[text.find("different worktree"):text.find("Do not auto-exit")]
    assert "retry" in switch_section.lower() and "EnterWorktree" in switch_section
```

tests/test_plan_workflow_section.py
```diff
@@ -132,3 +132,18 @@ def test_header_no_abridge_phrase(self):
...
+# ---------------------------------------------------------------------------
+
+class TestSwitchRemedyInTemplate:
+    """Template Enter-worktree bullet must prescribe ExitWorktree+retry as primary
+    and demote cd to a last resort."""
+
+    def test_exit_worktree_retry_remedy_present(self):
+        assert "ExitWorktree" in _text()
+
+    def test_cd_is_last_resort(self):
```
lugassawan, line 151: **Low — `TestSwitchRemedyInTemplate` assertions are file-wide, not scoped to the Enter-worktree bullet**

Both tests assert that `"ExitWorktree"` / `"last resort"` appear *anywhere* in the template. Since the Exit bullet already contained `ExitWorktree(action=keep)` before this PR, a future refactor that removes the new Enter-worktree remedy while keeping the Exit bullet would still pass these tests.

**Suggested tighter assertions:**
```python
def test_exit_worktree_retry_remedy_present(self):
    text = _text()
    enter_bullet = text[text.find("Enter worktree"):text.find("Resume note")]
    assert "ExitWorktree" in enter_bullet

def test_cd_is_last_resort(self):
    text = _text()
    enter_bullet = text[text.find("Enter worktree"):text.find("Resume note")]
    assert "last resort" in enter_bullet.lower()
```

skills/workflow-development/SKILL.md
lugassawan: Addressed in 8142b5831b472968b7325b4f6148e08167df23cc: appended "for non-rimba checkouts only" to the resume-tail `cd` fallback clause.

tests/test_worktree_session_skill.py
lugassawan: Addressed in 8142b5831b472968b7325b4f6148e08167df23cc: `test_retry_enter_worktree_after_exit` now scopes its assertion to the switch_section paragraph (between 'different worktree' and 'Do not auto-exit').

tests/test_plan_workflow_section.py
lugassawan: Addressed in 8142b5831b472968b7325b4f6148e08167df23cc: both `TestSwitchRemedyInTemplate` tests now slice to the Enter-worktree bullet before asserting.

PR #470, opened 2026-06-23
[test] Assert every workflow uses: is SHA-pinned (#457)

### LINKED ISSUE #457: [audit] tests: add SHA-pin assertion to workflow hardening test suite
## Audit findings — tests

> Scope: `testing` · Severity range: `Low–Low`
> Source: codebase audit (deep · all-domain · 2026-06-17)

### testing

| Severity | File:Line | Symptom | Suggested fix |
|----------|-----------|---------|---------------|
| Low | `tests/test_workflow_hardening.py:1` | The hardening test validates `timeout-minutes` + concurrency, but nothing asserts the `@<40-hex>` SHA-pin pattern — a future downgrade of a vendor action to `@v2` stays green | Add a parametrized test asserting every non-`actions/*` `uses:` line matches `@[0-9a-f]{40}` |

**Root cause:** The "vendor actions must be SHA-pinned" invariant lives only in a review convention
and an inline comment, not an executable assertion — so it is unenforced once the originating author
moves on.

**Reasoning chain:** 1. `grep` of `tests/*.py` for vendor action names / the 40-hex pattern returns
nothing. 2. The repo pattern is to mirror every CI invariant into a Python test
(`test_pr_validation.py`, `test_workflow_hardening.py`). 3. SHA-pinning is the one hardening property
with no test. 4. ∴ it can silently regress.

**Counter-evidence considered:** "Dependabot edits these lines, so a human reviews" — review is not a
deterministic gate and Dependabot PRs are often fast-merged; the rest of the hardening surface is
test-enforced precisely because review alone was deemed insufficient.

**Coupling:** This test enforces the fix tracked in the companion `[audit] workflows: security &
reliability findings` issue (security section). File/fix both together for a complete change.

## Provenance

Filed by `workflow-audit-emit-issues` from an in-session codebase audit (2026-06-17).
Re-running re-files — check for duplicates before confirming.

lugassawan:
## Summary
- Adds `test_every_uses_is_sha_pinned` to `tests/test_workflow_hardening.py`, parametrized over all workflow files
- Two module-level regexes: `USES_RE` (captures the action ref, stripping inline `# vX.Y.Z` comments) and `SHA_PIN_RE` (asserts `@<40hex>`, case-insensitive)
- Skips local composite refs (`./`) and `docker://` refs (no git ref to pin); guards against false matches from `run:` heredocs via `"/" not in ref` check
- All 16 existing `uses:` lines across 5 workflow files already pass; the guard fails the moment any action is downgraded to a tag or branch

## Test Plan

- [ ] `pytest tests/ -v` → 1392 passed (includes 5 new `test_every_uses_is_sha_pinned[<workflow>]` cases)

### Type-tailored checks ([test])
- [x] TDD red: temporarily replaced `lycheeverse/lychee-action@<sha>` with `@v2` → `test_every_uses_is_sha_pinned[pr.yml]` failed listing `lycheeverse/lychee-action@v2`
- [x] TDD green: reverted pin → all 5 parametrized cases pass
- [x] `bash scripts/validate.sh` → passes
- [x] Full suite `pytest tests/ -v` → 1392 passed, no regressions in `test_every_job_has_timeout_minutes` / `test_pr_yml_has_concurrency_guard`

Closes #457


tests/test_workflow_hardening.py
```diff
@@ -56,6 +56,38 @@ def test_every_job_has_timeout_minutes(filename):
     )
 
 
+# Matches any uses: line and captures the action ref (stops at whitespace, so a trailing
+# "# v7.0.0" comment is excluded). Handles both "- uses:" and bare "uses:" step forms.
+USES_RE = re.compile(r"^\s*(?:-\s+)?uses:\s*(\S+)", re.MULTILINE)
```
lugassawan, line 64: **Low** — `USES_RE` scans the full file text with `re.MULTILINE`, so a `run:` step whose heredoc or comment line contains `uses: org/repo@tag` would be matched and evaluated. The `"/" not in ref` guard already reduces blast radius, and no current workflow has this pattern. Worth noting this limitation in a comment if the guard is ever relaxed.

tests/test_workflow_hardening.py
```diff
@@ -56,6 +56,38 @@ def test_every_job_has_timeout_minutes(filename):
...
+    for match in USES_RE.finditer(text):
+        ref = match.group(1)
+        # Local composite actions (./path) and docker refs have no git ref to pin — skip.
+        if ref.startswith("./") or ref.startswith("docker://"):
+            continue
+        # Skip values without "/" — real action refs are "owner/repo@..."; bare words can
+        # appear inside run: heredoc blocks that happen to contain a "uses:" line.
+        if "/" not in ref:
+            continue
+        if not SHA_PIN_RE.search(ref):
```
lugassawan, line 87: **Low** — `SHA_PIN_RE.search(ref)` works correctly (the `$` anchor and `\S+` capture prevent false matches), but the intent is clearer with a comment noting that `.search()` plus end-anchor is semantically equivalent to `.fullmatch()` for this purpose. Alternatively, document it as is — the existing comment on `SHA_PIN_RE` already explains the pattern well.

tests/test_workflow_hardening.py
lugassawan: Addressed in 1717f443285e63ccfa89d19631a0d5ad904da294: expanded the USES_RE comment to document the run: heredoc limitation and explain why the `"/" not in ref` guard covers the common case.

tests/test_workflow_hardening.py
lugassawan: Addressed in 1717f443285e63ccfa89d19631a0d5ad904da294: added an inline comment at the .search() call site noting that .search() + end-anchor ($) is semantically equivalent to fullmatch for this ref string.

PR #480, opened 2026-07-02
[feat] Add paired-guard symmetry heuristic to reviewer

### LINKED ISSUE #472: [feat] Reviewer should compare newly-added guard methods against their sibling/paired guards
## Goal

When a diff adds or modifies a guard/eligibility/validation method, the `reviewer` agent should
identify its sibling/paired method that implements the same conceptual check and compare their
predicate sets — flagging any predicate present in one but missing in the other.

## Motivation

In a multi-round PR review, round 1 caught the headline concerns (a stale-read / replica-lag
issue and a method-contract change) but under-scrutinized a newly-added guard method's
*eligibility completeness*. The new guard omitted one predicate that its paired sibling guard
enforced — concretely, a producer-side guard and a consumer-side guard for the same business
rule had diverged. The divergence was only caught in round 2, after a follow-up commit triggered
a focused re-read and the reviewer explicitly placed the two methods side-by-side.

This class of bug is high-signal and easy to miss without an explicit prompt to look for it.
Paired methods are common: producer-guard vs. consumer-guard, `validate` vs. `apply`,
`canX` vs. `shouldX`, optimistic-check vs. authoritative-check. When a diff touches one half of
such a pair, the predicate sets should match (or the divergence should be intentional and
documented).

A read-only check of the current plugin confirms the heuristic is absent:
- `agents/reviewer.md` Process section covers generic "understand callers and blast radius"
  but never says to compare a new method against an existing sibling. Line 6 actively
  discourages reading related files "just in case," which can work against this.
- `skills/principle-code-review/SKILL.md` Four-Axis Review Lens (Correctness / Design integrity)
  is the natural home but names no such pattern.

## Surface

- [x] Subagent — `reviewer`
- [x] Skill — `principle-code-review`

## Sketch

Add a Process step to the reviewer agent and a bullet under the Four-Axis Review Lens, e.g.:

> **Paired-guard symmetry.** When the diff adds or changes a guard / eligibility / validation
> method, grep for its sibling (producer↔consumer, `validate`↔`apply`, `canX`↔`shouldX`).
> Compare the predicate sets. Flag any predicate enforced by one side but not the other as a
> potential completeness gap — unless the divergence is intentional and documented in code.

This is a targeted grep-then-compare, consistent with the agent's existing "strategic, not blind"
guidance — it does not require binge-reading related files.

## Acceptance

- [ ] Reviewing a diff that adds/modifies a guard method, the reviewer locates sibling methods
      implementing the same conceptual check and compares predicate sets.
- [ ] A predicate present in one sibling but absent in the other is flagged (or the divergence
      is confirmed intentional + documented).
- [ ] Guidance lands in `agents/reviewer.md` and/or `skills/principle-code-review/SKILL.md`.

---
_Reported via `/swe-workbench:report-issue` — plugin v0.1.21, Claude Code 2.1.186._

lugassawan:
## Summary

- Add a "Paired-guard symmetry" step to `agents/reviewer.md`'s Process (new step 7, renumbering the old "Diff-size-aware path" step to 8): when a diff adds/changes a guard/eligibility/validation method, grep for its sibling (producer↔consumer, `validate`↔`apply`, `canX`↔`shouldX`) and compare predicate sets, flagging any predicate enforced on one side but missing on the other unless the divergence is intentional and documented.
- Extend the **Correctness** bullet of `skills/principle-code-review/SKILL.md`'s Four-Axis Review Lens to name paired-guard predicate gaps, without introducing a fifth axis.
- Add `tests/test_reviewer_agent.py` and `tests/test_principle_code_review_skill.py` (mirroring the existing `_read`/`_section` helper pattern from `tests/test_debugger_agent.py`) asserting the new guidance is present.

## Test Plan

- [x] `pytest tests/ -v` — 1404/1404 pass (includes 6 new structural tests, written RED before the implementation, GREEN after)

- [x] New reviewer heuristic wording and placement verified against every acceptance criterion in #472 (independently confirmed by a plan-alignment review agent that ran the new tests against the pre-change file content to prove they genuinely fail without the change)
- [x] Both `superpowers:requesting-code-review` and `swe-workbench:reviewer` diff reviews returned zero Critical/Important findings; three Low/Minor nits (unused test helper, fragile numeric cross-reference, confidence-floor interaction) were fixed and re-verified

Closes #472


agents/reviewer.md
```diff
@@ -17,7 +17,8 @@ You are a senior code reviewer. Your job is to catch the issues a careful collea
...
 5. Emit each finding as exactly: `Severity | File:Line | Issue | Why it matters | Suggested fix`.
 6. **Strategic, not blind.** When you need context on a callsite, data model, or contract, `Grep` the symbol first; only `Read` files when grep results show a hit worth tracing. Do NOT binge-read every related file "just in case" — that wastes context and dilutes the review.
-7. **Diff-size-aware path.** Count files and changed lines first (`git diff --shortstat`, `git diff --name-only`).
+7. **Paired-guard symmetry.** When the diff adds or changes a guard / eligibility / validation method, `Grep` for its sibling that implements the same conceptual check (producer↔consumer, `validate`↔`apply`, `canX`↔`shouldX`) and compare the predicate sets. Flag any predicate enforced by one side but not the other as a completeness gap, subject to the confidence floor from step 0 — unless the divergence is intentional and documented in code. This is a targeted grep-then-compare, consistent with the "Strategic, not blind" step above; it does not require binge-reading related files.
```
lugassawan: **Low** | The new step still cross-references `step 0` by number ("subject to the confidence floor from step 0") even though the immediately-preceding fix commit on this branch replaced a sibling numeric cross-reference ("consistent with step 6" → "consistent with the 'Strategic, not blind' step above") specifically to avoid silent breakage on reorder.

**Why it matters:** If `## Process` is ever renumbered (a step inserted above "Load heuristics"), this reference silently points to the wrong step — the exact failure mode the sibling fix was written to prevent, applied inconsistently within the same paragraph.

**Suggested fix:** `subject to the confidence floor from the "Load heuristics" step` instead of `from step 0`.

lugassawan: Addressed in bd0c3e3e5009b0193e1e579e6b53277ed19c8584: replaced the numeric `step 0` cross-reference with a name-based one ("Load heuristics" step), matching the same fix already applied to the sibling "step 6" reference.

PR #483, opened 2026-07-02
[fix] Refresh stale-but-valid pins in lockfile generate mode

lugassawan:
## Summary
- `scripts/lock-pip-requirements.sh`'s `compile()` generate branch compiled `pip-compile` with `--output-file` pointed directly at the existing committed lockfile, so pip-compile treated still-valid pins as a resolution seed and kept them even after they'd drifted from upstream — unlike the `--check` branch, which compiles into a fresh empty `mktemp` and is always a clean re-resolution.
- Consequence: the script's own error message ("Run scripts/lock-pip-requirements.sh to regenerate") was a no-op for stale-but-valid drift, and `.github/workflows/dependabot-lockfile.yml`'s auto-commit step (same code path) could silently miss transitive version bumps.
- Fix: add `--upgrade` to the generate-branch `pip-compile` call, forcing a fresh re-resolution that matches `--check`'s behavior. Fixing the shared `compile()` function repairs both callers — no workflow edit needed.
- Note: this makes generate-mode runs always perform a full upgrade sweep (not an incremental compile), so Dependabot auto-commits and local `regenerate` runs may now show broader transitive-pin diffs than before — this is the intended behavior per #481, not a side effect.

## Test Plan
- [ ]

- [x] Reproduced: added `TestGenerateRefreshesStalePins`, confirmed it fails against the unmodified script (stale pin `fakepkg==1.0.0` survives)
- [x] Verified: after adding `--upgrade`, the same test passes (`fakepkg==1.1.0` replaces the stale pin)
- [x] `pytest tests/test_lock_pip_requirements.py -v` — 15 passed, including the pre-existing `TestCheckDriftGate` (unaffected)
- [x] `pytest tests/ -v` — 1405 passed (full suite)
- [x] `shellcheck --severity=warning scripts/lock-pip-requirements.sh` — clean

Closes #481


scripts/lock-pip-requirements.sh
```diff
@@ -84,7 +87,11 @@ compile() {
     echo "$out is up to date."
   else
     # Run from REPO_ROOT so pip-compile records relative paths in the header.
+    # --upgrade forces a fresh re-resolution: without it, pip-compile treats the
+    # existing --output-file as a pin seed and keeps stale-but-still-valid pins
+    # (the --check branch avoids this by compiling into an empty mktemp). See #481.
     (cd "$REPO_ROOT" && "$VENV_DIR/bin/pip-compile" \
```
lugassawan, line 98: **Low**

`--upgrade` is now unconditional in generate mode, so every local invocation of the script (not just CI's dependabot-lockfile job) will re-resolve *all* transitive pins to latest-allowed on every run, not just the ones affected by a source-file change. A developer who edits `tests/requirements.txt` to add one package and runs the script locally may see unrelated transitive version bumps in the diff.

This is explicitly the ticket's (#481) suggested fix and matches the CI use case it targets, so non-blocking — optional: mention in `--help` output or a script comment that generate mode always performs a full re-resolution.

tests/test_lock_pip_requirements.py
```diff
@@ -89,8 +113,34 @@ def _make_repo(tmp_path: Path) -> dict:
...
         "fi",
     ]
+    # Upgrade-aware stub for generate-mode tests (issue #481): parses
+    # --output-file and --upgrade, and only overwrites the output when
+    # --upgrade is present — modeling pip-compile keeping a seeded pin
+    # otherwise.  Reads the "latest resolution" content from a separate
+    # fixture file (never `grep committed > committed`: in generate mode
+    # --output-file *is* the committed lock, so redirecting into it would
+    # truncate it before it could be read).
+    pip_compile_upgrade_lines = [
```
lugassawan: **Low**

`pip_compile_lines` and `pip_compile_upgrade_lines` are two near-duplicate stub scripts (arg-parsing loop, `--output-file` handling) that could be merged into one stub parameterized by an env var, since they're never both needed for a single test run.

Minor duplication; doesn't affect correctness or test isolation, and keeping them separate does make each stub's semantics easier to read in isolation. Optional: consider a single stub script gated by an env flag (e.g. `UPGRADE_AWARE=1`) if a third pip-compile behavior variant is ever needed — not worth touching otherwise.

scripts/lock-pip-requirements.sh
lugassawan: Addressed in f1670670413166a87ba2b8eb4e6b9cd1e67b6ec8: added a --help note and an inline script comment clarifying that generate mode always performs a full --upgrade re-resolution.

tests/test_lock_pip_requirements.py
lugassawan: Addressed in f1670670413166a87ba2b8eb4e6b9cd1e67b6ec8: merged the two near-duplicate pip-compile test stubs into a single script gated by an UPGRADE_AWARE env var, dropping the now-unneeded upgrade_aware param on _make_repo.

PR #489, opened 2026-07-02
[feat] Add synthesis mode to /report-issue blank-arg routing

### LINKED ISSUE #475: [feat] report-issue: synthesize accumulated memory into ranked enhancement insights
## Problem

`/report-issue`'s blank-argument routing surfaces up to 3 *raw* candidate thoughts from conversation + memory, but never aggregates across the accumulated `feedback_*`/`project_*` entries. Recurring pain points and themes spanning many entries stay invisible, so the highest-value plugin enhancements are easy to miss.

## Value

Maintainers get themed, ranked enhancement insights synthesized across *all* accumulated memory — turning the memory store into a self-improvement signal instead of a flat 3-item list.

## Acceptance criteria

- In the synthesis mode, `/report-issue` loads user/project memory (`MEMORY.md` + symlinked `feedback_*`/`project_*`) and the current conversation, and aggregates them into themed swe-workbench enhancement insights.
- Insights are presented as a **ranked digest with preview**; the user picks which to file (one or several). The existing confirm-before-file gate is preserved.
- Borrows the grouped-preview pattern from `workflow-audit-emit-issues`; reuses the memory path-derivation recipe already in `report-issue` (:20).
- Filing targets the plugin repo, as report-issue does today.

## Impact / Effort

**Impact: M/L** — Converts latent memory into actionable roadmap signal. Recurring themes across feedback entries become visible; maintainers can rank enhancements by prevalence and impact.

**Effort: M** — Synthesis layer over the existing scan + grouped-preview reuse. Reuses memory path logic (:20) and preview pattern from `workflow-audit-emit-issues`; no new infrastructure.

## Additional context

**Related decisions:**
- `/capture` routing (feedback_capture_routing.md): bugs→debugger, features→product-manager. This enhancement strengthens the product-manager path.
- Memory recipe (commands/report-issue.md:20): path-derivation logic already in place; synthesis layer builds on top.
- Grouped preview pattern (skills/workflow-audit-emit-issues/SKILL.md): preview grouped findings before filing; user picks which to file.

**Implementation notes:**
- The synthesis mode runs when `/report-issue` is invoked with no arguments (existing blank-argument routing).
- Aggregation strategy: parse all memory entries, cluster by theme (e.g., "CI/automation", "template discovery", "labeling"), rank by prevalence + recency, present top 5–7 insights with the option to file one or several.
- Preserve the existing confirm-before-file gate; do not auto-file.

lugassawan:
## Summary

- Add a Step 0a mode selector to `/report-issue`'s blank-argument routing: quick pick (existing 3-candidate behavior, wrapped verbatim under a "Branch A" heading) vs. synthesize (new "Branch B")
- Branch B aggregates all `MEMORY.md` + linked `feedback_*.md`/`project_*.md` entries and the current conversation, clusters them into emergent semantic themes (not a fixed taxonomy), ranks by prevalence + recency + conversation-touch boost, and keeps the top 5-7 as candidate enhancement insights
- Two-turn pick-then-confirm flow: Turn 1 prints a ranked digest (no filing), the user picks numbers, Turn 2 shows a final preview of only the picked insights (still no filing) with redaction/label/dup-scan applied, and a literal `confirm` files one GitHub issue per pick via a numbered-tempfile + single `.cmd` sidecar pattern, reusing the existing delegation block's auth check, redaction pass, version footer, and duplicate scan
- Both required reviewers (code-review skill + `swe-workbench:reviewer` subagent) ran in parallel; fixed their High/Medium findings (a `metadata.type` schema mismatch against real memory files, and a template-discovery/body-shape contradiction) plus two Low findings (partial-failure temp-file cleanup, missing no-premature-filing test coverage)

## Test Plan

- [x] `python3 -m pytest tests/test_report_issue_command.py -q` — 24/24 pass (13 pre-existing + 11 new synthesis-mode assertions), written RED-first
- [x] `pytest tests/` — 1445/1445 pass (full suite, CI parity)
- [x] `bash scripts/validate.sh` — 0 issues; `plugin.json`/`marketplace.json` untouched (no version bump)

- [x] New behavior (synthesize mode) manually traced end-to-end via the command's own text: mode selector -> Branch B steps 1-8 -> edge cases, confirming an LLM agent could follow it unambiguously

Closes #475


commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
+
+2. **Harvest conversation signal.** Separately note which plugin-related themes the current conversation itself touches. Keep this as its own set — it feeds the recency boost in step 4, it is not merged into the memory set.
+
+3. **Cluster into emergent themes.** Group the loaded entries by the semantic theme carried in their `description`/body — clusters emerge from the content itself. This is NOT a fixed taxonomy and not a path-prefix grouping (memory entries have no paths); invent a short theme label per cluster rather than picking from a predefined list. Each cluster becomes one candidate insight.
+
+4. **Rank.** Order candidate insights by **prevalence** (cluster size) first; break ties by **recency** — an entry's position in `MEMORY.md` order — with a boost for any cluster the conversation also touches per step 2. Keep the top **5–7** insights.
```
lugassawan: **Medium**: Step 4's "Keep the top 5–7 insights" has no floor carve-out for the case where clustering genuinely produces fewer than 5 themes from more than 2 raw entries (the only documented floor exception is "1–2 entries — too few to cluster," a different condition). E.g. 6 memory entries that cluster into 3 distinct themes leave the model with a written instruction to "keep top 5–7" but only 3 available — it may pad with weak or duplicate insights to satisfy the stated range, contradicting "clusters emerge from the content itself" (step 3).

**Why it matters:** The digest could present artificially inflated/duplicated themes to hit a floor that isn't actually achievable, undermining the acceptance criterion that insights be genuinely themed and ranked by prevalence.

**Suggested fix:** Extend the edge-case list (or step 4 itself) with: "If fewer than 5 emergent clusters exist, present all available clusters — do not pad to reach the 5–7 range," mirroring the existing "don't pad" language already used for the 1–2-entries case.

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
... — final preview of picks.** Parse the reply into the set of picked digest numbers. Run delegation step 1 (auth + repo check) once. Branch B never runs delegation step 4 (template discovery) and never adopts a repo template's body shape — every picked insight always uses the fixed Synthesis issue body shape below. Per pick, run: label discovery seeded directly at delegation step 5c's commit-tag mapping (the title always starts with `[feat]`, mapping to `enhancement`, then the verbatim/substring match chain against the repo's label list — no match found still omits `--label`, same as delegation step 5d), and delegation step 6 (duplicate scan). Capture one shared `date +%s` timestamp. For each pick `<n>` (its digest number), write its body via the **Write** tool to `/tmp/report-issue-lugassawan-swe-workbench-<ts>-<n>.md`. Write ONE shared `.cmd` sidecar at `/tmp/report-issue-lugassawan-swe-workbench-<ts>.cmd`, containing one `gh issue create --repo lugassawan/swe-workbench --title "..." --body-file <path-n> [--label "enhancement"]` line per pick — one enhancement issue per pick, never bundled. Print each selected body with its `Redacted:` / `Label:` / `Possibly related:` lines, then:
```
lugassawan: **Low**: Branch B step 7 says it runs "delegation step 1 (auth + repo check) once," which per its own spec requires surfacing `Filing into: lugassawan/swe-workbench` in the preview — but the Turn 2 preview line enumeration here only lists `Redacted:` / `Label:` / `Possibly related:`, omitting `Filing into:`.

**Why it matters:** Minor internal inconsistency: a reader following "runs delegation step 1" literally would expect the `Filing into:` line to appear in Branch B's Turn 2 preview, but the explicit line list doesn't include it. Low impact since the repo target is invariant and already stated elsewhere, but worth tightening for spec self-consistency.

**Suggested fix:** Add `Filing into:` to the enumerated preview lines in step 7, or explicitly note that Branch B's Turn 2 preview omits it as redundant (since the repo is already invariant/known).

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench))._

commands/report-issue.md
lugassawan: Addressed in 460b7d2: added a floor carve-out to step 4 — when fewer than 5 emergent clusters exist, present all available clusters rather than padding to the 5–7 range.

commands/report-issue.md
lugassawan: Addressed in 460b7d2: added the missing `Filing into:` line to Branch B's Turn 2 preview line enumeration in step 7, matching what delegation step 1 actually surfaces.

tests/test_report_issue_command.py
```diff
@@ -175,3 +175,156 @@ def test_report_issue_step9_cleanup_on_success_only():
     assert "failure" in text.lower() or "on failure" in text.lower() or "retry" in text.lower(), (
         "commands/report-issue.md must state that temp files are left intact on failure (for retry)"
     )
+
+
+# --- Synthesis mode assertions (issue #475) ---
+
+def _branch_b_slice(text):
```
lugassawan, line 182: **Medium**: `_branch_b_slice()` slices from `"### Branch B"` to end-of-file with no closing boundary, so it also swallows the entire downstream shared delegation block (steps 1–9). This means neither of the two previously-flagged-and-fixed defects (the 5–7 padding floor, and the missing `Filing into:` line) has real regression coverage: running the exact same assertions from `test_report_issue_synthesize_ranks_by_prevalence` (checks for `"5–7"`) and `test_report_issue_synthesize_pick_then_confirm_order` (checks for `"final preview"`/`"Filing into"`) against the pre-fix commit (7c30406) still returns `True` — because the matched substrings come from unrelated text later in the shared delegation section, not from the Branch-B-specific sentences that were actually fixed.

**Why it matters:** A future edit to Branch B step 4 or step 7 could silently reintroduce either bug and the suite would stay green.

**Suggested fix:** Bound `_branch_b_slice` to end where the shared delegation block starts (e.g. up to the `---\n\nDelegate to the` marker), and add assertions that pin the literal fixed sentences, e.g. `assert "do not pad to reach the 5–7 range" in branch_b` and a check that `"Filing into:"` appears specifically within step 7's own sentence rather than anywhere in the slice.

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)) — follow-up pass._

commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
...of picks.** Parse the reply into the set of picked digest numbers. Run delegation step 1 (auth + repo check) once. Branch B never runs delegation step 4 (template discovery) and never adopts a repo template's body shape — every picked insight always uses the fixed Synthesis issue body shape below. Per pick, run: label discovery seeded directly at delegation step 5c's commit-tag mapping (the title always starts with `[feat]`, mapping to `enhancement`, then the verbatim/substring match chain against the repo's label list — no match found still omits `--label`, same as delegation step 5d), and delegation step 6 (duplicate scan). Capture one shared `date +%s` timestamp. For each pick `<n>` (its digest number), write its body via the **Write** tool to `/tmp/report-issue-lugassawan-swe-workbench-<ts>-<n>.md`. Write ONE shared `.cmd` sidecar at `/tmp/report-issue-lugassawan-swe-workbench-<ts>.cmd`, containing one `gh issue create --repo lugassawan/swe-workbench --title "..." --body-file <path-n> [--label "enhancement"]` line per pick — one enhancement issue per pick, never bundled. Print each selected body with its `Filing into:` / `Redacted:` / `Label:` / `Possibly related:` lines, then:
```
lugassawan: **Medium**: Turn 2's preview enumeration lists `Filing into:` / `Redacted:` / `Label:` / `Possibly related:` but omits `Title:`. The "Synthesis issue body" shape keeps `Title` as a field distinct from the Sections/Footer body content — mirroring how the single-issue delegation flow (step 8's preview) keeps `Title:` separate from the fenced body. Since Title isn't part of the fenced Sections/Footer body either, the literal drafted title string is never shown to the user in Turn 1 or Turn 2.

**Why it matters:** The user is asked to reply `confirm` and file up to 7 GitHub issues without ever seeing the exact `--title` value that will be used — undermining the stated confirm-before-file design goal (an LLM-drafted title with a typo, wrong theme, or off-topic phrasing would go straight to `gh issue create` unreviewed).

**Suggested fix:** Add `Title:` to the enumerated preview-line list in step 7 (and to the Turn 1 digest entry format in step 6), e.g. "Print each selected body with its `Title:` / `Filing into:` / `Redacted:` / `Label:` / `Possibly related:` lines".

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)) — follow-up pass._

commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
+   ```
+   Reply 'confirm' to file all, or `drop N` / `edit N`.
+   ```
+   **Wait for `confirm`. Do NOT run `gh issue create` on this turn.** After `drop N` or `edit N`, re-print the updated preview and wait for `confirm` again — never file immediately after an edit.
```
lugassawan: **Low**: Step 7/8 don't cover the case where the user `drop`s every remaining pick before replying `confirm` — the sidecar would end up empty and `confirm` would run zero `gh` lines with no explicit "nothing to file" message.

**Why it matters:** Minor UX gap: a user who drops all picks gets silent no-op behavior instead of a clear exit message.

**Suggested fix:** Add "If all picks are dropped, exit cleanly with a 'No insights remain to file' message instead of waiting for `confirm`."

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)) — follow-up pass._

tests/test_report_issue_command.py
lugassawan: Addressed in 7c31e7e: bounded _branch_b_slice() to end at the delegation-block marker, and added two assertions that pin the literal fixed sentences (the 5-7 padding floor and the Title:/Filing into: preview-line enumeration), verified to fail against the pre-fix commit.

commands/report-issue.md
lugassawan: Addressed in 7c31e7e: added Title: to the enumerated preview lines in both the Turn 1 digest (step 6) and the Turn 2 final preview (step 7).

commands/report-issue.md
lugassawan: Addressed in 7c31e7e: added an explicit 'No insights remain to file' exit for both an initially-empty pick set and a drop N that empties it, instead of silently proceeding to a no-op confirm.

commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
...ith "No insights remain to file" instead of proceeding. Run delegation step 1 (auth + repo check) once. Branch B never runs delegation step 4 (template discovery) and never adopts a repo template's body shape — every picked insight always uses the fixed Synthesis issue body shape below. Per pick, run: label discovery seeded directly at delegation step 5c's commit-tag mapping (the title always starts with `[feat]`, mapping to `enhancement`, then the verbatim/substring match chain against the repo's label list — no match found still omits `--label`, same as delegation step 5d), and delegation step 6 (duplicate scan). Capture one shared `date +%s` timestamp. For each pick `<n>` (its digest number), write its body via the **Write** tool to `/tmp/report-issue-lugassawan-swe-workbench-<ts>-<n>.md`. Write ONE shared `.cmd` sidecar at `/tmp/report-issue-lugassawan-swe-workbench-<ts>.cmd`, containing one `gh issue create --repo lugassawan/swe-workbench --title "..." --body-file <path-n> [--label "enhancement"]` line per pick — one enhancement issue per pick, never bundled. Print each selected body with its `Title:` / `Filing into:` / `Redacted:` / `Label:` / `Possibly related:` lines, then:
```
lugassawan: **Medium**: The new opening sentence of step 7 ("If the reply drops every remaining pick (or the picked set is otherwise empty), exit cleanly with 'No insights remain to file' instead of proceeding") contradicts step 6's already-established rule for the exact same reply. Step 6 explicitly states: "An out-of-range number or `none` re-prompts once, then exits without filing" — and the edge-case list repeats this: "Out-of-range number or `none`: re-prompt once; file nothing." Step 7's new sentence fires on parsing that same Turn-1-reply-into-picked-set and instead exits immediately with a different message, skipping the promised one-time re-prompt. Additionally, "the reply drops every remaining pick" borrows `drop N` vocabulary that hasn't been introduced to the user yet at this point (the `drop N` command is only offered in the Turn 2 prompt printed later in the same step) — so the clause doesn't logically apply to the reply actually being parsed here, suggesting it was copy-pasted from the end-of-step-7 `drop N` handling without adapting to context.

**Why it matters:** This is fix-on-fix drift — the round-2 fix for the "no message on empty pick set" finding resolved that gap but left two contradictory instructions for what should happen on an empty/invalid Turn 1 reply.

**Suggested fix:** Scope the new sentence explicitly to *after* the re-prompt has already failed once, e.g.: "If, after the one re-prompt from step 6, the picked set is still empty, exit cleanly with 'No insights remain to file'." Drop the "drops every remaining pick" clause from this position entirely — that clause only makes sense in the `drop N` context later in the same step.

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)) — second follow-up pass._

commands/report-issue.md
```diff
@@ -36,6 +51,48 @@ If `$ARGUMENTS` is **empty**, do NOT error. Instead:
...
+   ```
+   Reply 'confirm' to file all, or `drop N` / `edit N`.
+   ```
+   **Wait for `confirm`. Do NOT run `gh issue create` on this turn.** After `drop N` or `edit N`, re-print the updated preview and wait for `confirm` again — never file immediately after an edit. If `drop N` empties the pick set, exit cleanly with "No insights remain to file" instead of re-prompting for `confirm`.
```
lugassawan, line 78: **Low**: Neither of the two new "No insights remain to file" sentences (round-2 fix) has a pinning test, unlike the round-1 fixes which now have dedicated assertions (tests/test_report_issue_command.py:252, 277-280) precisely because an earlier finding showed an untested claimed-fix can silently regress.

**Why it matters:** The same regression risk that motivated the earlier test-coverage fix now applies to this fix too.

**Suggested fix:** Add an assertion (e.g. `assert branch_b.count("No insights remain to file") == 2` or two positional checks) to `tests/test_report_issue_command.py` pinning both occurrences.

_Posted by \`reviewer\` ([swe-workbench](https://github.com/lugassawan/swe-workbench)) — second follow-up pass._

commands/report-issue.md
lugassawan: Addressed in 26a2f9f: scoped the empty-pick-set exit to fire only after step 6's own re-prompt has already failed, and dropped the premature drop-N clause from that position.

commands/report-issue.md
lugassawan: Addressed in 26a2f9f: added test_report_issue_synthesize_no_insights_remain_exit, pinning both occurrences plus the corrected re-prompt-scoping wording (verified it fails against the pre-fix commit).

PR #491, opened 2026-07-02
[feat] Add /sync command for guided branch-sync with conflict resolution

lugassawan:
## Summary

- Add `/swe-workbench:sync` command — brings the current branch up to date with the detected default branch, delegating the mechanical merge/rebase to rimba (MCP → binary → shell fallback), never hardcoding `main`.
- Add `workflow-branch-sync` orchestrator skill: preflight guard (refuses on default branch/detached HEAD, offers stash-or-abort on a dirty tree) → optional rimba overlap advisory → mechanical sync → per-file interactive conflict resolution → leaves the result local and prompts before pushing (`--force-with-lease` only under `--rebase`).
- Add `conflict-resolver` subagent — advisory-only (no Edit/Write), reasons per-hunk about a conflicting file, and emits a `KEEP-MINE`/`KEEP-MAIN`/`MANUAL` verdict with rationale.
- Add `apply-resolution.sh`, which is the single place that translates user intent (`mine`/`main`) into git's `--ours`/`--theirs` — correctly inverted under rebase vs merge — and handles delete/modify conflicts by falling back to `git rm` when the chosen side has no blob.
- Add `preflight-guard.sh` and `detect-conflicts.sh` companion scripts, plus registration in `docs/catalog.md`, `README.md`, and `agents/shared/workflows.md`.

## Test Plan

- [x] Changed skills/commands/agents load without errors — `python3 scripts/validate.py` passes clean.
- [x] Examples in the diff were exercised manually — the ours/theirs inversion and delete/modify fallback were verified against real induced git conflicts (merge and rebase) in a scratch repo before landing the fix; the eval-injection fix was verified RED (fails without `%q`) then GREEN (passes with it).
- [x] Full test suite green: `pytest tests/ -q` → 1522 passed, including 6 new test files covering each script/doc plus injection and delete/modify regression tests.

Closes #485


skills/workflow-branch-sync/SKILL.md
```diff
@@ -0,0 +1,152 @@
...
+- **`OPERATION=merge`** → `git commit --no-edit`.
+- **`OPERATION=rebase`** → `git rebase --continue`. **A rebase can pause again on the next replayed commit** — re-run Step 4 (`detect-conflicts.sh`) and loop back into Step 5 if `OPERATION=rebase` with a fresh `UNMERGED` list. Only proceed to Step 6 once `git rebase --continue` completes the rebase entirely.
+
+### Step 6 — Leave Local & Prompt Before Push
+
+Report the resolution summary: one line per file — which side was kept (or "manual") and the one-line rationale from Step 5. **Never auto-push.**
+
+Prompt: "Sync complete locally on `$CURRENT_BRANCH`. Push now?"
+
+- **Yes, and `OPERATION` was `merge`** → `git push`.
```
lugassawan: **High** | Step 6's push branching only handles `OPERATION=merge`/`OPERATION=rebase`, but `detect-conflicts.sh` reports `OPERATION=none` for a **clean** sync (the common no-conflict case per Step 4). Neither documented branch applies here — if the resolved strategy was `--rebase` and it completed cleanly, this leaves no path to the required `--force-with-lease` push, undermining the force-push-only-under-rebase contract for the majority case.

**Fix:** capture the resolved strategy into its own variable in Step 3 (e.g. `SYNC_STRATEGY=merge|rebase`), independent of `OPERATION` (which correctly means "conflict resolution in progress" in Steps 4/5), and branch Step 6 on `SYNC_STRATEGY` instead of `OPERATION`.

skills/workflow-branch-sync/SKILL.md
```diff
@@ -0,0 +1,152 @@
...
+_RT="${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}"
+_SCRIPTS="$_RT/skills/workflow-branch-sync/scripts"
+eval "$("$_SCRIPTS/preflight-guard.sh")"
+```
+
+Emits `CURRENT_BRANCH`, `DEFAULT_BRANCH` (detected — never hardcode `main`), `IS_DEFAULT`, `DETACHED`, `DIRTY`.
+
+- **`IS_DEFAULT=1`**: refuse. Report "already on `$DEFAULT_BRANCH` — nothing to sync onto itself" and stop.
+- **`DETACHED=1`**: refuse. Report "detached HEAD — checkout a branch first" and stop.
+- **`DIRTY>0`**: offer **stash-or-abort** before touching history — `git stash push -u -m "branch-sync: pre-sync stash"` if the user opts to stash, otherwise stop and let the user commit or discard first. Never proceed with a dirty tree.
```
lugassawan: **High** | Step 1's dirty-tree guard stashes with `git stash push -u -m "branch-sync: pre-sync stash"` when the user opts to stash, but no later step (2 through 6), the Failure Mode Table, or Common Mistakes ever pops or references restoring that stash.

A user who opts to stash gets their uncommitted work silently parked indefinitely — Step 6's resolution summary gives no indication a stash exists or needs restoring. Repeated syncs would pile up unlabeled/forgotten stashes.

**Fix:** after the mechanical sync succeeds (Step 3) or the conflict loop completes (Step 5), pop the pre-sync stash if one was created, surfacing any pop-conflict the same way file conflicts are surfaced — or, if leaving it stashed is intentional, say so explicitly in Step 6's report.

skills/workflow-branch-sync/SKILL.md
lugassawan: Addressed in 3591d18c54266dfc147086d07f7c72376f6487df: decoupled Step 6's push branching into a dedicated SYNC_STRATEGY variable (captured in Step 3), independent of OPERATION — which is `none` on a clean sync and had no push path for that case under --rebase.

skills/workflow-branch-sync/SKILL.md
lugassawan: Addressed in 3591d18c54266dfc147086d07f7c72376f6487df: Step 6 now restores the pre-sync stash (git stash pop) before reporting when Step 1 set STASHED=1, surfacing a conflicting pop the same way a file conflict from Step 5 is surfaced (show both sides, resolve, then git stash drop).

skills/workflow-branch-sync/scripts/apply-resolution.sh
```diff
@@ -0,0 +1,56 @@
...
+    ;;
+esac
+
+if [ "$OP" = "merge" ]; then
+  [ "$SIDE" = "mine" ] && GIT_SIDE="ours" || GIT_SIDE="theirs"
+else
+  [ "$SIDE" = "mine" ] && GIT_SIDE="theirs" || GIT_SIDE="ours"
+fi
+
+CHECKOUT_ERR=$(git checkout "--$GIT_SIDE" -- "$FILE" 2>&1) && git add -- "$FILE" || {
```
lugassawan: **Low** | `CHECKOUT_ERR=$(git checkout "--$GIT_SIDE" -- "$FILE" 2>&1) && git add -- "$FILE" || { ... }` is an `A && B || C` chain (shellcheck SC2015). If `git checkout` succeeds but the subsequent `git add` fails (e.g. permission error, disk full), the `||` block still runs but `$CHECKOUT_ERR` holds the (successful) checkout output rather than the add failure — the script still exits 1 correctly, but the emitted error message is misleading and does not describe the actual failure.

**Fix:** split into two statements — `git checkout "--$GIT_SIDE" -- "$FILE"` with its own error handling, then a separate `git add` with its own.

lugassawan: Addressed in d9d9ca140f97512ff07410c691526ade97c72f26: restructured the checkout/add chain from A && B || C into if/else, so a failing git add now surfaces its own error instead of the checkout's (successful) output being misattributed.

PR #515, opened 2026-07-11
[fix] Close newline/tab rm bypass in bash_guard.sh

lugassawan:
## Summary

- `hooks/bash_guard.sh`'s separator normalization (`tr`) only folded `;|&` to spaces before the fast-exit `case` gate checked for `rm`/`git`. A newline or tab before `rm` slipped past the case gate entirely and never reached the capable grep check below it — e.g. `echo hi\nrm -rf ~` was silently allowed.
- Extended the `tr` set from `';|&'` (3 chars) to `';|&\n\t'` (5 chars), keeping SET1/SET2 equal length, so newline/tab now fold to spaces before the gate runs.
- Added 4 regression tests to `tests/test_hooks.py::TestRmRfBlocker`: two BLOCKED cases (newline- and tab-separated `rm -rf`) and two ALLOWED cases (newline- and tab-separated safe commands) proving the fix does not over-block, for either separator.
- Also resolves finding #1 of #501 (the `tr` multiline bypass); findings #4/#6/#8 of #501 remain open and are intentionally out of scope for this PR.
- Note: as a side effect of folding newlines to spaces, a *literal* multi-line mention of a blocked pattern (e.g. `rm -rf /` appearing as text inside a heredoc or a multi-line commit message) is now also blocked, even when `rm` isn't actually invoked. This is an accepted fail-closed trade-off for a security guard, consistent with the hook's existing "parse error blocks by default" posture.

## Test Plan

- [x] `pytest tests/test_hooks.py -v` — RED confirmed (2 new BLOCKED cases failed pre-fix), GREEN confirmed (all pass post-fix)
- [x] `pytest tests/ -v` — 1560/1560 pass
- [x] `shellcheck hooks/bash_guard.sh` — 0 issues
- [x] Manual guard invocation via stdin JSON for both new BLOCKED cases and the ALLOWED case — behavior matches expectations

- [x] Bug reproduced pre-fix (RED), fix verified to close it (GREEN)

Closes #502

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -71,6 +71,8 @@ class TestRmRfBlocker:
         "ls;rm -rf /",
         "ls;rm -rf /Users/foo",
         "ls&&rm -rf /home/foo",
+        "echo hi\nrm -rf ~",
+        "true\trm -rf /",
     ])
     def test_blocked(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
@@ -91,6 +93,7 @@ class TestRmRfBlocker:
         "rm -r ./dist",
         "rm -rf /UsersHome",
         "rm -rf /homestead",
+        "echo hi\nrm -rf ./build",
     ])
     def test_allowed(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -94,6 +94,7 @@ class TestRmRfBlocker:
         "rm -rf /UsersHome",
         "rm -rf /homestead",
         "echo hi\nrm -rf ./build",
+        "true\trm -rf ./build",
     ])
     def test_allowed(self, guard_script, cmd):
         result = run_guard(guard_script, cmd)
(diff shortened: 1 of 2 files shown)


tests/test_hooks.py
```diff
@@ -91,6 +93,7 @@ def test_blocked(self, guard_script, cmd):
         "rm -r ./dist",
         "rm -rf /UsersHome",
         "rm -rf /homestead",
+        "echo hi\nrm -rf ./build",
```
lugassawan, line 96: **Low** | tests/test_hooks.py:96

The ALLOWED corpus only adds a newline-separated non-destructive case (`"echo hi\nrm -rf ./build"`); no tab-separated ALLOWED counterpart, even though a tab-separated BLOCKED case was added just above (`"true\trm -rf /"`). Manually verified `"true\trm -rf ./build"` correctly exits 0 with the fix — not a bug, but the test matrix is asymmetric (newline gets both BLOCKED+ALLOWED coverage, tab only gets BLOCKED), leaving the tab non-overblock path unverified by CI.

Suggested fix: add `"true\trm -rf ./build"` to the `test_allowed` parametrize list for symmetry with the newline case.

lugassawan: Addressed in 5785457: added `"true\trm -rf ./build"` to the `test_allowed` parametrize list for symmetry with the newline case.

PR #516, opened 2026-07-11
[fix] Add macOS pytest CI + warn on transitive lockfile adds

lugassawan:
## Summary

- `pr.yml`'s `pytest` job now runs a `[ubuntu-latest, macos-latest]` matrix (`fail-fast: false`) — every other job stays `ubuntu-latest`-only. Three hook scripts (`hooks/bash_guard.sh`, `hooks/worktree_permission_grant.sh`, `hooks/skill_usage_record.sh`) have deliberate BSD-only code paths that only the pytest job's `subprocess.run`-driven tests actually exercise; `shellcheck` output is platform-identical and `lockfile-check` risks pip-compile platform-resolution false-fails, so both stay ubuntu-only.
- `scripts/check-lockfile-additions.sh` now also flags new **transitive** pip packages (previously silent) via a non-blocking `::warning::` plus an optional `$GITHUB_STEP_SUMMARY` line. New **top-level** packages still hard-fail (exit 1) exactly as before, so Dependabot's lockfile auto-commit step in `dependabot-lockfile.yml` keeps running unaffected.
- Fixed two bugs surfaced during review: a `paste -d ', '` multi-char delimiter that garbled 2+ package names in the warning, and a `GITHUB_STEP_SUMMARY` env leak in `tests/conftest.py`'s `_CLEAN_ENV` that would have written fabricated warnings into the real `pytest` job's CI summary on every run.
- Added a regression test covering the case where a package is transitive-only in the base lock and gets promoted to top-level in the new lock — locks in that it still hard-fails (no duplicate `::warning::`), per follow-up review feedback.

⚠️ **Deploy-time note:** the matrix renames the `pytest` status check to `pytest (ubuntu-latest)` / `pytest (macos-latest)`. If branch protection requires a check literally named `pytest`, update the required-checks list or merges will hang.

## Test Plan

- [x] `pytest tests/ -v` — 1561/1561 pass (includes new/extended lockfile-guard tests)
- [x] `shellcheck scripts/check-lockfile-additions.sh` — 0 issues
- [x] `actionlint .github/workflows/pr.yml` — clean
- [x] Manual smoke test: transitive-only addition → exit 0 + `::warning::`; `GITHUB_STEP_SUMMARY` set → warning line lands in the file; unset → no leak

- [x] Reproduced both pre-fix bugs locally (`paste -sd ', '` garbling multi-package names; `GITHUB_STEP_SUMMARY` leaking into `pytest` job output) before fixing
- [x] Verified the guard script's hard-fail path (new top-level package) is unchanged and cannot be miscategorized as transitive-only

Closes #503

### DIFF
--- a/scripts/check-lockfile-additions.sh
+++ b/scripts/check-lockfile-additions.sh
@@ -36,18 +36,38 @@ extract_top_level() {
   ' | sort -u
 }
 
+# ── All-packages extractor ─────────────────────────────────────────────────
+# Same package set, but ignores the -r *.txt predicate — every pinned
+# package (top-level or transitive) is included.
+
+extract_all_packages() {
+  awk '
+    /^[A-Za-z0-9._-]+==/ {
+      name = $1
+      sub(/==.*/, "", name)
+      if (name != "") print name
+    }
+  ' | sort -u
+}
+
 # ── Per-lockfile check ─────────────────────────────────────────────────────
 
 failed=0
 
 for lock in "$@"; do
   # Base set: HEAD content (pre-regen, pre-commit).
+  # NOTE: baseline is HEAD, not main — on a `synchronize` after the bot has
+  # already committed a regen to this branch, a transitive package warned on
+  # in an earlier push won't re-warn. Acceptable: this signal is non-blocking.
   # When the lockfile is new (absent from HEAD), every package is treated as an
   # addition — this is intentional; a new lockfile always requires human review.
   if git cat-file -e "HEAD:${lock}" 2>/dev/null; then
-    base=$(git show "HEAD:${lock}" | extract_top_level)
+    base_content=$(git show "HEAD:${lock}")
+    base=$(printf '%s\n' "${base_content}" | extract_top_level)
+    base_all=$(printf '%s\n' "${base_content}" | extract_all_packages)
   else
     base=""
+    base_all=""
   fi
 
   # New set: working-tree content (post-regen).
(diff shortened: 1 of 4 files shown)


scripts/check-lockfile-additions.sh
```diff
@@ -71,6 +92,29 @@ for lock in "$@"; do
...
   fi
+
+  # Transitive additions: new packages (any provenance) minus new top-level
+  # additions already reported above. Non-blocking — surfaced for visibility
+  # only, so the Dependabot auto-commit step keeps running.
+  added_all=$(comm -13 \
+    <(printf '%s\n' "${base_all}" | grep -v '^$') \
+    <(printf '%s\n' "${new_all}"  | grep -v '^$'))
+  added_transitive=$(comm -23 \
+    <(printf '%s\n' "${added_all}" | grep -v '^$') \
```
lugassawan, line 103: **Low:** No automated regression test exercises the case where a package already present as transitive-only in the base lockfile gets *promoted* to top-level in the new lockfile (i.e. the `added_all`/`added` overlap-exclusion path with a shared package name, distinct from the two-distinct-new-packages case `test_transitive_and_top_level_addition_exits_one_and_warns` already covers).

Manually verified this is handled correctly today (exit 1, single `::error::`, no duplicate `::warning::`), but it's new interaction surface introduced by the `added_all` construction in this PR — a codified test would guard against a future refactor silently reintroducing double-counting.

Optional: add a case where a package is transitive-only in the base lock and top-level in the new lock, asserting exit 1 with no `::warning::` for that package name.

lugassawan: Addressed in e88088b1dab0bf8f0d838780069aa56de6049a3c: added test_transitive_promoted_to_top_level_exits_one_no_warning covering the case where a package is transitive-only in the base lock and top-level in the new lock — asserts exit 1 (top-level hard-fail) with no duplicate ::warning::.

PR #517, opened 2026-07-11
[fix] Bound runtime gh calls with a per-call timeout wrapper

lugassawan:
## Summary

- Every `gh` CLI call in the 4 `runtime/` PR-helper scripts (`fetch-pr.sh`, `reply-and-resolve.sh`, `sync-pr-metadata.sh`, `preflight-pr.sh`) ran unbounded — a stalled GitHub API response never returns and never errors, which can leave `reply-and-resolve.sh` mid-loop with a review thread partially replied/resolved.
- Adds `runtime/gh-timeout.sh`, a subprocess wrapper (matches the repo's existing sibling-subprocess resolution convention, not a sourced helper) that runs `gh` under `timeout -k 5 <secs>` (default 60s, overridable via `GH_TIMEOUT_SECS`), falls back to `gtimeout` (macOS/coreutils) if `timeout` is absent, and degrades to unbounded `gh` if neither exists.
- Rejects `GH_TIMEOUT_SECS=0` (falls back to 60s) — GNU `timeout`/`gtimeout` treat a `0` duration as "disable the timer", so a `0` override would otherwise silently reintroduce an unbounded call.
- On a 124 (timeout) exit, prints a stderr diagnostic naming the timed-out `gh` call so a stall isn't misreported as e.g. "PR not found".
- Rewrites all `gh` call sites in the 4 scripts to go through the wrapper; no behavior change for callers.

## Test Plan

- [x] `shellcheck --severity=warning runtime/gh-timeout.sh runtime/*.sh` — 0 issues
- [x] `pytest tests/ -v` — 1567/1567 pass, including 10 new `test_gh_timeout_script.py` cases
- [x] Manual smoke: `GH_TIMEOUT_SECS=2 bash runtime/gh-timeout.sh api rate_limit` succeeds against the real GitHub API; on a host with neither `timeout` nor `gtimeout` installed, a slow stubbed `gh` ran unbounded to completion, confirming the degrade path

- [x] Reproduced the root cause: unbounded `gh` calls with no application-level deadline (Go `net/http` sets none on the response-body read)
- [x] Verified the fix bounds every call site in scope and a regression test (`GH_TIMEOUT_SECS=0`) guards the exact foot-gun both reviewers independently flagged

Closes #504

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -19,6 +19,7 @@ the swe-workbench plugin and are invoked at runtime by skills, commands, and age
 | `clean-state-files.sh` | Safe `rm -f` for per-invocation `/tmp` state files |
 | `doctor.sh` | Read-only preflight check of runtime dependencies (gh, git, jq, rimba, claude) |
 | `fetch-pr.sh` | Fetch a PR's metadata JSON via `gh pr view`; exits 1 if the PR is inaccessible |
+| `gh-timeout.sh` | Run a `gh` call under a per-call deadline (default 60s, override via `GH_TIMEOUT_SECS`); degrades to unbounded `gh` when neither `timeout` nor `gtimeout` is on PATH |
 | `preflight-pr.sh` | Consolidated pre-flight for PR-review skills: `gh auth` gate → `fetch-pr.sh` → emits `BASE`, `HEAD_SHA`, `AUTHOR_LOGIN`, `OWNER`, `REPO`, `STATE` as `printf %q`-quoted eval-able `KEY=VALUE` lines |
 | `reply-and-resolve.sh` | Post a PR review thread reply (REST) and optionally resolve it (GraphQL) |
 
--- a/runtime/fetch-pr.sh
+++ b/runtime/fetch-pr.sh
@@ -5,6 +5,7 @@ set -euo pipefail
 PR="${1:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
 OUT="${2:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
 FIELDS="${3:?Usage: fetch-pr.sh <PR> <out_path> <json_fields>}"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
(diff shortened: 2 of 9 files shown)


runtime/README.md
```diff
@@ -19,6 +19,7 @@ the swe-workbench plugin and are invoked at runtime by skills, commands, and age
 | `clean-state-files.sh` | Safe `rm -f` for per-invocation `/tmp` state files |
 | `doctor.sh` | Read-only preflight check of runtime dependencies (gh, git, jq, rimba, claude) |
 | `fetch-pr.sh` | Fetch a PR's metadata JSON via `gh pr view`; exits 1 if the PR is inaccessible |
+| `gh-timeout.sh` | Run a `gh` call under a per-call deadline (default 60s, override via `GH_TIMEOUT_SECS`); degrades to unbounded `gh` when neither `timeout` nor `gtimeout` is on PATH |
```
lugassawan, line 22: Low | Design/informational: the linked issue (#504) suggested "a shared `_gh()`/wrapper" but this PR ships a separate script invoked via `bash "$SCRIPT_DIR/gh-timeout.sh" ...` at ~8 call sites, each re-deriving its own `SCRIPT_DIR`. Not a correctness issue — tests confirm it works, and arguably a standalone script is more testable than a sourced function — but worth a note that this was a deliberate interpretation of the suggested fix, not an oversight.

lugassawan: Deliberate choice, not an oversight: the plan for #504 explicitly resolved on a subprocess wrapper over a sourced `_gh()` function — it reuses the repo's existing sibling-subprocess resolution pattern (documented at preflight-pr.sh, e.g. fetch-pr.sh calling it), avoids introducing a second inter-script mechanism (sourcing) alongside the existing one, keeps shellcheck clean with zero directives, and satisfies the RUNTIME_SCRIPTS existence/chmod/syntax guard as a real executable. The per-call-site `SCRIPT_DIR` re-derivation is the cost of that tradeoff, accepted for a Low-severity reliability fix. Leaving unresolved is fine — no action needed, just documenting the rationale for future readers.

PR #518, opened 2026-07-11
[fix] cleanup-merged: internal timeout guard + partial-deletion detection in sync-and-verify.sh

lugassawan:
## Summary

- `sync-and-verify.sh` ran a blocking `git pull --ff-only` that fires the rimba `post-merge` hook (`rimba clean --merged --force`) as a side effect; an external tool-call timeout (e.g. a 2-minute limit) could SIGKILL the whole process tree mid-hook, leaving a worktree registered in `.git` with its directory already deleted — a silent, hard-to-diagnose partial deletion.
- Adds a pure-bash `SYNC_TIMEOUT` watchdog (90s default, user-configurable) around the checkout-then-pull sync step: backgrounds it under `set -m` in its own process group, polls with `kill -0`, and on timeout sends `kill -TERM`/`kill -KILL` to the negative PID (whole group) to reap both git and any orphaned hook child.
- Adds a stateless partial-deletion probe: scans `git worktree list --porcelain` for any registered worktree whose directory is missing on disk and reports it via a new additive `HOOK_INTERRUPTED=0|1` stdout line (extends the existing `WORKTREE_GONE=0|1` one-line contract to two lines).
- The script stays **verify-only** — it signals via stdout/stderr and documents recovery, it never auto-remediates (no `git branch -D`, no `git worktree prune` inside the script).
- `SKILL.md` documents the new field, the `SYNC_TIMEOUT < external-timeout` invariant, a new Failure Mode Table row, and a manual recovery example (`git worktree prune` then delete the stale branch) — with explicit honest framing that the guard makes corruption *detectable and recoverable*, not *prevented*.

## Test Plan

- [x] `pytest tests/ -v` — 1561/1561 pass, including two new tests: a state-fixture test (`git worktree add` + `rm -rf` its dir → asserts `HOOK_INTERRUPTED=1`) and a watchdog test (`PATH`-shimmed slow `git` stub + `SYNC_TIMEOUT=1` → asserts exit 0, clean two-line stdout, and the stub subprocess actually reaped via `pgrep`)
- [x] `shellcheck -S warning` on all `*.sh`/`*.bash` — 0 issues
- [x] `bash scripts/validate.sh` — passes

- [x] Reproduced the original bug's failure mode locally (standalone `set -m` group-kill script against a `sleep 30 & sleep 30; wait` nested-child stand-in for git + orphaned hook child — confirmed the group kill reaps both, no orphan left via `pgrep`)
- [x] Verified the fix resolves it without regressing the existing one-line `WORKTREE_GONE` contract (all 8 pre-existing `test_sync_and_verify.py` tests still pass unchanged)

Closes #496

### DIFF
--- a/skills/workflow-cleanup-merged/SKILL.md
+++ b/skills/workflow-cleanup-merged/SKILL.md
@@ -248,10 +248,12 @@ A `SYNC_TIMEOUT` firing (or an external tool-call kill) can land mid-`rm` inside
 ```bash
 cd "$MAIN_REPO"
 # Identify which registered worktree(s) have no directory on disk (same
-# check Block D performs internally — a plain grep on porcelain output
-# cannot answer this, since "worktree <path>" lines don't self-report state):
-git worktree list --porcelain | awk '/^worktree /{print $2}' | while read -r w; do
-  [ -d "$w" ] || echo "MISSING: $w"
+# check Block D performs internally). Line-based, not awk field-splitting —
+# a path containing a space would silently vanish under `awk '{print $2}'`:
+git worktree list --porcelain | while IFS= read -r line; do
+  case "$line" in
+    "worktree "*) w=${line#worktree }; [ -d "$w" ] || echo "MISSING: $w" ;;
+  esac
 done
 git worktree prune                              # drops the stale registration(s) for the missing dir(s)
 git branch -D <stale-branch-name>                # the local ref that survived the interrupted rm
--- a/skills/workflow-cleanup-merged/SKILL.md
+++ b/skills/workflow-cleanup-merged/SKILL.md
@@ -90,12 +90,17 @@ The script: derives `MAIN_REPO=` (main worktree root via `git worktree list --po
 
 When the rimba post-merge hook is active (see `### rimba + post-merge hook (fast path)`), `git pull` fires the hook as a side-effect, which removes the merged worktree and local branch automatically. A sync failure on the fast path forces fall-through to the rimba-binary or shell strategy — it does NOT abort cleanup.
 
+**Internal timeout guard.** The checkout-then-pull sync step runs under a `SYNC_TIMEOUT` watchdog (`90` seconds by default, override via the `SYNC_TIMEOUT` env var) — a pure-bash `set -m` job-group backgrounding, not the `timeout` binary (absent on stock macOS). If the pull hangs, the watchdog `kill`s the whole process group at `SYNC_TIMEOUT` seconds. **Invariant the caller must hold:** `SYNC_TIMEOUT` must stay below the harness's external tool-call timeout (headroom for Block C/D plus a 2s kill grace) — the script cannot self-enforce this. **What the guard does and does not do:** it does NOT prevent corruption — an external kill can still land mid-`rm` inside the rimba post-merge hook. It converts an otherwise-uncontrolled external kill (which would also SIGKILL this script mid-hook, leaving the failure silent) into a controlled in-script timeout that still runs Block D's detection below. Firing at 90s, below the common 120s external tool-call limit, usually lands during the network pull itself (a clean kill, no hook involved) rather than mid-hook.
+
+**Hook-interruption detection (stateless, not event-based).** After the sync, the script probes `git worktree list --porcelain` for any registered worktree whose directory is missing on disk — the canonical signal that a `post-merge` hook `rm` (this run's or a prior one's) was interrupted mid-deletion, leaving the worktree registration and branch ref alive in `.git` while the files are gone. This is state-based rather than event-based on purpose: an external kill takes down this script too, so it cannot reliably observe its own interruption — but a stale registration on the next run always tells the truth.
+
 ### Step 4 — Remove Worktree
 
-`sync-and-verify.sh` (Step 3) emits `WORKTREE_GONE=0|1` into the shell environment via `eval`.
+`sync-and-verify.sh` (Step 3) emits two `KEY=VALUE` lines into the shell environment via `eval`: `WORKTREE_GONE=0|1` and `HOOK_INTERRUPTED=0|1`.
 
 - **`WORKTREE_GONE=1`**: both the worktree and local branch are already gone — the hook did its job. No further action is needed in Step 4; proceed to Step 5. The script reports `LOCAL_DELETED=0` (local already gone) and still attempts the remote delete.
 - **`WORKTREE_GONE=0`**: hook did not fire (or rimba refused due to dirty/unpushed state). Select a removal strategy from `## Worktree Removal Strategies` below. Execute only the first strategy whose preconditions hold.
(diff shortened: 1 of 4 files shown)


skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -234,6 +239,25 @@ git worktree remove "$WORKTREE"
...
+### Recovery Example — Interrupted Hook
+
+A `SYNC_TIMEOUT` firing (or an external tool-call kill) can land mid-`rm` inside the rimba post-merge hook, leaving a worktree registered in `.git` with its directory already gone from disk. `sync-and-verify.sh` reports this as `HOOK_INTERRUPTED=1` and prints a recovery hint to stderr. The probe scans **all** registered worktrees, not just `$HEAD_REF`'s — `HOOK_INTERRUPTED=1` can point at an unrelated stray left over from a different branch's interrupted cleanup, so identify the specific entry before assuming it's the one you're cleaning up:
+
+```bash
+cd "$MAIN_REPO"
+# Identify which registered worktree(s) have no directory on disk (same
+# check Block D performs internally — a plain grep on porcelain output
+# cannot answer this, since "worktree <path>" lines don't self-report state):
+git worktree list --porcelain | awk '/^worktree /{print $2}' | while read -r w; do
```
lugassawan: **Medium** | Recovery Example's `awk` command silently drops worktree paths containing spaces

The diagnostic loop `git worktree list --porcelain | awk '/^worktree /{print $2}'` field-splits on whitespace, so a worktree registered at a path containing a space (e.g. macOS `~/Library/CloudStorage/...` paths) prints nothing — a silent false-negative in exactly the "reported clearly, not silent failure" scenario issue #496 is meant to guarantee. Block D itself avoids this by stripping the `"worktree "` prefix (`${_line#worktree }`) rather than field-splitting.

**Suggested fix:** mirror Block D's line-based approach instead of `awk` field-splitting:
```bash
git worktree list --porcelain | while IFS= read -r line; do
  case "$line" in
    "worktree "*) w=${line#worktree }; [ -d "$w" ] || echo "MISSING: $w" ;;
  esac
done
```

_Posted by the `reviewer` subagent via [swe-workbench](https://github.com/lugassawan/swe-workbench) `/swe-workbench:review`._

tests/test_sync_and_verify.py
```diff
@@ -168,6 +175,117 @@ def test_branch_absent_trunk_default(self, git_repo_trunk):
...
+        result = _run_script(git_repo, BRANCH, default_branch="main")
+        _assert_contract(result, "0", hook_interrupted="1")
+
+
+class TestSyncAndVerifyWatchdog:
+    """The internal SYNC_TIMEOUT guard (issue #496) must reap a hung `git pull`
+    within the configured window — headless (no tty, no job control) — without
+    polluting stdout and without leaking the killed subprocess."""
+
+    def test_watchdog_kills_hung_pull_and_stays_clean(self, git_repo, tmp_path):
```
lugassawan, line 205: **Medium** | No test exercises the combined `TIMED_OUT=1 && HOOK_INTERRUPTED=1` path (sync-and-verify.sh's `if [ "$TIMED_OUT" -eq 1 ]` branch), which is the exact scenario issue #496 was filed for

`test_watchdog_kills_hung_pull_and_stays_clean` only hangs `git pull` (no `rm`), so it reaps a benign timeout with `HOOK_INTERRUPTED=0`. `test_partial_deletion_detected` manually deletes a stray worktree directory without ever triggering the watchdog, so `TIMED_OUT=0` there. The branch that emits the distinct "internal timeout (${SYNC_TIMEOUT}s) interrupted the post-merge hook" stderr message — the specific message the acceptance criteria call out for the watchdog-caused case — is therefore unverified by any test, even though it passes shellcheck and looks correct by inspection.

**Suggested fix:** add a test that stubs a hung `git pull` (like the existing watchdog test) but also has the stub simulate the hook's `rm` by deleting a sibling registered worktree directory before/while sleeping, so `SYNC_TIMEOUT` fires with a genuinely missing worktree present — then assert `HOOK_INTERRUPTED=1` and that stderr contains "internal timeout" (not the generic partial-deletion message).

_Posted by the `reviewer` subagent via [swe-workbench](https://github.com/lugassawan/swe-workbench) `/swe-workbench:review`._

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 967cb1795218bc9b229b5f49002d3a4cbee92c31: replaced awk field-splitting with a line-based read/case loop (mirrors Block D's own probe) so worktree paths containing spaces are no longer silently dropped.

tests/test_sync_and_verify.py
lugassawan: Addressed in 967cb1795218bc9b229b5f49002d3a4cbee92c31: added test_watchdog_timeout_concurrent_with_hook_interruption, which stubs a hung git pull AND deletes a sibling registered worktree's directory, so SYNC_TIMEOUT fires with a genuine partial deletion present. Asserts HOOK_INTERRUPTED=1 and the timeout-corroborated stderr wording specifically (not the generic partial-deletion message).

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -15,8 +15,44 @@ MAIN_REPO=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
...
+    break
+  fi
+  sleep 1
+  elapsed=$((elapsed + 1)) # NOT ((elapsed++)) — returns 1 (falsy) at 0, trips set -e
+done
+pull_rc=0
+wait "$job" 2>/dev/null || pull_rc=$? # `wait; pull_rc=$?` on separate lines is unreachable under set -e
+set +m
+
+if [ "$TIMED_OUT" -eq 0 ] && [ "$pull_rc" -ne 0 ]; then
```
lugassawan: **High** | When `SYNC_TIMEOUT` fires but no worktree is mid-deletion (`TIMED_OUT=1`, `HOOK_INTERRUPTED=0`), the script emits **no stderr message at all**

The `TIMED_OUT -eq 0` guard on this line suppresses the old "best-effort failed" warning whenever the watchdog fired, and the Block D message block (lines 82-88) only fires when `HOOK_INTERRUPTED=1`. Per this PR's own SKILL.md documentation this is the *common* case ("Firing at 90s... usually lands during the network pull itself (a clean kill, no hook involved) rather than mid-hook"), not an edge case.

The watchdog forcibly `kill -KILL`s `git checkout`/`git pull` mid-operation and the caller receives a clean two-line `WORKTREE_GONE=0 / HOOK_INTERRUPTED=0` stdout with zero diagnostic — silently indistinguishable from a fully successful sync. `$MAIN_REPO` could be left checked out to the wrong ref or mid-fetch with a stale `main`, and nothing surfaces it. This partially defeats issue #496's own acceptance criterion that timeout activity be "reported clearly" — it's currently reported only when coincident with a hook interruption.

**Suggested fix:** emit a stderr message whenever `TIMED_OUT -eq 1`, independent of `HOOK_INTERRUPTED`, e.g. an `elif [ "$TIMED_OUT" -eq 1 ]` branch: `echo "sync-and-verify: internal timeout (${SYNC_TIMEOUT}s) killed the checkout/pull — $DEFAULT_BRANCH sync may be incomplete, verify manually" >&2`. Add a test asserting non-empty, timeout-naming stderr for the `TIMED_OUT=1, HOOK_INTERRUPTED=0` case specifically.

_Posted by the `reviewer` subagent via [swe-workbench](https://github.com/lugassawan/swe-workbench) `/swe-workbench:review --check-followup`._

lugassawan: Addressed in a44d9a291d61979787198996ef640786c8c9eaea: TIMED_OUT=1 now always emits a stderr diagnostic (independent of HOOK_INTERRUPTED), and test_watchdog_kills_hung_pull_and_stays_clean now asserts stderr names the timeout for the plain (no-hook-interruption) case.

PR #521, opened 2026-07-12
[feat] Add --check-redundancy to sync: surface functional duplication main already provides

lugassawan:
## Summary
- Add an opt-in `--check-redundancy` flag to `/swe-workbench:sync` (issue #510) that surfaces *functional* duplication a textual merge/rebase diff structurally cannot see — a branch adding a file that the default branch independently grew elsewhere (renamed/relocated), which git reports as no conflict at all.
- New `skills/workflow-branch-sync/scripts/redundancy-scope.sh` — deterministic gather script, pure function of `(MERGE_BASE, PRE_SYNC_HEAD, DEFAULT_REF)`. Emits whole-file candidates the branch added, an inbound-reference guard (`refs=<count>`), and what the default branch added/changed in the same window.
- New `agents/redundancy-assessor.md` — advisory-only subagent that reasons about functional equivalence and emits a per-candidate `AUTO-APPLY|ESCALATE|NONE` sentinel; tier rule restricts `AUTO-APPLY` to whole-file, zero-reference candidates.
- `skills/workflow-branch-sync/SKILL.md` — Step 3 now captures a merge-base baseline *before* the mechanical sync (flag-gated); new Step 6 (Redundancy Assessment) runs the gather script + assessor and validates every subagent-reported id against the script's own enumerated candidates before acting — never trusting agent free text for an actionable path. Former Step 6 (push) renumbered to Step 7.
- `commands/sync.md` — parses/strips `--check-redundancy`, threads `CHECK_REDUNDANCY=on|off` to the skill.
- Plain `sync` (no flag) is byte-for-byte unchanged — `CHECK_REDUNDANCY` defaults to `off`.

## Test Plan
- [x] Changed skills/commands/agents load without errors (`python3 scripts/validate.py` passes)
- [x] Examples in the diff were exercised manually (hand-built diverged repos, real `git merge`, direct script invocation — see below)
- [x] Full pytest suite passes (1620/1620), `shellcheck` clean on all `*.sh`

- [x] New behavior verified end-to-end: a branch-added file with a genuine main-side duplicate is correctly surfaced as a zero-reference candidate after a real merge; a live reference elsewhere (including inside a file main only modified) is still caught and never silently excluded.
- [x] Default path unaffected: plain `sync` (flag absent) never captures a merge-base, never dispatches the gather script or subagent, and the skill test suite still asserts the unchanged fast path.
- [x] Three rounds of adversarial review (two parallel first-pass reviewers, one targeted re-review, one final hand-verified adversarial pass) found and fixed two High-severity bugs in the refs inbound-reference guard and one Important gap in the Edit-escalation commit path before merge.

Closes #510


skills/workflow-branch-sync/SKILL.md
```diff
@@ -111,13 +124,34 @@ After every file in `UNMERGED` has been resolved and staged:
...
+3. **Reason** (advisory, never mutates): dispatch the `swe-workbench:redundancy-assessor` subagent with the full `_REDUND_OUT` (every `CANDIDATE`/`MAIN_ADD` record) and the `MERGE_BASE..PRE_SYNC_HEAD` branch diff.
+4. **Validate before acting** — this is the load-bearing invariant of this step: the skill never trusts the subagent's free-text for an actionable path. Parse each `**Redundancy: AUTO-APPLY|ESCALATE|NONE** id=<n>` sentinel it emits, cross-check every `id` against the enumerated `CANDIDATE id=<n>` lines `redundancy-scope.sh` produced, and **reject any id the script did not enumerate** — never act on an agent-invented id. The actionable file path always comes from the script's own `CANDIDATE path=<p>` record for that id, never from the subagent's prose.
+5. **Tiered gate** — per validated finding:
+   - `NONE` → no action.
+   - `AUTO-APPLY` (script-confirmed whole-file candidate, `refs=0`) → `git rm <script-path>`, stage, then `git commit -m "[refactor] drop redundant <path>, superseded by main"`. List this commit explicitly in the Step 7 summary — that's the human-eyeball moment the never-auto-push gate exists for.
```
lugassawan: **High** | Issue: The "Validate before acting" step (item 4, line 141) cross-checks the sentinel `id` against the script's enumerated `CANDIDATE id=<n>` lines, but never re-derives that same id's `refs=<count>` from `_REDUND_OUT` before this AUTO-APPLY bullet executes the no-prompt `git rm` + commit. The `refs=0` precondition asserted here ("script-confirmed whole-file candidate, refs=0") is only enforced by the subagent's own prompt (advisory, not a hard gate) — a subagent that mislabels a `refs>0` or symbol-level candidate as AUTO-APPLY has that file removed and committed with no deterministic cross-check, which cuts against the acceptance criterion that judgment-required findings are "never applied silently."\n\n**Why it matters:** this is the one tier that mutates the tree and commits without asking the user — extending the existing "never trust agent free-text for a path/id" invariant to the tier label itself (not just the path/id) closes the gap.\n\n**Suggested fix:** in item 4, add an explicit deterministic re-check: for every `AUTO-APPLY id=<n>` sentinel, look up that id's `CANDIDATE ... refs=<count>` line from `_REDUND_OUT`; if `refs != 0`, downgrade to `ESCALATE` before the tiered gate runs, regardless of the subagent's label. Add a regression test asserting this cross-check language exists in Step 6.

lugassawan: Addressed in d7958874b6dd81d7b1065039e4443bb4df126573: Step 6 item 4 now re-derives the sentinel's own `refs=<count>` from redundancy-scope.sh's CANDIDATE line and downgrades AUTO-APPLY to ESCALATE whenever refs is nonzero, closing the gap where the tier label itself was trusted at face value. Added a regression test pinning this language.

PR #523, opened 2026-07-12
[fix] Diagnose + prevent EnterWorktree tracking desync (#497)

lugassawan:
## Summary

- `hooks/workflow_resume_hint.sh`: probe `git rev-parse --git-dir --git-common-dir` independently of the sidecar-checkpoint gate; fire a defensive "WORKTREE RE-ANCHOR ADVISORY" nudge whenever cwd resolves to a linked worktree — including same-path (sidecar `worktree_root` already matches live root) and sidecar-less resumes, not just the pre-existing path-mismatch case
- `skills/workflow-worktree-session/SKILL.md` (Mode C): reword the `ExitWorktree` no-op diagnosis from an assumed "confirms cd-entry" to an ambiguity-aware probe (git-dir/git-common-dir divergence + sidecar `worktree_root` match), stated without over-claiming — cd-entry produces identical git-state evidence, so the honest conclusion is "may have been lost to compaction," not a definitive cause; recovery is unchanged either way
- Same false-certainty reframe applied to the two duplicated variants: `skills/workflow-cleanup-merged/SKILL.md` (Step 3a + Common Mistakes) and `skills/workflow-development/templates/plan-workflow-section.md` (Exit bullet)
- `docs/workflow-state.md`: document that harness `EnterWorktree` session anchoring does not survive compaction (root cause of #497) and update the nudge's coverage-limit note

## Test Plan

- [x] `shellcheck hooks/workflow_resume_hint.sh` — 0 issues
- [x] `pytest tests/ -v` — 1683/1683 pass
- [x] Live smoke test: piped a `SessionStart(compact)` payload into the hook from inside an actual linked worktree with no sidecar — confirmed the standalone nudge fires; repeated from the main checkout — confirmed empty output + exit 0

- [x] Reproduced the original gap: a no-op sidecar-less/same-path resume inside a linked worktree previously produced no nudge (silent) — new tests fail against the pre-fix hook, pass after
- [x] Verified the fix via the regression case (main checkout, no sidecar → still silent, exit 0) so the broadened probe does not fire outside linked worktrees

Closes #497


docs/workflow-state.md
```diff
@@ -56,7 +56,7 @@ but different branches never share a checkpoint file.
...
...r the worktree re-anchor nudge:** The resume hook is compaction-only (`SessionStart` with `matcher: "compact"`). As of issue #497, the nudge fires independently of the sidecar gate: it probes `git rev-parse --git-dir --git-common-dir` directly and fires whenever cwd resolves to a linked worktree — including the *same-path* case (a fresh sidecar's `worktree_root` already matches the live root) and the *sidecar-less* case (no checkpoint exists at all, e.g. a workflow that never wrote one). Only when cwd is **not** a linked worktree does the hook stay silent. It **cannot** redirect to a *foreign* worktree it has no checkpoint for (e.g. a session that was always anchored in the wrong dir but never compacted) — path mismatch still only produces the `REQUIRED` variant when a sidecar records a differing `worktree_root`. It also does not fire on a cold fresh launch or session "continuation" (a non-compaction resume) — this remains a residual coverage limit, since the hook's `matcher: "compact"` only targets compaction-triggered `SessionStart` events. The worktree_root nudge is defense-in-depth; the primary guidance lives in `skills/workflow-worktree-session/SKILL.md` and the plan template.
```
lugassawan, line 59: **Medium** | Issue: Acceptance criterion #1 from issue #497 ("Session state persistence verified across context recompression / **continuation** events") is only partially met — this fix covers compaction only.

**Why it matters:** This is transparently disclosed right here in the diff, so it isn't a hidden defect — but it means #497's silent-desync failure mode can still occur, unfixed and undiagnosed, for a `--resume`/`--continue` session that never compacted. That's the same class of bug the ticket was filed for, just via a different trigger path.

**Suggested fix:** Either scope this explicitly as a follow-up issue (referencing #497) so it isn't lost, or extend `hooks.json`'s `SessionStart` matcher to also cover the resume/continuation case if harness support allows it.

_Flagged by `reviewer` during PR review._

hooks/workflow_resume_hint.sh
```diff
@@ -7,11 +7,25 @@
...
 # Fail-open: exit 0 unconditionally. A broken hook must never block startup.
 
+# Emits a standalone re-anchor nudge when cwd is a linked worktree and no usable
+# checkpoint exists for this branch (missing, stale, wrong version, or wrong
+# branch — all "nothing to resume" cases). Relies on bash's dynamic scoping to
+# read is_linked_worktree/root/branch from the calling main() invocation.
+maybe_standalone_advisory() {
+    [ "$is_linked_worktree" -eq 1 ] || return 0
+    local advisory
+    advisory="WORKTREE RE-ANCHOR ADVISORY: This session resumed inside a linked worktree at \`${root}\` (branch ${branch}) after compaction. No usable workflow checkpoint was found for this branch, but compaction may still have dropped EnterWorktree tracking even though your cwd is already correct. Call \`EnterWorktree(path=${root})\` now to be safe — this is idempotent and harmless if you are still anchored. A later \`ExitWorktree\` no-op is consistent with this, though not proof — cd-entry looks identical."
```
lugassawan: **Low** | Issue: Three near-identical `WORKTREE RE-ANCHOR ADVISORY` string literals (this line, plus lines 153 and 158) duplicate wording almost verbatim.

**Why it matters:** Not a bug, but future wording tweaks require updating three places in lockstep; a mismatch between them would be easy to miss in review.

**Suggested fix:** Optional — extract the common advisory body into a small `format_advisory()` helper parameterized by the one varying clause ("no usable checkpoint" vs "matches the recorded worktree_root" vs generic).

_Flagged by `reviewer` during PR review._

docs/workflow-state.md
lugassawan: Addressed: filed #524 to track the continuation/cold-resume coverage gap as an explicit follow-up, since extending the compaction-only matcher is a separate scope decision.

hooks/workflow_resume_hint.sh
lugassawan: Addressed in b31229d: extracted the three near-duplicate advisory strings into a shared format_advisory() helper parameterized by the one varying clause.

Issue #370, opened 2026-05-29
[feat] New skill: integration-truewatch


lugassawan:
## Product framing

- **Problem.** Engineers using TrueWatch for APM/observability have no way to pull TrueWatch context (traces, metrics, service health) into the swe-workbench session — they must context-switch to the TrueWatch UI.
- **Value.** A TrueWatch trace ID or service identifier injects observability context directly into performance and incident investigation sessions for teams not on Datadog.
- **Acceptance criteria.** A TrueWatch integration skill: accepts trace ID or service identifier; fetches relevant metrics/traces via the TrueWatch API; formats a structured context block consistent with `integration-datadog`'s output shape.
- **Impact / Effort.** Impact: S/M — TrueWatch serves teams not on Datadog; overlaps with `integration-datadog` in purpose. Effort: S — follows same pattern as `integration-datadog` if TrueWatch exposes a REST API.

> **Redundancy note:** `integration-truewatch` and `integration-datadog` serve the same observability-context purpose — one per provider. Keep both if your team uses TrueWatch; prune one if you prefer a single observability integration.
> **Naming decision:** Using `integration-truewatch` (the `integration-*` prefix proposed in #335's fix). See #335 for the naming bug and #182 for the broader MCP tool surface RFC.

## Goal
Ship `integration-truewatch` so TrueWatch observability context is injectable into performance and incident investigation sessions, mirroring `integration-datadog`.

## Motivation
Teams not on Datadog still need observability context injection. TrueWatch provides APM/metrics in the same problem space; a dedicated skill serves those teams without requiring a platform switch.

## Surface
- [x] Skill — `integration-*`

## Sketch (optional)
Trigger: TrueWatch trace ID or service name token. Fetch: TrueWatch API (trace details, service health snapshot, recent anomalies). Returns: trace summary, latency breakdown, error rate, service health, recent alerts. Output shape mirrors `integration-datadog`.

## Acceptance (optional)
- [ ] Trigger: TrueWatch trace ID or service identifier
- [ ] Fetches: trace details, metrics, service health, recent alerts
- [ ] Output shape consistent with `integration-datadog`
- [ ] Pairs with `workflow-performance-investigation` and `workflow-incident-response`
- [ ] `scripts/validate.py` passes
- [ ] Links #335 (naming) and #182 (MCP surface) in skill frontmatter or comments

lugassawan:
Superseded by #511 — same consolidation (niche vendor mirroring Datadog).

Issue #365, opened 2026-05-29
[feat] New skill: integration-linear


lugassawan:
## Product framing

- **Problem.** Engineers working against Linear issues (e.g. `ENG-123`) have no way to pull issue context (title, description, acceptance criteria, status, assignee) into the swe-workbench session — they must context-switch to the Linear UI.
- **Value.** Direct parallel to `ticket-context`'s Jira support: Linear issue context is injected automatically when a `ENG-123`-style token appears, giving Claude full issue context without a tab switch.
- **Acceptance criteria.** A Linear integration skill: detects `<TEAM>-<N>` tokens; fetches title, description, acceptance criteria, status, and recent activity via the Linear API; formats a structured context block matching `ticket-context`'s output shape.
- **Impact / Effort.** Impact: M — Linear is among the most-used issue trackers in startups and scaleups. Effort: M — API integration, credential management.

> **Naming decision:** Using `integration-linear` (the `integration-*` prefix proposed in #335's fix) rather than the bare-name `linear-context` pattern of `ticket-context`. See #335 for the naming bug and #182 for the broader MCP tool surface RFC.

## Goal
Ship `integration-linear` so Linear issue context (`ENG-123`) is automatically injected into swe-workbench sessions, mirroring `ticket-context`'s Jira support.

## Motivation
Linear is the dominant issue tracker for engineering teams at startups and scaleups. The Jira path in `ticket-context` serves enterprise; Linear serves the startup/growth segment.

## Surface
- [x] Skill — `integration-*`

## Sketch (optional)
Trigger: `[A-Z]+-\d+` token pattern (configurable team prefix). Fetch: Linear GraphQL API (`issue(id: "ENG-123")`). Returns: title, description, state, assignee, priority, labels, recent comments. Output shape mirrors `ticket-context`.

## Acceptance (optional)
- [ ] Trigger token: `<TEAM>-<N>` (e.g. `ENG-123`)
- [ ] Fetches: title, description, acceptance criteria, status, assignee, labels, recent activity
- [ ] Output matches `ticket-context` context-block shape
- [ ] Credential/API token config documented
- [ ] `scripts/validate.py` passes
- [ ] Links #335 (naming) and #182 (MCP surface) in skill frontmatter or comments

lugassawan:
Superseded by #511 — Linear is a tracker; folds into `ticket-context`, not a new skill.

PR #525, opened 2026-07-12
[feat] Generalize external-service context injection by category

lugassawan:
## Summary
- Generalizes external-service context injection from a single vendor-specific skill (`ticket-context`) into a category-based "ports and adapters in prose" pattern — adding a provider is now an adapter-level change to an existing `*-context` skill, not a new skill.
- Adds `check_adapter_blocks()` to `scripts/validate.py`: every `skills/*-context/SKILL.md` must have a `## Adapters` section made of `### <Provider>` blocks, each carrying four bold-labeled fields (**Trigger** / **Fetch** / **Extract → block fields** / **Degrade**) in order. Also generalizes catalog routing so the whole `*-context` family (not just the literal `ticket-context`) routes to `agents/shared/workflows.md`.
- Restructures `ticket-context` into the new shape (behavior-preserving) and adds a **Linear** adapter, resolving the Jira/Linear key collision via URL-primary + availability tiebreak (a bare key stays Jira unless a Linear MCP is connected and Atlassian is not — zero regression to existing Jira routing).
- Adds **`observability-context`** (Sentry adapter) — every block carries a verbatim `**Evidence class:** production signal (Phase-1 framing) — NOT a profile` marker, contractually barred from being treated as profiling evidence.
- Adds **`comms-context`** (Slack permalink + PagerDuty incident adapters) — the first `*-context` skill with two adapters for one output kind, proving the same pattern `ticket-context` already uses for three trackers.
- Wires both new skills into `/swe-workbench:debug`'s correctness path (canonical ticket-context prelude line kept byte-identical — E3 parity), `workflow-performance-investigation` Phase 1 only, and tightens `performance-tuner`'s refusal predicate to explicitly exclude APM/observability dashboards as profile evidence.
- Documents the adapter-authoring recipe in `docs/extending.md` and adds `docs/catalog.md` entries for both new skills.
- **Supersedes #365–#370** (the six per-vendor follow-up issues for Linear/Sentry/Slack/PagerDuty/Datadog/TrueWatch) — those are closed as superseded by this category-based mechanism, not via closing keywords in this PR body, since adding the remaining providers (Datadog, TrueWatch) is now a follow-up adapter addition, not a new skill.

## Test Plan
- [x] `pytest tests/ -v` — 1682/1682 passing
- [x] `pytest tests/test_skill_triggers.py -v` — all trigger prompts (incl. new Linear/Sentry/Slack/PagerDuty fixtures) rank their skill #1 with ≥0.1 BM25 margin; fixed one IDF-drift regression in `ticket-context`'s Linear prompt caused by adding the new skills to the corpus
- [x] Behavioural spot-check: confirmed via direct regex/parsing trace that a `linear.app` URL routes to the Linear adapter while a bare `PROJ-123` key still routes to Jira; confirmed the `## Adapters` section-boundary detection correctly excludes headings outside it (added regression test)
- [x] Two independent whole-branch reviews (zero Critical/Important findings across both; the handful of Minor findings — an empty-`## Adapters` lint gap, a catalog doc-drift on the Linear mention, a boundary-detection test-coverage gap — were fixed in follow-up commits)

- [x] New-behaviour paths (Linear/Sentry/Slack/PagerDuty adapters, debug.md wiring) exercised via the trigger harness and manual regex tracing above

Closes #511


scripts/validate.py
```diff
@@ -451,14 +451,84 @@ def check_skill_trigger_fixtures():
...
+def _check_adapter_block_field_order(skill_md, provider, block):
+    """Confirm the 4 required bold-labeled fields appear in `block`, in order.
+
+    Searches for each label starting from just after the previous match, so a
+    label that is either absent or appears before an earlier label (i.e. out
+    of order) is reported the same way: as the first offending field.
+    """
+    cursor = 0
+    for label in _ADAPTER_FIELD_LABELS:
+        pat = re.compile(r'^[ \t]*-\s+\*\*' + re.escape(label) + r':\*\*', re.MULTILINE)
```
lugassawan, line 560: **Medium** — `check_adapter_blocks` scans raw text with `MULTILINE` regexes and never strips fenced code blocks. An adapter block containing a fenced example with a line shaped like `## X` or `- **Trigger:**` inside it would either truncate the section early (`_H2_BOUNDARY_RE`, line 457) or match a label inside the fence (this line, and the section-slicing at line 491), producing a false pass/fail.

Why it matters: this validator is the sole enforcement of the adapter contract, and `docs/extending.md` hands adapter authors a fenced template containing exactly these tokens (`### <Provider>`, `- **Trigger:**`). An author illustrating a recipe with a fenced snippet inside a real `skills/*-context/SKILL.md` would silently defeat the check. Not currently triggered by any of the three shipped skills (none embed a fence inside `## Adapters`).

Suggested fix: strip fenced (```) regions before applying the section-boundary and field-order regexes, so headings/labels inside illustrative examples are not scanned as structure.

scripts/validate.py
```diff
@@ -491,7 +561,7 @@ def _expected_slice(sid):
         for fname, prefixes in _SLICE_FILES.items():
             if any(sid.startswith(p) for p in prefixes):
                 return fname
-        if sid in _WORKFLOW_EXTRAS:
+        if sid in _WORKFLOW_EXTRAS or sid.endswith("-context"):
```
lugassawan: **Low** — `_WORKFLOW_EXTRAS = frozenset({"ticket-context"})` (line 540) is now redundant: the routing here is `sid in _WORKFLOW_EXTRAS or sid.endswith("-context")`, and `"ticket-context"` already ends with `"-context"`, so the frozenset can never be the deciding factor in a routing decision.

Why it matters: dead branch — harmless today, but a latent confusion for the next editor, who may assume the set carries a case the `endswith` check does not.

Suggested fix: drop the frozenset and the `sid in _WORKFLOW_EXTRAS or` clause, or keep the set only if you foresee a future non-`-context` workflow extra (and add a comment saying so). No behavior change either way.

scripts/validate.py
lugassawan: Addressed in 87a74cab4e9653acbef11fe2eda6ba3939e74195: added _strip_fenced_code_blocks() and apply it before any structural regex runs in check_adapter_blocks() — a fenced example inside ## Adapters can no longer truncate the section early or masquerade as a real field. Two regression tests added (masked heading, masked field label).

scripts/validate.py
lugassawan: Addressed in 87a74cab4e9653acbef11fe2eda6ba3939e74195: dropped the redundant _WORKFLOW_EXTRAS frozenset and the 'sid in _WORKFLOW_EXTRAS or' clause — routing is now just sid.endswith("-context"), no behavior change.

scripts/validate.py
```diff
@@ -451,14 +451,105 @@ def check_skill_trigger_fixtures():
...
+_ADAPTER_FIELD_LABELS = ("Trigger", "Fetch", "Extract → block fields", "Degrade")
+_ADAPTERS_HEADING_RE = re.compile(r'^##(?!#)[ \t]+Adapters[ \t]*$', re.MULTILINE)
+# Next top-level ('##', not '###+') heading — bounds the '## Adapters' section.
+_H2_BOUNDARY_RE = re.compile(r'^##(?!#)[ \t]+\S', re.MULTILINE)
+# One '### <Provider>' sub-heading starts one adapter block.
+_H3_PROVIDER_RE = re.compile(r'^###(?!#)[ \t]+(.+?)[ \t]*$', re.MULTILINE)
+# A fenced code block: opening ``` or ~~~ (3+) at some indent, any content,
+# then a closing fence of the SAME character and indent. DOTALL so '.' spans
+# newlines; non-greedy so an unrelated later fence doesn't get swallowed.
+_FENCE_RE = re.compile(r'^([ \t]*)(`{3,}|~{3,})[^\n]*\n.*?^\1\2[ \t]*$', re.MULTILINE | re.DOTALL)
```
lugassawan: **Low** — `_FENCE_RE` requires the closing fence to be a byte-identical backreference (`\2`) to the opening run and assumes LF; it does not strip when the close is longer than the open (CommonMark-legal, e.g. open ` + "```" + ` / close ` + "````" + `) or when the file uses CRLF.

Why it matters: an adapter author whose fenced example closes with a longer fence, or whose `SKILL.md` is CRLF, would see the original false-positive resurface as a spurious lint failure. Impact is a visible, self-correcting CI red — never a silent bypass (confirmed via adversarial check: no way for a genuinely malformed block to be masked by this gap). The CRLF assumption is already shared by every structural regex in this checker (`_ADAPTERS_HEADING_RE` etc.), so it introduces no new inconsistency.

Suggested fix: non-blocking/optional. If tightened later, allow the close run to be >= the open run length (not expressible via a plain backreference — would need a width-aware substitution) and normalize CRLF before scanning.

lugassawan: Addressed in 9e0b594: rewrote _strip_fenced_code_blocks() as a line-based scan that requires the closing fence to have >= the opening fence's length (not an exact backreference), and strips a trailing '\r' before matching so CRLF files work. Two regression tests added (longer-close fence, CRLF).

.gitignore
```diff
@@ -11,3 +11,4 @@ scripts/__pycache__/
 .venv/
 .rimba/settings.local.toml
 .claude/cache/
+.superpowers/
```
lugassawan, line 14: **Low** — this PR adds `.superpowers/` to `.gitignore`, unrelated to #511's category-context-injection scope.

Why it matters: minor scope creep — a feature PR mixing in an incidental tooling-ignore makes the diff slightly less atomic. Harmless and non-blocking; zero runtime impact.

Suggested fix: optional. Could move to a separate housekeeping commit, or leave as-is.

PR #534, opened 2026-07-14
[fix] sync-and-verify: detect subtree-wipe + split adaptive watchdog budget (#532)

lugassawan:
## Summary

- Fix watchdog-kills-hook race: split `SYNC_TIMEOUT` into an adaptive `PULL_BUDGET` + `HOOK_BUDGET` (scaled by `$HEAD_REF`'s own tracked-file count, capped at 570s total) so the rimba post-merge hook isn't starved of its cleanup window on large worktrees. An explicit `SYNC_TIMEOUT` override still wins verbatim, preserving existing pinned test behavior.
- Fix `HOOK_INTERRUPTED` false negative: add a targeted probe on `$HEAD_REF`'s own worktree that flags a ≥90% tracked-file deletion ratio (root intact, subtree wiped) — a state the existing missing-root-directory scan cannot see, and for which the existing `git worktree prune` recovery advice is a no-op.
- Update `SKILL.md` to document the required ~600000ms Bash-tool timeout for the Step 3b invocation (must exceed the script's ~570s internal cap) and the new intact-root/wiped-subtree recovery path (`git -C <wt> restore .` or re-run `rimba clean --merged --force`).
- Validate the new `PULL_TIMEOUT` env var the same way `SYNC_TIMEOUT` already is (review finding: a non-numeric value previously crashed the script with an unbound-variable error and zero stdout output).
- Address first-round review feedback: document `HOOK_BUDGET`'s single-worktree scope limit, treat a failed `git status` probe as inconclusive rather than silently clean, treat `SYNC_TIMEOUT=""` as invalid input like other malformed values, and add an integration test for the combined `TIMED_OUT` + `SUBTREE_WIPED` scenario.

## Test Plan

- [x] `pytest tests/ -v` — 1778/1778 pass, including 7 new tests across both review rounds (subtree-wipe detected, no false positive on a lightly-dirty worktree, adaptive budget scales with tracked-file count, non-numeric `PULL_TIMEOUT` falls back gracefully, empty `SYNC_TIMEOUT` falls back gracefully, a failed subtree probe is flagged inconclusive rather than silently clean, combined `TIMED_OUT`+`SUBTREE_WIPED` message-selection precedence)
- [x] `shellcheck skills/workflow-cleanup-merged/scripts/sync-and-verify.sh` — 0 issues

- [x] Reproduced the bug via new red tests before the fix (confirmed both false-negative and crash reproduce against the pre-fix script)
- [x] Verified the fix against the reproduction (tests green post-fix)
- [x] Regression-checked the existing pinned watchdog/hook-interrupted tests remain byte-for-byte unchanged in behavior

Closes #532


skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -14,21 +14,67 @@ MAIN_REPO=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
...
+  case "$PULL_BUDGET" in
+    ''|*[!0-9]*)
+      echo "sync-and-verify: invalid PULL_TIMEOUT='$PULL_BUDGET' (must be a non-negative integer) — falling back to 90" >&2
+      PULL_BUDGET=90
+      ;;
+  esac
+  K=50
+  HOOK_CAP=480
+  TOTAL_CAP=570
+  HOOK_BUDGET=$(( (HEAD_WT_FILES + K - 1) / K ))
```
lugassawan, line 79: **Medium:** `HOOK_BUDGET` is computed only from `$HEAD_REF`'s own tracked-file count, but the post-merge hook it's budgeting for is `rimba clean --merged --force`, which sweeps **all** currently-merged worktrees in one invocation, not just the one being processed. `orchestration.md` drives this script in a loop, one iteration per merged PR — on the first iteration of a multi-PR cleanup round, the single `git pull` can trigger the hook to clean every merged worktree at once, but the watchdog budget only accounts for the current branch's files. That's arguably closer to the real-world trigger for #532 (a large monorepo with many merged worktrees) than a single oversized worktree.

Suggested fix: either sum tracked-file counts across all worktrees whose branch is already merged into `$DEFAULT_BRANCH` (not just `$HEAD_REF`'s), or document explicitly that the adaptive budget is scoped to a single-worktree cleanup and is known-insufficient for bulk `--merged` sweeps.

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -81,8 +127,40 @@ while IFS= read -r _line; do
...
+# subtree of tracked files underneath it is wiped — the loop above sees
+# `[ -d "$_wt" ]` succeed and HOOK_INTERRUPTED stays a false 0, and
+# `git worktree prune` (the recovery advice below) is a no-op for this shape:
+# prune only clears missing *directories*, never files missing inside a live
+# one. Ratio, not "any deletion", so a worktree that's just legitimately dirty
+# doesn't get flagged as interrupted.
+SUBTREE_WIPED=0
+if [ -n "${HEAD_WT:-}" ] && [ -d "$HEAD_WT" ] && [ "${HEAD_WT_FILES:-0}" -gt 0 ]; then
+  _dels=0
+  if _d=$(git -C "$HEAD_WT" status --porcelain 2>/dev/null | awk '
```
lugassawan, line 153: **Medium:** `git -C "$HEAD_WT" status --porcelain 2>/dev/null` silently discards stderr. If this command errors (e.g. the worktree's `.git` pointer file or `.git/worktrees/<id>` admin metadata was itself among the files reaped before the kill landed — plausible depending on `rm` traversal order/timing), the `if _d=$(...); then` guard fails, `_dels` stays `0`, and the ratio evaluates to `0%` — a silent false negative in the exact probe this PR adds to catch partial deletion, with no diagnostic hint that the check itself failed to run.

Suggested fix: on failure of the `git status` probe, treat it as inconclusive-but-suspicious (e.g. still flag `HOOK_INTERRUPTED=1` with a distinct "could not verify worktree state" message) rather than silently defaulting to "0% deleted."

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
```diff
@@ -14,21 +14,67 @@ MAIN_REPO=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
...
-    SYNC_TIMEOUT=90
-    ;;
-esac
+#
+# TOTAL_CAP (570s) below is coupled to the Bash-tool-call timeout the caller
+# (SKILL.md Step 3b) must pass when invoking this script (~600s). The internal
+# watchdog MUST fire before that external timeout, or the external kill lands
+# first and Block D never runs — the script cannot self-enforce this invariant;
+# it is documented here AND in SKILL.md Step 3b.
+if [ -n "${SYNC_TIMEOUT:-}" ]; then
```
lugassawan: **Low:** Behavior change for `SYNC_TIMEOUT=""` (explicitly set to an empty string, vs. unset): previously `${SYNC_TIMEOUT:-90}` treated empty-as-unset and fell back to `90`. Now `[ -n "${SYNC_TIMEOUT:-}" ]` is false for an empty string, routing it into the *adaptive* budget path instead of the flat 90s fallback — a silent behavior change for an edge case with no test coverage (`test_non_numeric_sync_timeout_falls_back_to_default` only exercises `SYNC_TIMEOUT="90s"`, not `""`).

Low risk in practice (empty-string env vars are rare), but worth a one-line test or comment acknowledging the intentional change, or treating empty explicitly like the invalid-input case.

tests/test_sync_and_verify.py
```diff
@@ -353,6 +353,162 @@ def test_non_numeric_sync_timeout_falls_back_to_default(self, git_repo):
...
+
+        bulk_files = sorted(wt_path.glob("bulk*.txt"))
+        for f in bulk_files[:2]:  # 2/100 = 2% deletion ratio — legitimately dirty, not interrupted
+            f.unlink()
+
+        result = _run_script(git_repo, BRANCH, default_branch="main")
+        _assert_contract(result, "0", hook_interrupted="0")
+
+
+class TestSyncAndVerifyAdaptiveBudget:
```
lugassawan, line 600: **Low (test gap):** No test exercises the combined scenario #532 actually reports: the internal watchdog killing the hook *mid-rm on the very worktree it's cleaning* (`TIMED_OUT=1` and `SUBTREE_WIPED=1` on the same `HEAD_WT`), which determines message precedence (the script prioritizes the subtree message over the timeout message). Existing tests cover the two mechanisms independently: `TestSyncAndVerifySubtreeWipeProbe` synthesizes file deletion directly (no watchdog involved), and `TestSyncAndVerifyWatchdog.test_watchdog_timeout_concurrent_with_hook_interruption` combines the watchdog with a missing-root *stray* worktree (a different branch, not `$HEAD_REF`'s own).

Suggested fix: add an integration test using a stub `post-merge`-style hook (similar to the existing `sleep 30` stub pattern) that partially deletes `$HEAD_REF`'s own worktree files before being reaped by the watchdog, asserting the subtree message (not the generic timeout message) wins.

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
lugassawan: Addressed in b3452f9: documented the single-worktree scope limit of HOOK_BUDGET in a code comment (summing across all merged worktrees would add real complexity for a scenario orchestration.md already runs sequentially per-PR, so this is a documented known limitation rather than a behavior change).

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
lugassawan: Addressed in b3452f9: a failed git status probe (e.g. the worktree's own admin metadata reaped mid-flight) now flags HOOK_INTERRUPTED=1 with a distinct "could not verify" message, instead of silently defaulting to 0% deleted. Added an integration test that reproduces the failure via a stub git that removes the worktree's .git pointer mid-pull.

skills/workflow-cleanup-merged/scripts/sync-and-verify.sh
lugassawan: Addressed in b3452f9: presence is now checked via ${SYNC_TIMEOUT+set} instead of ${SYNC_TIMEOUT:-}, so an explicit-but-empty SYNC_TIMEOUT hits the same invalid-input case guard as "90s" and falls back to 90 with a warning, instead of silently taking the adaptive path. Added a regression test.

tests/test_sync_and_verify.py
lugassawan: Addressed in b3452f9: added test_watchdog_timeout_on_own_worktree_subtree_wipe_wins_message, which reproduces the exact #532 scenario (TIMED_OUT=1 and SUBTREE_WIPED=1 both true on the same worktree, via a stub git that wipes 95/100 tracked files then hangs) and asserts the subtree restore message wins over both prune-based messages.

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -86,21 +86,28 @@ _SCRIPTS="$_RT/skills/workflow-cleanup-merged/scripts"
...
+1. **Missing-root scan** — probes `git worktree list --porcelain` for any registered worktree whose *top-level directory* is missing on disk — the signal that a `post-merge` hook `rm` (this run's or a prior one's) was interrupted mid-deletion before reaching the worktree's own root removal, leaving the worktree registration and branch ref alive in `.git` while the whole directory is gone.
+2. **Targeted subtree-wipe probe (#532)** — checks `$HEAD_REF`'s own worktree specifically: if its root directory still exists but ≥90% of its tracked files (by `git status --porcelain` deletion count vs. the `git ls-files` total) are gone, the hook was interrupted *after* deleting most files but *before* removing the root itself. The missing-root scan above cannot see this case — the directory is still there. A ratio threshold (not "any deletion") avoids flagging a worktree that's just legitimately dirty. Both checks are state-based rather than event-based on purpose: an external kill takes down this script too, so it cannot reliably observe its own interruption — but a stale registration or a wiped subtree on the next run always tells the truth.
```
lugassawan: **Low:** The new `SUBTREE_PROBE_FAILED` ("could not verify") state — added in the latest commit to fix prior finding #2 — has no corresponding entry here or in the Failure Mode Table / Recovery Example sections. SKILL.md still describes only "two checks" and documents two `HOOK_INTERRUPTED=1` causes (root-missing, subtree-wiped); the third case (probe itself failed) isn't mentioned anywhere.

An on-call engineer who hits the new "could not verify $HEAD_REF's worktree state ... probe itself failed" stderr message during a live incident (the exact scenario prior finding #2 targeted) won't find a matching row in the Failure Mode Table, and may misapply the "root missing" (`git worktree prune`, a no-op here) or "subtree wiped" (`restore .`, which needs functioning git metadata) guidance instead. Stays Low rather than Medium since the inline stderr message does carry ad-hoc advice ("inspect $HEAD_WT manually").

Suggested fix: add a third bullet under "Hook-interruption detection" (probe-failure / inconclusive state), a matching Failure Mode Table row, and optionally a short Recovery Example — mirroring how `SUBTREE_WIPED` was documented in this same diff.

lugassawan: Addressed in ebe7ea9: added a third Hook-interruption-detection bullet, a Step 4 sub-bullet, and a Failure Mode Table row documenting the probe-failure ("could not verify") state alongside the existing root-missing and subtree-wiped cases.

PR #538, opened 2026-07-18
[feat] workflow-pr-review-post: pending-review posting, byline off inline, public-only remark

lugassawan:
## Summary

- Inline PR-review comments no longer carry the reviewer byline/remark on any repo, public or private (issue #531) — enforced structurally (`comments[]` bodies come from `finding.body` verbatim) and by an explicit input-contract clause + regression test.
- Migrated inline posting from a per-finding `POST /pulls/{n}/comments` loop to an atomic pending-review model: findings are pre-validated against the diff, assembled into a `comments[]` JSON array via `jq`, and posted in one `POST /pulls/{n}/reviews` call (`gh api --input -`, not bracket-indexed field flags — those build a stringified-key object, not an array, and GitHub rejects it).
- The ` ([swe-workbench](https://github.com/lugassawan/swe-workbench))` remark in the review summary byline now appends only when the target repo is confirmed public (`gh repo view --json isPrivate` returns `"false"`); private/unknown-visibility repos omit it (fail-safe). Callers now author an identity-only `BYLINE`; the core owns the remark as a single constant.
- Self-review now submits `event=COMMENT` (GitHub blocks self-`APPROVE` but allows self-`COMMENT`) instead of skipping submit entirely, so a self-authored review's inline/pr-level comments still land under the new atomic model.
- On a 422 from the atomic POST, the core re-fetches `HEAD_SHA`, re-validates/reassembles against it, and retries once; a second failure (or a non-422 failure, never blind-retried) falls back to posting each comment individually.

## Test Plan

- [x] `pytest tests/ -v` — 1829/1829 pass, including new regression coverage for the #531 byline-removal, public-only-remark gating, self-review `event=COMMENT`, empty-`comments[]` fall-through, and atomic-422-fallback contracts

- [x] New behaviour verified end-to-end via TDD (failing assertions written first, then the `SKILL.md` prose edited to match) and two independent code-review passes (plan-alignment + diff correctness/security/design), including a live verification that `gh api --input -` with a `jq`-built JSON payload correctly produces the `comments[]` array (the original bracket-indexed `-f` approach was caught and fixed in review — it silently builds a JSON object, not an array).

Closes #531

### DIFF
--- a/skills/workflow-pr-review-post/SKILL.md
+++ b/skills/workflow-pr-review-post/SKILL.md
@@ -200,8 +200,11 @@ if [ "$N" -gt 0 ]; then
 
     PAYLOAD=$(jq -n --arg commit_id "$HEAD_SHA" --arg event "$EVENT" --arg body "$SUMMARY" --argjson comments "$COMMENTS_JSON" \
       '{commit_id: $commit_id, event: $event, body: $body, comments: $comments}')
-    RESP2=$(gh api --method POST "repos/${OWNER}/${REPO}/pulls/${PR}/reviews" --input - <<<"$PAYLOAD" 2>&1)
-    [ $? -eq 0 ] && { posted_inline=$N; SUBMITTED=true; }
+    RESP2=$(gh api --method POST "repos/${OWNER}/${REPO}/pulls/${PR}/reviews" --input - <<<"$PAYLOAD" 2>&1); RESP2_OK=$?
+    # Retry failure is the same ambiguous class one level deeper — reuse the read-your-write check.
+    [ "$RESP2_OK" -ne 0 ] && LANDED=$(gh api "repos/${OWNER}/${REPO}/pulls/${PR}/reviews" \
+      -q "[.[] | select(.user.login==\"$CURRENT_USER\" and .commit_id==\"$HEAD_SHA\")] | length")
+    { [ "$RESP2_OK" -eq 0 ] || [ "${LANDED:-0}" -gt 0 ]; } && { posted_inline=$N; SUBMITTED=true; }
   else
     # Network/5xx — ambiguous: the POST may have landed server-side even though the
     # client saw a failure. Never blind-retry (no idempotency key); instead confirm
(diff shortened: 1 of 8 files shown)


skills/workflow-pr-review-post/SKILL.md
```diff
@@ -136,30 +136,104 @@ if [ "$DECISION" = "COMMENT" ] && [ "$BLOCKING_SCOPE" = "OUT-OF-DIFF-ONLY" ] \
...
+    # unchanged); demote anything newly out-of-diff to a pr-level comment.
+    HEAD_SHA=$(gh pr view "$PR" --json headRefOid -q .headRefOid)
+    COMMENTS_JSON="[]"; STILL_IN_DIFF=()
+    for row in "${INLINE_SURVIVORS[@]}"; do   # each row resolves REPO_PATH, LINE, BODY
+      if line_is_in_diff "$REPO_PATH" "$LINE" "$HEAD_SHA"; then   # same predicate as Step 2's pre-validate pass
+        COMMENTS_JSON=$(jq --arg path "$REPO_PATH" --argjson line "$LINE" --arg body "$BODY" \
+          '. + [{path: $path, line: $line, side: "RIGHT", body: $body}]' <<<"$COMMENTS_JSON")
+        STILL_IN_DIFF+=("$row")
+      else
+        gh pr comment "$PR" --body "$BODY" && posted_pr_level=$((posted_pr_level + 1))
```
lugassawan: **Medium**: The 422-retry and model-A-fallback demotion paths post each newly out-of-diff/failed finding as its own separate `gh pr comment` call, one per row, instead of batching them into a single comment.

Step 2 establishes "Batch ALL pr-level findings into **one** general PR comment" specifically to avoid PR clutter — the same goal #531 is chasing for inline bylines. If more than one finding is demoted in the same run, the PR ends up with N separate general comments instead of one.

Suggested fix: accumulate demoted rows into a buffer during the loop and issue a single `gh pr comment` after the loop, mirroring Step 2's PR-level batching pattern.

skills/workflow-pr-review-post/SKILL.md
```diff
@@ -136,30 +136,104 @@ if [ "$DECISION" = "COMMENT" ] && [ "$BLOCKING_SCOPE" = "OUT-OF-DIFF-ONLY" ] \
...
+    BYLINE_FULL="${BYLINE}${REMARK}. Posted ${N} inline comment(s) and ${posted_pr_level} PR-level note(s), deduped ${deduped}."
+    if [ "$IS_SELF_REVIEW" = true ]; then SUMMARY=$(printf '%s\n' "$BYLINE_FULL")
+    else SUMMARY=$(printf '**Review Decision: %s**\n\n%s\n' "$DECISION" "$BYLINE_FULL"); fi
+
+    PAYLOAD=$(jq -n --arg commit_id "$HEAD_SHA" --arg event "$EVENT" --arg body "$SUMMARY" --argjson comments "$COMMENTS_JSON" \
+      '{commit_id: $commit_id, event: $event, body: $body, comments: $comments}')
+    RESP2=$(gh api --method POST "repos/${OWNER}/${REPO}/pulls/${PR}/reviews" --input - <<<"$PAYLOAD" 2>&1)
+    [ $? -eq 0 ] && { posted_inline=$N; SUBMITTED=true; }
+  fi
+  # Any other failure (network, 5xx) is NOT retried — never blind-retry a POST
```
lugassawan: **Medium**: On a non-422 (network/5xx) failure of the atomic `POST /pulls/{n}/reviews`, the flow unconditionally falls through to the model-A per-comment fallback with no check for whether the atomic request actually landed server-side before the client saw the failure.

The doc explicitly reasons about this ambiguous-success class for the *same* endpoint (never blind-retry, no idempotency key) but does not apply the same caution to falling through to a *different* endpoint with the identical double-post effect.

Suggested fix: before falling to model-A on a non-422 failure, do a cheap read-your-write check (list reviews, filter by current user + recent timestamp) to confirm the atomic review did not actually land; only fall back if confirmed absent.

lugassawan (commented):

skills/workflow-pr-review-post/SKILL.md
lugassawan: Addressed in c8fd26362077e643e9f3233903d1a4b570513513: batched demoted rows into a single pr-level comment per loop (retry + fallback), mirroring Step 2's own batching pattern.

skills/workflow-pr-review-post/SKILL.md
lugassawan: Addressed in c8fd26362077e643e9f3233903d1a4b570513513: added a read-your-write check (list reviews, filter by current user + commit_id) before conceding to the model-A fallback on a non-422 failure.

skills/workflow-pr-review-post/SKILL.md
```diff
@@ -136,30 +136,113 @@ if [ "$DECISION" = "COMMENT" ] && [ "$BLOCKING_SCOPE" = "OUT-OF-DIFF-ONLY" ] \
...
+    [ "$DEMOTED_N" -gt 0 ] && gh pr comment "$PR" --body "$DEMOTED_BODY" \
+      && posted_pr_level=$((posted_pr_level + DEMOTED_N))
+    INLINE_SURVIVORS=("${STILL_IN_DIFF[@]}"); N=$(jq 'length' <<<"$COMMENTS_JSON")
+    BYLINE_FULL="${BYLINE}${REMARK}. Posted ${N} inline comment(s) and ${posted_pr_level} PR-level note(s), deduped ${deduped}."
+    if [ "$IS_SELF_REVIEW" = true ]; then SUMMARY=$(printf '%s\n' "$BYLINE_FULL")
+    else SUMMARY=$(printf '**Review Decision: %s**\n\n%s\n' "$DECISION" "$BYLINE_FULL"); fi
+
+    PAYLOAD=$(jq -n --arg commit_id "$HEAD_SHA" --arg event "$EVENT" --arg body "$SUMMARY" --argjson comments "$COMMENTS_JSON" \
+      '{commit_id: $commit_id, event: $event, body: $body, comments: $comments}')
+    RESP2=$(gh api --method POST "repos/${OWNER}/${REPO}/pulls/${PR}/reviews" --input - <<<"$PAYLOAD" 2>&1)
```
lugassawan: **Medium**: The 422-retry's second atomic POST (`RESP2`) has no read-your-write confirmation on failure — only the *first* attempt's non-422 branch does.

If the retry genuinely lands server-side but the client sees a transient network/5xx error reading the response, `[ $? -eq 0 ]` is false, `SUBMITTED` stays unset, and execution falls straight into the model-A fallback loop, which re-posts every comment in `INLINE_SURVIVORS` individually — duplicating every comment that actually landed via the retry. This is the same "ambiguous failure" class the sibling fix was written to close, just one level deeper (retry-of-retry).

Suggested fix: wrap the retry POST's failure path with the same read-your-write check used for the first attempt (reusing `$HEAD_SHA`/`$CURRENT_USER`, now refreshed) before falling to the model-A fallback.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment(s) and 0 PR-level note(s), deduped 0.

skills/workflow-pr-review-post/SKILL.md
lugassawan: Addressed in 9184410accd3a4a5bddfd6bf1dc70bdc28119d3f: extended the read-your-write check to the retry POST's failure path too, reusing the refreshed HEAD_SHA/CURRENT_USER — same protection the first attempt already had.

PR #539, opened 2026-07-18
[feat] cleanup-merged: PR-scoped residual sweep (Step 5)

lugassawan:
## Summary
- Add `skills/workflow-cleanup-merged/scripts/sweep-residuals.sh` — a PR-scoped backstop that force-removes leftover ephemeral rimba worktrees (`pr-review-<N>`, `pr-followup-<N>`, `address-feedback-<N>`, incl. bare-`N` `/tmp` fallbacks) and orphaned `/tmp/swe-workbench-*` state JSON for a specific merged PR, when `workflow-pr-review`/`-followup`/`workflow-address-feedback`'s own cleanup failed or was interrupted. Skips worktrees with uncommitted changes rather than force-removing them.
- Wire it into `workflow-cleanup-merged/SKILL.md` as a new automatic **Step 5 — Residual Sweep (PR-scoped)**, running after worktree removal and *before* branch deletion (now **Step 6**) — a stale `address-feedback` worktree checks out the PR's real head branch directly, so the sweep must clear it before branch deletion is attempted or `git branch -D` would be silently refused. Report renumbered to **Step 7**. Never touches the shared `/tmp/swe-workbench-*` parent directories and never force-deletes an `address-feedback` branch (may be the PR's real head branch).
- Update `docs/catalog.md` and `agents/shared/workflows.md` with a one-clause mention of the new backstop step.

## Test Plan
- [x]

- [x] New behavior covered by tests exercising real git repos / real `/tmp` fixtures (no mocking), incl. the PR-head-safety guarantee (`address-feedback` branch survives), bad-arg handling, nullglob state-file reap, and eval/cwd-trap safety
- [x] `shellcheck skills/workflow-cleanup-merged/scripts/sweep-residuals.sh` — 0 warnings
- [x] `bash scripts/validate.sh` — passes (SKILL.md at 299/300 lines, catalog complete, refs resolve)
- [x] `pytest tests/ -v` — 1852/1852 pass
- [x] Two independent reviews (`superpowers:requesting-code-review` whole-branch + `swe-workbench:reviewer` safety-focused) — both "Ready to merge: Yes", no Critical/Important findings

Closes #535


skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -126,14 +126,24 @@ The script always attempts the remote delete regardless of whether the local bra
 
 Capital `-D` is used for the local delete: squash-merged branches are not merge ancestors of `main`; lowercase `-d` would refuse.
 
-### Step 6 — Report
+### Step 6 — Residual Sweep (PR-scoped)
```
lugassawan: **High** | Why it matters | Suggested fix

New Step 6 (residual sweep) runs *after* Step 5 (branch deletion), but Step 6 is the only step that can free a branch that a stale `address-feedback-<N>` worktree is holding checked out. `address-feedback-<N>` worktrees check out the PR's real head branch directly (`git worktree add "$WT" "${PR_BRANCH}"` in workflow-address-feedback/SKILL.md:132) — the same branch Step 1 resolves as `<headRefName>` and Step 5's `delete-branches.sh` force-deletes via a bare `git branch -D "$HEAD_REF"`. If a stale address-feedback worktree from a prior interrupted run (exactly the scenario this PR targets) still holds that branch checked out, git refuses the delete, `delete-branches.sh` aborts before printing its `LOCAL_DELETED=`/`REMOTE_DELETED=` contract, and `eval "$(...)"` swallows the failure silently — while Step 6, which would actually free the branch by removing the stale worktree, has already been skipped. There's no documented recovery path (e.g. "run Step 6, then retry Step 5") in the Failure Mode Table. This defeats the backstop's own purpose in exactly the case it targets, and it's untested — `tests/test_sweep_residuals.py`'s address-feedback test uses a distinct throwaway branch, never the actual `<headRefName>` collision.

**Suggested fix:** run the worktree-removal portion of the sweep before Step 5's branch deletion, or have `delete-branches.sh` detect "checked out elsewhere" and add a Failure Mode Table row directing the agent to run Step 6 then retry Step 5.

skills/workflow-cleanup-merged/scripts/sweep-residuals.sh
```diff
@@ -0,0 +1,158 @@
...
+  # AND its basename is bare "<N>" / "<N>-followup", not the task label — so
+  # the scan above cannot find it. Check the known literal paths directly.
+  if [ -z "$PR_REVIEW_WT" ] && [ -d "/tmp/swe-workbench-pr-review/$N" ]; then
+    PR_REVIEW_WT="/tmp/swe-workbench-pr-review/$N"
+  fi
+  if [ -z "$PR_FOLLOWUP_WT" ] && [ -d "/tmp/swe-workbench-pr-review/$N-followup" ]; then
+    PR_FOLLOWUP_WT="/tmp/swe-workbench-pr-review/$N-followup"
+  fi
+
+  reap_one_worktree() {
```
lugassawan, line 125: **Medium** | Why it matters | Suggested fix

`reap_one_worktree` force-removes worktrees/branches (`git worktree remove --force`, `git branch -D`) with no dirty/unpushed check — unlike Step 4's primary-target flow (`probe-worktree.sh` aborts on `DIRTY > 0`/`UNPUSHED > 0`) and unlike this same skill file's own stated contract ("Does not force-delete branches with uncommitted work — no `--force`, ever."). The residual sweep fires specifically for *interrupted* runs, precisely when a worktree is most likely to hold uncommitted, unpushed edits (e.g. a killed workflow-address-feedback session mid-edit, before its own commit step). Reporting is quantity-only (`SWEPT_WORKTREES=<n>`) with no visibility into whether a swept worktree actually contained uncommitted diffs — this can silently and irreversibly destroy local-only work.

**Suggested fix:** before force-removing each matched worktree, run a cheap `git -C "$path" status --porcelain` check and, if non-empty, either skip that worktree (report separately for manual inspection) or emit a distinct stderr warning so the discard is visible instead of silent.

skills/workflow-cleanup-merged/SKILL.md
```diff
@@ -272,18 +282,7 @@ git branch -D <stale-branch-name>                # the local ref that survived t
 
 ### Recovery Example — Intact Root, Wiped Subtree (#532)
```
lugassawan, line 273: **Low** | Why it matters | Suggested fix

The diff also condenses the "Root intact, subtree wiped" Recovery Example into a pointer to the existing Failure Mode Table row. Content is preserved and accurate, so this isn't a correctness issue — just scope creep unrelated to issue #535's residual sweep (done to offset the file under the 300-line orchestrator cap per the plan). No action required if the author considers this in-scope cleanup.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 7c4ed4eb83364acd7e4781d545ab956aed2cfe52: reordered Step 5 (Residual Sweep) to run before Step 6 (Delete Branches), so a stale address-feedback worktree holding the PR's real head branch checked out is cleared before branch deletion runs. Added prose explaining why, plus doc cross-reference fixes throughout.

skills/workflow-cleanup-merged/scripts/sweep-residuals.sh
lugassawan: Addressed in 7c4ed4eb83364acd7e4781d545ab956aed2cfe52: reap_one_worktree now checks git status --porcelain before force-removing any matched worktree; if uncommitted changes are present it skips that worktree (with a stderr warning) instead of discarding local-only work. New regression test: TestDirtyWorktreeSkipped.

skills/workflow-cleanup-merged/SKILL.md
lugassawan: Addressed in 7c4ed4eb83364acd7e4781d545ab956aed2cfe52: restored the full Recovery Example — Intact Root, Wiped Subtree (#532) walkthrough (code block + explanation) rather than the condensed pointer. Reclaimed the line-cap budget via spacing tightening in the Step 5/6/7 block instead, so the file stays at exactly 300 lines.

PR #563, opened 2026-07-26
[feat] Make code-comment reduction a verified gate

lugassawan:
## Summary
- Adds `runtime/comment-scan.py`: a pure diff-in/findings-out scanner (unified diff on stdin, findings + footer on stdout, never exits non-zero). Four detectors — `OVER_CAP`, `COMMENTED_OUT`, `RESTATES` (must-triage) and `DENSITY` (informational) — across 13 languages/extensions, with stable `detector:file:line` ids.
- Wires the scan into `agents/code-impl.md`, `agents/debugger.md`, `agents/refactorer.md`, `agents/test-writer.md` verify steps and `skills/workflow-development/SKILL.md` Phase 3, via a single shared include (`agents/shared/comment-scan.md`) — not five copies of the same instruction.
- Enforcement is **verdict accounting**, not a hard gate: Phase 3 evidence must carry one `KEEP <id> <reason>` or `FIXED <id>` line per must-triage finding, with `FIXED` claims re-confirmed by a re-run.
- Pins `code-impl.md`'s existing status semantics: `DONE` requires a clean scan or all-`FIXED`; any `KEEP` downgrades to `DONE_WITH_CONCERNS`.
- `agents/reviewer.md` is **deliberately unchanged** — it also backs `workflow-pr-review`/`workflow-pr-review-followup` on external PRs where this script never ran; narrowing it would strip mechanical comment findings from that entire product to solve a Phase 3/4 dedup problem that's accepted as v1 duplication (follow-up: pass verdicts to the reviewer as suppression context).
- `runtime/README.md` documents the one `.py` script in a directory otherwise full of `.sh`; `runtime/doctor.sh` gains a `python3` probe.

## Calibration

Corpus: this repo's own recent Python/shell diffs (6 commits, ~2,100 lines) plus external Go (`spf13/cobra`, 7 commits), Rust (`BurntSushi/ripgrep`, 4 commits), and TypeScript (`sindresorhus/ky`, 6 commits) history — ~5,500 lines of real diff across 23 samples, hand-classified.

- **OVER_CAP** — deterministic count against `principle-clean-code`'s existing cap table, not independently precision-fitted (the threshold is the source of truth, not tuned). Spot-checked ~10 hits across shell/python/rust/go for correct line-counting and exclusion behavior — e.g. ripgrep's `pathutil.rs` (14-line rustdoc over an 8-line cap) and this repo's `bash_guard.sh` WHY-comments are all legitimate must-triage candidates, most of which would reasonably `KEEP`.
- **COMMENTED_OUT** — 2 real hits in the corpus: 1 confirmed true positive (ripgrep's `literal.rs`, genuinely dead `*self = GramQuery::from_set_or(set);`) and 1 accepted false-positive class (a comment citing example code verbatim — `env={**_CLEAN_ENV, ...}.` in this repo's `conftest.py` — indistinguishable from real dead code by text pattern alone; left for the `KEEP` mechanism to absorb). Found and fixed 3 additional false positives *during* calibration before reaching this state: `match`/`package` colliding with Rust/Go keywords when used as ordinary English verbs/nouns, and a 13-word sentence ending in `;`. Fixed via a word-count gate on keyword-first-word matches, a comma/prose-score gate on bare trailing punctuation, and (found in Phase 4 review) an RHS-shape gate on the assignment heuristic after `# area = pi * r squared` was flagged as dead code.
- **RESTATES** — 0 real hits across the ~5,500-line corpus (reviewed/merged code rarely contains an obviously-redundant comment — review usually catches it first). The 0.7 overlap threshold stays a conservative default guided by the plan's own worked example (`# Returns the user's id` / `return user.id;`), pending real `KEEP`/`FIXED` data from actual use.
- **DENSITY** — 0 real hits (no sampled diff crossed the 40-added-code-line floor at ≥50% ratio); informational only, no precision claim needed.

## Phase 4 review

Three rounds of review (plan-alignment + two rounds of diff-correctness) found and fixed 7 real, reproduced bugs in the scanner:
- rename-with-modification silently skipped **all** detectors for the whole file, not just the untouched-comment case — the most severe finding, since it defeated the gate on exactly the move+edit shape `refactorer`'s `-M` wiring depends on
- the leading-of-file `OVER_CAP` exemption blanket-suppressed doc comments too (a module docstring at line 1 could never be flagged — the single most common real over-cap case)
- a leading shebang broke module-docstring recognition
- `//!` (Rust/C# inner doc comments) wasn't recognized as doc-mode
- run-continuation checked only diff-view adjacency, not line-number contiguity, so unrelated far-apart hunks could merge into one bogus multi-hunk `OVER_CAP` span
- `is_generated()` scanned the whole file instead of the header, so an unrelated mid-file "DO NOT EDIT" comment could skip the whole file

Self-scanning this PR's own diff surfaces 10 must-triage `OVER_CAP` findings. Explicit verdicts (per-finding, not narrative — closing the dogfooding gap a follow-up review round flagged: shipping the gate without recorded verdicts anywhere durable for its own flagged blocks):

- `KEEP OVER_CAP:runtime/comment-scan.py:2` — module docstring; documents the whole scanner's philosophy/scope/failure-posture, mirrors `hooks/secret_guard.py`'s own long module docstring precedent.
- `KEEP OVER_CAP:runtime/comment-scan.py:41` — legend for `doc_mode`'s 5 string values; each encodes a real structural difference not evident from the field name alone.
- `KEEP OVER_CAP:runtime/comment-scan.py:136` — calibration provenance; records real precision numbers and what got tightened, so a future maintainer doesn't re-litigate constants blind.
- `KEEP OVER_CAP:runtime/comment-scan.py:174` — documents why the rename exemption is hunk-scoped, not file-wide (the bug a later review round found and this PR fixed).
- `KEEP OVER_CAP:runtime/comment-scan.py:254` — documents the generated-marker scan-window trade-off (see follow-up #2 below).
- `KEEP OVER_CAP:runtime/comment-scan.py:452` — explains why `_PY_DEF_OPEN_RE` exists alongside `_PY_DEF_RE` (the async-def/multi-line-signature fix below) — without it, a future reader could "simplify" by deleting the looser regex and reintroduce the bug.
- `KEEP OVER_CAP:runtime/comment-scan.py:508` — explains why the docstring gap-abort must not consume the boundary-crossing entry — a subtle index bug a shorter comment wouldn't have prevented on the next edit.
- `KEEP OVER_CAP:runtime/comment-scan.py:541` — documents the leading-of-file-exemption bug (blanket-suppressed doc comments) and why it's now doc-gated.
- `KEEP OVER_CAP:runtime/comment-scan.py:585` — documents the `# area = pi * r squared` false positive and why RHS shape is checked.
- `KEEP OVER_CAP:runtime/comment-scan.py:599` — documents the calibration-derived word-count gate with the real motivating example (`match the force pattern...`) that produced it.

**Known follow-ups (non-blocking, filed for v2 — this is an advisory-only tool, no security/data-loss path):**
1. The rename exemption matches by set membership, not occurrence count — a single removed line can exempt multiple new same-text additions *within the same hunk* (narrow: requires identical-text duplicate additions in one hunk).
2. Widening `_GENERATED_MARKER_SCAN_LINES` from 10 to 20 (to catch a generated marker after a license-header-length preamble) trades a narrower miss for a narrower false-suppression window on hand-written headers that happen to contain the phrase "DO NOT EDIT".

## Address-feedback round

A follow-up review pass on this PR found and fixed one more real bug plus a consistency gap:
- `_PY_DEF_RE` didn't match `async def` or a multi-line function signature, so a docstring under an async function (or one whose signature spans multiple lines) was invisible to the scanner entirely — not just exempt from `OVER_CAP`, uncounted for `DENSITY` too. Fixed: the regex now accepts `async def`, and a backward scan (`_preceded_by_def_or_class`) finds the declaration across an unclosed multi-line signature.
- `agents/debugger.md`, `agents/refactorer.md`, `agents/test-writer.md` were wired to run the comment scan in Verify, but (unlike `code-impl.md`) their Output contracts had no field naming where `KEEP`/`FIXED` verdict ids should surface — a real risk under summary-only consumption. Added an explicit line to each.

**A second followup review round then found the first fix was itself incomplete:** the backward scan used a fixed 7-line lookback cap, so a signature with 7+ one-per-line parameters (the default shape under black/ruff's magic-trailing-comma style) reproduced the exact original bug, just narrowed rather than closed. Fixed by replacing the fixed cap with a bracket-depth walk (track unmatched `)`/`]`/`}` vs `(`/`[`/`{` while scanning backward until depth resolves to zero at the declaration line), with a generous 50-line safety valve for pathological input rather than a correctness bound. The multi-line-signature fixture now uses a 7-parameter signature so this class of regression is caught by tests, not just adversarial review.

## Test Plan
- [x] `pytest tests/ -v` — 2117/2117 pass (27 new/extended tests for the scanner + wiring)

- [x] Piped a hand-built diff containing one instance of each detector into `python3 runtime/comment-scan.py`; confirmed findings, ids, and footer counts
- [x] Fixed one finding, re-ran, confirmed it was absent and `KEEP` findings persisted
- [x] Ran the scan against this PR's own diff — comes back with 9 defensible `KEEP`s (see above), no unaccounted must-triage findings
- [x] Verified against 23 real diffs from this repo + 3 external OSS repos (Go/Rust/TS) during calibration

Closes #546


lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 5 inline comment(s) and 0 PR-level note(s), deduped 0.

runtime/comment-scan.py
```diff
@@ -0,0 +1,813 @@
...
+                    has_example = any(_EXAMPLE_MARKER_RE.search(b) for b in block_bodies)
+                    runs.append(Run(block_lines, block_bodies, block_is_doc, has_example, boundary is not None and block_lines[0].lineno == boundary))
+        i += 1
+
+    return runs
+
+
+# ── Python docstring extraction ───────────────────────────────────────────────
+
+_PY_DEF_RE = re.compile(r"^\s*(def|class)\s+\w+.*:\s*$")
```
lugassawan: **High** — `_PY_DEF_RE` doesn't match `async def` or a multi-line function signature, so `_extract_python_docstrings`'s `preceding_ok` check fails and the docstring is never recognized at all — no OVER_CAP fires, and the lines don't even count toward the DENSITY tally.

**Why it matters:** `async def` is a mainstream, ubiquitous Python shape; this is a systematic false negative on a large share of real-world Python, undermining the acceptance criterion that comment density is enforced by a checkable gate.

**Suggested fix:** extend the regex to accept the `async` prefix, e.g. `r"^\s*(async\s+def|def|class)\s+\w+.*:\s*$"`. For the multi-line-signature case, track "inside an open signature" state so `preceding_ok` doesn't require the immediately-prior view line to itself be the declaration.

agents/debugger.md
```diff
@@ -49,7 +49,7 @@ Call this out even when the minimal fix does not address it. Silence signals the
 3. **Confirm root cause** — one sentence, backed by a concrete artifact.
 4. **Write the regression test first** — it must fail against current code for the stated reason.
 5. **Apply the minimal fix** — smallest diff that turns the test green. No bundled cleanups.
-6. **Verify** — full relevant test suite green. Note anything newly suspicious.
+6. **Verify** — full relevant test suite green. Note anything newly suspicious. Run the comment scan per @./shared/comment-scan.md and account for every must-triage finding (`KEEP <id> <reason>` or `FIXED <id>`).
```
lugassawan, line 52: **Medium** — the comment-scan step is wired into Verify here, but the `## Output contract` section below has no field for the required `KEEP <id> <reason>` / `FIXED <id>` verdict lines — unlike `code-impl.md`, which added a dedicated paragraph mapping these verdicts into its structured output.

**Why it matters:** under a summary-only consumption model, verdict lines with no designated home risk being dropped when this agent runs standalone (reachable via `/swe-workbench:debug` outside full workflow-development orchestration), silently defeating the must-triage accounting this gate depends on.

**Suggested fix:** add an explicit line to the Output contract naming where KEEP/FIXED ids surface, mirroring `code-impl.md`'s treatment.

agents/refactorer.md
```diff
@@ -28,7 +28,7 @@ You are a refactoring specialist. You improve structure without changing observa
 2. **Coverage audit.** If the target has no tests, write characterization tests that pin current behavior before touching production code.
 3. **Plan.** Emit an ordered list of moves from `swe-workbench:principle-refactoring`'s Fowler catalog.
 4. **Execute.** One step at a time. Run tests after each. Commit per step when practical.
-5. **Verify.** Run the full suite at the end. Diff the public API to confirm nothing external changed.
+5. **Verify.** Run the full suite at the end. Diff the public API to confirm nothing external changed. Run the comment scan per @./shared/comment-scan.md and account for every must-triage finding (`KEEP <id> <reason>` or `FIXED <id>`) — the `-M` rename detection it relies on matters here specifically, since this agent moves functions and their doc comments without rewriting them.
```
lugassawan, line 31: **Medium** — same gap: the comment-scan instruction here has no corresponding slot in the `## Outputs` section for KEEP/FIXED verdict ids.

**Why it matters:** same risk as the debugger.md finding — arguably more acute here, since `refactorer` is explicitly the agent most likely to trigger the rename-exemption path and therefore most likely to generate `KEEP` judgment calls that need to reach the orchestrator.

**Suggested fix:** fold verdict ids into "Post-execution verification report" explicitly by name, so it's not left to the agent's improvisation.

agents/test-writer.md
```diff
@@ -52,7 +52,7 @@ The boundary line: domain ↔ infrastructure is the only seam where test doubles
 3. Enumerate behaviours: happy path, boundaries, error paths. Skip pure plumbing covered by higher-level tests.
 4. Write the smallest test that fails for the right reason, then verify it passes against current code.
 5. Apply Arrange / Act / Assert with a blank line between sections.
-6. Run the relevant test command; report pass / fail.
+6. Run the relevant test command; report pass / fail. Run the comment scan per @./shared/comment-scan.md and account for every must-triage finding (`KEEP <id> <reason>` or `FIXED <id>`).
```
lugassawan, line 55: **Medium** — same gap: the comment-scan instruction here has no corresponding item in the numbered `## Output contract` (1–5) for verdict ids.

**Why it matters:** same risk as the other two agent-wiring findings.

**Suggested fix:** add a 6th output-contract item for comment-scan verdicts, consistent with the other three agents needing the same fix.

runtime/comment-scan.py
```diff
@@ -0,0 +1,813 @@
+#!/usr/bin/env python3
+"""Advisory comment-quality scanner for issue #546's verified gate.
```
lugassawan: **Low** — this module docstring (32 lines) exceeds this same PR's own python doc cap of 8 when self-scanned, along with 8 further inline-comment blocks elsewhere in this file.

**Why it matters:** non-blocking dogfooding gap — shipping the gate without `KEEP` verdicts recorded anywhere durable for its own script's largest doc comment and several inline blocks. Several of the flagged blocks are legitimate WHY-focused rationale earning their length (documenting the exact bugs three review rounds found), but that judgment isn't recorded as an explicit, greppable verdict anywhere.

**Suggested fix:** record explicit `KEEP <id> <reason>` verdicts in the PR body (the precision writeup already lists them narratively) — not required to block merge, but worth closing the loop for the flagship example of the new discipline.

runtime/comment-scan.py
lugassawan: Addressed in 4c781f8: extended `_PY_DEF_RE` to accept `async def`, and added a bounded backward scan (`_preceded_by_def_or_class` + `_PY_DEF_OPEN_RE`) that finds the declaration across an unclosed multi-line signature.

agents/debugger.md
lugassawan: Addressed in 4c781f8: added a comment-scan verdict slot to the Output contract, naming where KEEP/FIXED ids surface.

agents/refactorer.md
lugassawan: Addressed in 4c781f8: added a comment-scan verdict slot to the Outputs section (folded into Post-execution verification report).

agents/test-writer.md
lugassawan: Addressed in 4c781f8: added a 6th Output-contract item for comment-scan verdicts.

runtime/comment-scan.py
lugassawan: Addressed: recorded explicit KEEP <id> <reason> verdicts in the PR body for all 10 self-scan findings, closing the dogfooding gap — not a code change, so no fix SHA.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench))

Confirmed the 5 prior fixes (async def / multi-line signature recognition, output-contract slots on debugger/refactorer/test-writer, durable KEEP verdicts) — all hold correctly.

Found one more real bug during the followup pass: the multi-line-signature backward scan (`_preceded_by_def_or_class`) used a fixed 7-line lookback, so a signature with 7+ one-per-line parameters (the default shape under black/ruff's magic-trailing-comma style) reproduced the original bug — docstring invisible to the scanner, uncounted for OVER_CAP and DENSITY — just narrowed rather than closed.

Addressed in 796ce33: replaced the fixed cap with a bracket-depth walk (tracks unmatched `)`/`]`/`}` vs `(`/`[`/`{` while scanning backward until depth resolves to zero at the declaration line), with a generous 50-line safety valve for pathological input rather than a correctness bound. The multi-line-signature fixture now uses a 7-parameter signature so this class of regression is caught by tests going forward.

Also noted (Low, non-blocking): an earlier commit bundled a prose reflow of an unrelated paragraph into the same hunk as a feature addition in `skills/workflow-development/SKILL.md` — done to fit under the file's 300-line cap after the feature addition, not gratuitous reformatting. Not actioning further; noted for the record.

Posted 0 inline comment(s) and 1 PR-level note(s), deduped 0.

PR #564, opened 2026-07-26
[feat] Add echo-vs-printf shell footgun doc and CI enforcement

lugassawan:
## Summary
- Documents the `echo` vs `printf` shell footgun in `docs/shell-echo-vs-printf.md`: zsh (the user's default macOS login shell) expands backslash escapes in `echo`'s argument, silently corrupting JSON piped or redirected through it.
- Adds `scripts/validate.py::check_no_echo_var_hazard`, a new structural check that scans fenced ```bash blocks under `skills/`, `commands/`, `agents/` (including `reference/` subdirs) for a variable piped or redirected through `echo`, with a quote-aware scanner that correctly handles embedded `;`, `&&`/`||`/bare `&` command boundaries, fd-dup (`>&`) and `/dev/null` exemptions, and brace-group/case-arm command positions.
- Normalizes the 11 pre-existing sites across 6 files that had the hazard (`commands/extend.md`, and four `SKILL.md`/reference files under `skills/`) to `printf '%s'`/`printf '%s\n'`.
- Adds two `## Avoid` bullets to `skills/language-bash/SKILL.md` and three net-zero-line AC warnings to `skills/workflow-pr-review-post/SKILL.md` so the hazard is caught at both write-time (skill authoring) and CI-time.
- Following review, hardens the scanner against `&>`/`&>>` combined redirects, `$(...)`/backtick command substitution, and backslash line-continuations, and adds a companion `scripts/validate.py::check_no_printf_var_format` structural check: `printf "$VAR"` makes `$VAR` the format string itself — a distinct `%s`/`%n` format-string-injection risk from the `echo` hazard.

## Test Plan
- [x] `python3 -m pytest tests/ -q` — 2137/2137 pass
- [x] `bash scripts/validate.sh` — 0 issues (all 6 pre-existing violations found and fixed; live tree clean)
- [x] Reintroduced a real violation, confirmed `validate.sh` names the exact file/line, then reverted
- [x] Behavioral zsh spot-check reproduces the documented `jq` parse error for `echo`, and confirms `printf '%s'` parses clean
- [x] `lychee --offline` confirms all new relative doc links resolve

- [x] New behavior (`check_no_echo_var_hazard`) is covered by a dedicated test class (`TestCheckNoEchoVarHazard`, 20 cases) covering must-flag, must-not-flag, and quote-awareness/command-position hardening cases surfaced during review
- [x] Both 300-line-capped files (`workflow-pr-review-post/SKILL.md`, `workflow-address-feedback/SKILL.md`) verified at exactly 300 lines after edits

Closes #549


scripts/validate.py
```diff
@@ -1187,6 +1187,139 @@ def check_browser_tool_gate(cache=None):
...
+            i += 1
+            continue
+        if c in ('"', "'"):
+            quote = c
+            i += 1
+            continue
+        if c == ';':
+            return None
+        if c == '&':
+            return None  # bare '&' (background) or '&&' — both end the command
```
lugassawan, line 1275: **Severity: High**

`_find_echo_hazard_end` misses bash's `&>`/`&>>` combined redirect, letting a real echo-JSON hazard slip through undetected. `&` is checked before `>` in the scan loop, so `echo "$JSON" &> /tmp/payload.json` (a common idiom for capturing both stdout and stderr) is treated as a bare background job / `&&` and the scan returns `None` before ever reaching the real file redirect — verified directly: `_echo_hazard_in_line('echo "$JSON" &> /tmp/out.json')` returns `False`. This silently defeats the guard's purpose for a mainstream bash construct — the exact class of bug issue #549 exists to catch. Fix: before returning `None` on `&`, peek at the next char; if it's `>`, treat it like the existing `>` branch (advance past `&>`/`&>>`, then run the same target/`/dev/null` check) instead of terminating the scan. Add a regression test alongside the existing quote-aware hardening tests.

scripts/validate.py
```diff
@@ -1187,6 +1187,139 @@ def check_browser_tool_gate(cache=None):
...
+        window = line[m.end():]
+        hazard_end = _find_echo_hazard_end(window)
+        if hazard_end is None:
+            continue
+        if _VAR_REF_RE.search(window[:hazard_end]):
+            return True
+    return False
+
+
+def check_no_echo_var_hazard(cache=None):
```
lugassawan: **Severity: Medium**

The guard has no counterpart for the "dangerous fix" the sibling doc (`docs/shell-echo-vs-printf.md`) warns about — `printf "$VAR"` (variable used directly as the format string, a `%s`/`%n` format-string-injection risk) is named as forbidden in prose (this doc and `skills/language-bash/SKILL.md`) but nothing structurally lints for it. Consider a companion detector for `printf` invocations whose first non-flag argument is a bare `"$VAR"`/`${VAR}` rather than a literal format string, with test coverage matching `check_no_echo_var_hazard`.

scripts/validate.py
```diff
@@ -1187,6 +1187,139 @@ def check_browser_tool_gate(cache=None):
...
+# char; ')' covers a `case` arm ('a) echo ...'), '{' covers a brace group
+# ('{ echo ...; }'). An optional shell keyword (if/elif/then/else/do/while/
+# until) may sit between the position marker and 'echo' (e.g. 'elif echo ...
+# ; then') — the keyword itself must be at that same command position, so a
+# keyword mentioned in a trailing comment ('# if echo fails') still can't
+# reach this branch.
+_ECHO_CMD_POS_RE = re.compile(
+    r'(?:^|[;&|({)])\s*(?:(?:if|elif|then|else|do|while|until)\s+)?echo(?=\s|$)'
+)
+_VAR_REF_RE = re.compile(r'\$\{?[A-Za-z_][A-Za-z0-9_]*\}?')
```
lugassawan: **Severity: Medium**

`_VAR_REF_RE` doesn't match command/process substitution (`$(...)` or backticks) used directly as echo's argument, so `echo "$(cat file.json)" > out.txt` is not flagged even though the same escape-expansion risk applies to whatever string is echoed, not just bare variables — verified: `_echo_hazard_in_line('echo "$(cat file.json)" > out.txt')` returns `False`. Not hit in the live tree today, but nothing stops a future contributor from writing this exact pattern. Extend `_VAR_REF_RE` (or add a second pattern) to also match `$(` and backtick command-substitution openers, with a regression test.

scripts/validate.py
```diff
@@ -1187,6 +1187,139 @@ def check_browser_tool_gate(cache=None):
...
+                    continue  # unreadable — already reported by another check
+            else:
+                try:
+                    text = md.read_text(encoding="utf-8")
+                except OSError:
+                    continue
+            for block_match in _BASH_BLOCK_RE.finditer(text):
+                block = block_match.group(1)
+                block_start_line = text.count('\n', 0, block_match.start(1)) + 1
+                for offset, line in enumerate(block.splitlines()):
```
lugassawan: **Severity: Medium**

Per-line scanning misses hazards split across a backslash line-continuation: `echo "$JSON" \` on one line followed by `  > /tmp/payload.json` on the next is a single logical command with a real hazard, but each physical line is evaluated independently, so it evades detection entirely (verified: zero failures reported against a two-line continuation). Either flag any hazard-candidate line ending in an unquoted trailing `\` as "unable to verify — rewrite without continuation", or join continuation lines before scanning; add a regression test either way.

lugassawan: **Severity: Low**

Literal "bad-example" text inside a heredoc within a fenced ```bash block would be falsely flagged as a real hazard, since the scanner operates on raw lines regardless of heredoc context (comments are already safely exempted via the command-position regex, but heredoc body lines are not). No current occurrence in the live tree — a narrow, currently-theoretical gap inherent to a regex-based (non-parsing) lint. Consider noting the limitation in the function's docstring.

lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 5 inline comment(s) and 0 PR-level note(s), deduped 0.

scripts/validate.py
lugassawan: Addressed in cbd2236e8d7b83aeae6b3e53e2d1f9593f75c791: peeked for '>' after '&' before treating it as a separator, so '&>'/'&>>' now redirect-scans like a bare '>' (including the /dev/null exemption); added a regression test.

scripts/validate.py
lugassawan: Addressed in cbd2236e8d7b83aeae6b3e53e2d1f9593f75c791: added check_no_printf_var_format, a companion structural check flagging printf "$VAR"/"${VAR}"/$VAR (and the -v NAME form) used as a bare format string, with its own test class.

scripts/validate.py
lugassawan: Addressed in cbd2236e8d7b83aeae6b3e53e2d1f9593f75c791: extended _VAR_REF_RE to also match $( and a backtick, so command/process substitution used directly as echo's argument now flags too; added regression tests for both forms.

scripts/validate.py
lugassawan: Addressed in cbd2236e8d7b83aeae6b3e53e2d1f9593f75c791: added _join_bash_continuations to join a backslash-continued physical line with what follows before scanning, so a hazard split across a continuation is now caught as one logical command; added a regression test (plus one confirming an escaped trailing backslash is correctly NOT treated as a continuation).

lugassawan: Agreed this is a real (if narrow) limitation of a regex-based, non-parsing lint — a worked bad-example placed inside a heredoc body (rather than behind a '#' comment, which the command-position regex already exempts) would still be misread as a hazard. Noted explicitly in check_no_echo_var_hazard's docstring in cbd2236e8d7b83aeae6b3e53e2d1f9593f75c791 rather than fixed, since heredoc-body detection would meaningfully complicate the scanner for a currently-theoretical case with zero occurrences in the live tree. Happy to revisit if it ever actually bites.

scripts/validate.py
```diff
@@ -1187,6 +1187,254 @@ def check_browser_tool_gate(cache=None):
...
+    look like two independent physical lines (PR #564 review follow-up).
+    A trailing '\\\\' (escaped backslash, i.e. a literal backslash char) is
+    NOT a continuation.
+    """
+    i, n = 0, len(block_lines)
+    while i < n:
+        start = i
+        parts = [block_lines[i]]
+        while (
+            parts[-1].endswith('\\')
```
lugassawan: **Severity: Low**

`_join_bash_continuations`'s escaped-backslash guard (`endswith('\')` and not `endswith('\\')`) only distinguishes 1 vs 2 trailing backslashes. For a line ending in an *odd* count ≥3 (e.g. 3 or 5 trailing `\`), bash still treats it as a real line-continuation (the backslashes pair up as literal chars except the last one, which escapes the newline), but this code fails to join it, since `endswith('\\')` is true whenever the last two characters happen to be backslashes, regardless of total count. Confirmed empirically: joins correctly for 1/2/4/6 trailing backslashes but incorrectly fails to join for 3 and 5 — a hazard split across such a line would silently evade the check, the same bug class this fix commit just closed for count=1. Use backslash-count parity instead of a fixed suffix check, e.g. `trailing = len(line) - len(line.rstrip('\\')); is_continuation = trailing % 2 == 1`. Real-world likelihood is low (requires 3+ literal trailing backslashes right before a continuation), hence Low rather than Medium.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment(s) and 0 PR-level note(s), deduped 0.

scripts/validate.py
lugassawan: Addressed in 0bd1bf4d143662ab720473e3b7d4bdbcf33af2a0: switched the continuation guard to backslash-count parity (odd trailing count = real continuation) instead of a fixed 1-vs-2 suffix check, so 3+ trailing backslashes now join correctly too; added a regression test for the 3-backslash case.

PR #570, opened 2026-07-26
[chore] Migrate runtime/ callsites to bin/ wrappers (#560 B1)

lugassawan:
## Summary
- Add 9 `bin/swe-workbench-*` wrapper commands (one per `runtime/*.sh|py` script), each resolving its sibling `runtime/` script via `dirname "${BASH_SOURCE[0]}"` — never `$CLAUDE_PLUGIN_ROOT` — now that `<plugin>/bin` is on the Bash tool's PATH while the plugin is enabled.
- Migrate ~37 callsites across `skills/`, `commands/`, `agents/`, and `docs/` from the old `_RT="${CLAUDE_PLUGIN_ROOT:-$(git rev-parse --show-toplevel)}"` + `[ -f ... ]` guard + `bash "$_RT/runtime/<name>.sh"` pattern to a `command -v swe-workbench-<name>` preflight + bare `swe-workbench-<name>` invocation, with an actionable failure message ("reinstall or update the swe-workbench plugin") replacing the old dead-end "set CLAUDE_PLUGIN_ROOT and retry".
- Add `check_bin_wrappers()` to `scripts/validate.py` (prefix, exec-bit, `#!/usr/bin/env <interp>` shebang) and extend CI's shellcheck `additional_files` to cover the extension-less `bin/` wrappers.
- Rewrite ~16 test files' assertions from the old path/pattern checks to the new bare-command idiom; add `tests/test_bin_wrappers.py` (existence, exec bit, shebang, prefix, dirname-resolution, `bash -n`, and live-execution/stdin-passthrough checks).
- Add `docs/plugin-platform-decisions.md` recording three rulings from this migration (`CLAUDE_PLUGIN_DATA` considered-not-adopted, no closed-world plugin-schema validator in CI, the `bin/` dev-loop caveat) plus a `docs/README.md` index entry.

This is **B1 (expand) only** — `runtime/` and `hooks/inject_plugin_root.sh` remain in place and fully functional, so any callsite this migration missed keeps working exactly as before. B2 (deleting `runtime/`, the injector, and its `hooks.json` entry) is intentionally deferred to a follow-up issue, blocked on #557 and gated on a full release bake cycle for this PR.

## Test Plan
- [x] Examples in the diff were exercised manually (`bash bin/swe-workbench-doctor` from repo root and from `/tmp`; `git diff | bin/swe-workbench-comment-scan` vs `git diff | python3 runtime/comment-scan.py` byte-for-byte match)
- [x] `command -v` preflight fires correctly with the wrapper absent from PATH (non-zero exit, actionable message, verified via `env -i PATH=...`)
- [x] `runtime/` diff limited to README + 2 comment/usage-string updates — no deletions; injector and `hooks.json` untouched
- [x] `chmod -x` a wrapper → `validate.sh` fails; restored → passes again (new check has teeth)

- [x] `shellcheck --severity=warning $(git ls-files '*.sh') bin/*` — 0 issues
- [x] `pytest tests/ -v` — 2162/2162 pass
- [x] `lychee --config .lychee.toml` on the new doc's links — 0 errors

Closes #560


scripts/validate.py
```diff
@@ -908,6 +909,37 @@ def check_hook_scripts():
...
+
+
+def check_bin_wrappers():
+    bin_dir = ROOT / "bin"
+    if not bin_dir.is_dir():
+        return
+    for wrapper in sorted(bin_dir.iterdir()):
+        if not wrapper.is_file():
+            continue
+        rel = wrapper.relative_to(ROOT)
```
lugassawan, line 928: Low | scripts/validate.py:927 | `check_bin_wrappers()` validates each `bin/` wrapper (prefix, exec bit, shebang) but there's no reverse check that every `runtime/*.sh`/`.py` script has a corresponding `bin/` wrapper — completeness currently relies solely on the hardcoded `WRAPPERS` dict in `tests/test_bin_wrappers.py::test_bin_dir_has_exactly_the_expected_wrappers`. | If a future runtime script is added without updating that test dict, nothing in `scripts/validate.py`'s CI gate would catch the missing wrapper — the gap only surfaces if someone remembers to update the test alongside the new script. | Optional: add a clause in `check_bin_wrappers()` that also walks `runtime/*.sh`/`*.py` and asserts a same-named `bin/swe-workbench-<name>` wrapper exists, so the pairing is enforced by CI rather than convention.

lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment(s) and 1 PR-level note(s), deduped 0.

scripts/validate.py
lugassawan: Addressed in 073e1d9: added a reverse-direction check in check_bin_wrappers() so every runtime/*.sh|py script must have a matching bin/swe-workbench-<name> wrapper, enforced by CI rather than relying on the test file's hardcoded WRAPPERS dict alone.

PR #572, opened 2026-07-26
[feat] Add diff-line-lookup post-diff line resolver

lugassawan:
## Summary

- Add `runtime/diff-line-lookup.sh` — a POSIX-awk state machine that resolves a literal code snippet to its exact post-diff `path:line`, replacing the hand-counted offset arithmetic from a hunk header (`@@ -a,b +c,d @@`) agents were previously doing when assembling review comments.
- CLI: `swe-workbench-diff-line-lookup <path> <pattern> [--range=<rev-range> | --staged | --stdin]`. Refuses to guess: exits 2 (with every candidate on stderr) when a pattern matches more than one added line, and exits 1 when the pattern is only on a context or removed line (never on an added one).
- Wire the helper into both surfaces that motivated issue #551: `agents/reviewer.md` Step 5 now derives `File:Line` via the helper instead of hand-counting, and `skills/workflow-pr-review-post/SKILL.md` Step 2's pre-validate (the check that guards the atomic Reviews API batch from a single bad `(path, line)` 422'ing the whole review) now confirms each survivor via the same helper, reused by Step 4's retry logic.
- Two intentional deviations from the ticket, resolved via a `--grill` interrogation session (documented in code/commit history, not just decided silently):
  - `--sha=<sha>` → `--range=<rev-range>` — one semantic instead of two ("`--sha=HEAD`" and "default: HEAD" meaning different diffs under one word), and `--range` additionally expresses `main...HEAD`, the local equivalent of a PR diff.
  - Multi-line patterns are rejected (exit 64) instead of the ticket's "returns first match" — consistent with the chosen fail-loud-on-ambiguity posture rather than silently picking a candidate.
- Addressed two-reviewer feedback from Phase 4: reject an empty pattern (exit 64) instead of silently matching every added line (`index(s, "")` is always `> 0` in awk), map a failing `git diff` (e.g. an invalid `--range`) to the documented exit-64 contract instead of leaking git's own exit code, force `--no-color` on the internal `git diff` calls, and cache the PR diff fetch in `line_is_in_diff` (`$PR_DIFF`) so posting N findings costs one `gh pr diff` call instead of N.

## Test Plan

- [x] `pytest tests/ -v` — 2191/2191 pass
- [x] `shellcheck -S warning runtime/*.sh bin/swe-workbench-*` — 0 issues

- [x] New behavior demonstrated end-to-end against this PR's own real diff, not just synthetic fixtures — `bin/swe-workbench-diff-line-lookup runtime/README.md 'diff-line-lookup.sh' --range=main...HEAD` and `bin/swe-workbench-diff-line-lookup runtime/diff-line-lookup.sh 'DLL_PATTERN' --range=main...HEAD`; one result hand-checked against the raw hunk header arithmetic to confirm the tool's own math is trustworthy. The ambiguity path was also demonstrated live: `DLL_PATTERN` appears in two added lines in the script itself, and the tool correctly refused to guess between them.

Closes #551


skills/workflow-pr-review-post/SKILL.md
```diff
@@ -92,7 +92,7 @@ gh api graphql -F subjectId="$THREAD_HEAD_ID" -f query='
...
...e_is_in_diff`, defined once here (fetching the PR diff once into `$PR_DIFF` and reusing it across every survivor and Step 4's retry, never re-fetching per row) and reused by Step 4's retry: `command -v swe-workbench-diff-line-lookup >/dev/null 2>&1 || { echo "swe-workbench runtime commands not on PATH — reinstall or update the swe-workbench plugin." >&2; exit 1; }; PR_DIFF="${PR_DIFF:-$(gh pr diff "$PR")}"; line_is_in_diff() { local repo_path="$1" line="$2" head_sha="$3" content resolved; content=$(git show "${head_sha}:${repo_path}" | sed -n "${line}p") || return 1; [ -n "$content" ] || return 1; resolved=$(printf '%s\n' "$PR_DIFF" | swe-workbench-diff-line-lookup "$repo_path" "$content" --stdin 2>/dev/null) || return 1; [ "$resolved" = "${repo_path}:${line}" ]; }`. A line that no longer resolves (helper exits non-zero — not found, or ambiguous because the extracted content is too short to be distinctive) is treated the same as out-of-diff: **In-diff** (`line_is_in_diff` returns 0) → assemble into `comments[]` below. **Out-of-diff / unresolved** → re-anchor the finding `pr-level` and fold it into the PR-level batch below — this is the contract's existing home for out-of-diff rows.
```
lugassawan: **High** | `skills/workflow-pr-review-post/SKILL.md:95` | `line_is_in_diff`'s `sed -n "${line}p"` interpolates the finding's `line` value into a sed script unvalidated

The newly-defined `line_is_in_diff` function runs `git show "${head_sha}:${repo_path}" | sed -n "${line}p"` where `$line` traces back to a reviewing LLM agent's own `Line` field for a finding — content ultimately derived from analyzing the PR under review, which is exactly the kind of externally-controlled/adversarial input this tool's own threat model exists to handle (this is a PR-review bot; the PR diff/description is attacker-reachable text that can attempt to prompt-inject the reviewing agent). Nothing here asserts `$line` is a plain positive integer before it's spliced into the sed address. On a GNU sed runtime, an attacker-crafted value containing the `e` (execute) command — reachable if a prompt-injection payload in the PR content coerces the reviewing agent into emitting a non-numeric `line` for one of its own findings — runs an arbitrary shell command in the reviewing agent's execution environment, which typically holds live `git`/`gh` credentials. Even setting aside the adversarial framing, an ordinary bug upstream (a malformed or non-numeric `line`) turns into an opaque sed failure instead of a clean, predictable `return 1`.

**Suggested fix:** Validate `$line` is `^[0-9]+$` (e.g. `case "$line" in ''|*[!0-9]*) return 1 ;; esac`) at the top of `line_is_in_diff`, before it's ever interpolated into the `sed` invocation — this is a one-line, no-downside hardening step regardless of how likely the injection chain is judged to be.

runtime/diff-line-lookup.sh
```diff
@@ -0,0 +1,149 @@
...
+  next
+}
+
+!in_hunk && /^\+\+\+ / {
+  hdr = $0
+  sub(/^\+\+\+ /, "", hdr)
+  if (hdr == "/dev/null") {
+    in_target = 0
+  } else {
+    sub(/^b\//, "", hdr)
```
lugassawan, line 88: **Medium** | `runtime/diff-line-lookup.sh:88-89` | Path-header matching assumes the default `a/`/`b/` diff prefix, silently breaking under `diff.mnemonicPrefix`

The awk script strips a hardcoded `^b/` prefix (`sub(/^b\//, "", hdr)`) to compare the `+++` header against `$TARGET_PATH`. Reproduced: with `git config diff.mnemonicPrefix true` set, the header becomes `+++ w/f.txt` instead of `+++ b/f.txt`, the `sub()` no-ops, `hdr` stays `w/f.txt`, the equality check against `f.txt` fails, and the tool reports "pattern not found in diff" for a line that is unambiguously present and added. This degrades silently (exit 1, generic "not found" message) rather than erroring loudly, so a caller (e.g. `workflow-pr-review-post`'s `line_is_in_diff`) would treat a genuinely in-diff line as out-of-diff and mis-anchor otherwise-correct review comments to the PR-level batch instead of inline, with no diagnostic pointing at the real cause.

**Suggested fix:** Force a canonical prefix on the script's own internal `git diff` invocations (head/staged/range modes) via `--src-prefix=a/ --dst-prefix=b/`, independent of the caller's `diff.mnemonicPrefix`/`core.quotePath` settings, so the hardcoded `^b/` strip is always valid. (The `--stdin` mode is lower risk since `gh pr diff` output uses standard prefixes regardless of local git config.)

lugassawan (commented):

skills/workflow-pr-review-post/SKILL.md
lugassawan: Addressed in d8901cc72be66e650ea306a83eb88850e8a887b8: line_is_in_diff now validates $line matches ^[0-9]+$ (reject empty/non-numeric with a plain return 1) before it's ever interpolated into sed -n "${line}p".

runtime/diff-line-lookup.sh
lugassawan: Addressed in d8901cc72be66e650ea306a83eb88850e8a887b8: the script's own git diff invocations now force --src-prefix=a/ --dst-prefix=b/, so the hardcoded ^b/ header strip stays valid regardless of the caller's diff.mnemonicPrefix setting. Verified against a real repo with diff.mnemonicPrefix=true set (raw header showed i/ and w/, helper still resolved correctly).

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comment(s) and 1 PR-level note(s), deduped 0.

PR #576, opened 2026-07-27
[fix] Quote CLAUDE_PLUGIN_ROOT in all hook command strings

lugassawan:
## Summary

- Every `hooks/hooks.json` command string was a bare, unquoted `$CLAUDE_PLUGIN_ROOT/hooks/<script>`. When the plugin root path contains a space and the login shell is bash/sh (not zsh), the unquoted expansion word-splits before the shell resolves a command, and the hook silently never executes — for `bash_guard.sh` and `secret_guard.py` this is a silent loss of a security control with no error surfaced.
- Rewrites all 10 command strings to an explicit-interpreter, quoted form: `bash "${CLAUDE_PLUGIN_ROOT}"/hooks/<script>.sh` / `python3 "${CLAUDE_PLUGIN_ROOT}"/hooks/secret_guard.py`. Entry order, matchers, and events are unchanged.
- Adds `tests/test_hooks_json_wiring.py`: a runtime regression guard that runs every hooks.json command string via `bash -c` (pinned — never `$SHELL`/`sh`, since zsh doesn't word-split and would hide the bug) against a spaced `CLAUDE_PLUGIN_ROOT`, plus two hardcoded literal RED/GREEN tests reproducing the exact `exit 127` word-split failure from the issue and its quoted-form fix.
- Extends `scripts/validate.py::check_hooks_json()` with a closed-form command-shape regex and a check that no hook entry carries an `if` key, and adds a new `check_hook_script_permissions()` asserting every `hooks/*.sh`/`*.py` is executable (checks the exec bit, not an exact mode, so a `umask 002` checkout doesn't false-positive).
- Adds `## 4. Hook `if` conditions — considered, not adopted` to `docs/plugin-platform-decisions.md`, documenting a scope correction from the issue: analysis against the platform schema shows `if` buys nothing for *any* of the 10 hooks (not just the 5 "advisory" ones the issue's acceptance criteria called out) — see the PR comment on #557 for the full per-hook table.

## Test Plan

- [x] `bash scripts/validate.sh && pytest tests/ -v` — 2292 passed, 1 skipped (pre-existing)
- [x] `shellcheck -S warning` over all `.sh` files and `bin/swe-workbench-*` — 0 issues

- [x] Reproduced the original bug directly: `bash -c '$CLAUDE_PLUGIN_ROOT/hooks/bash_guard.sh'` against a spaced root exits 127 (word-split); confirmed the fixed form resolves correctly
- [x] Proved RED before GREEN: reverted `hooks/hooks.json` to its pre-fix content and confirmed `tests/test_hooks_json_wiring.py` fails with a genuine subprocess `exit 127` (not a Python-side error) across all 8 hook scripts; restored the fix and confirmed GREEN
- [x] Two independent review passes (plan-alignment + diff-level correctness/security/design) completed; both findings (RED-proof mechanism, exec-bit exact-mode fragility under `umask 002`) fixed and re-verified

Closes #557

### DIFF
--- a/docs/plugin-platform-decisions.md
+++ b/docs/plugin-platform-decisions.md
@@ -53,3 +53,33 @@ two consequences:
   prefix requirement stands regardless of this ordering — PATH placement is not something this
   repo controls or should rely on — but it's worth recording that the shadowing direction is the
   opposite of what might be assumed.
+
+## 4. Hook `if` conditions — considered, not adopted
+
+`.claude-plugin/schemas/plugin.schema.json` defines the command-hook object as `{type, command,
+if, shell, timeout, statusMessage, once, async, asyncRewake}`. `if` is documented as *"Permission
+rule syntax to filter when this hook runs (e.g. `Bash(git *)`). Only runs if the tool call matches
+the pattern."* — it filters a tool-call payload against a pattern; it is not a general-purpose
+guard.
+
+Applied to every entry in `hooks/hooks.json` (issue #557), no `if` buys anything:
+
+| Hook | Event | Why `if` buys nothing |
+|---|---|---|
+| `bash_guard.sh` | `PreToolUse` / `Bash` | `Bash(*)` is exactly redundant with `matcher` |
+| `inject_plugin_root.sh` | `PreToolUse` / `Bash` | Same |
+| `skill_usage_record.sh` | `PreToolUse` / `Skill` | `Skill(*)` is exactly redundant with `matcher` |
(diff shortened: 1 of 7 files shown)


lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comment(s) and 1 PR-level note(s), deduped 0.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comment(s) and 0 PR-level note(s), deduped 0.

PR #580, opened 2026-08-02
[feat] Package workflow-pr-review-post Steps 1-4 as bin/swe-workbench-pr-review-submit (#550)

lugassawan:
## Summary
- Packages `skills/workflow-pr-review-post/SKILL.md`'s Steps 1-4 (GraphQL thread fetch, Jaccard dedup, diff-line pre-validate, atomic Reviews-API submit with 422 retry and model-A fallback) as `bin/swe-workbench-pr-review-submit` — a unit-testable Python script instead of ~150 lines of hand-derived bash+jq prose.
- SKILL.md shrinks from 300 to ~120 lines: only the input contract, one script invocation, Step 5's AskUserQuestion CTA, Failure modes, and Common mistakes remain. Step 6 (`/tmp` threads-cache reap) is deleted — state now lives in memory for the run.
- Deliberate upgrade: GraphQL pagination is now actually implemented (previously "document, don't implement" in the old prose).
- Retargets 8 test files whose assertions pinned exact bash substrings that no longer exist to behavioral tests driving the new script through a stubbed `gh` on PATH (2 more than the plan's original 6 — `test_skill_repo_ambiguity.py` and `test_workflow_pr_review_followup_skill.py`'s shared prose also needed the `headRefOid`/canonical-remark-link text restored to the rewritten SKILL.md).
- Addressed two-reviewer feedback: added branch-level comments to the 422/5xx/fallback state machine, extended the byline/remark input-contract check to finding bodies (not just `--byline`), added a sibling-script preflight (mirrors the `command -v` guard every other `bin/` script carries), and narrowed a docstring claim about the read-your-write confirmation to note it requires known identity.

## Test Plan
- [x] `pytest tests/ -v` — 2317 passed, 1 skipped
- [x] `git diff | bin/swe-workbench-comment-scan` — 0 must-triage findings
- [x] `bin/swe-workbench-pr-review-submit --help` and an invalid `--findings-json` abort case, run by hand

- [x] New behavior (the script's dedup/pre-validate/submit state machine) verified via 70+ behavioral + unit tests, not just happy-path
- [ ] Live end-to-end check against a real PR's Reviews API (only exercisable with a throwaway PR + real network — not run in this session; flagged as a follow-up, see plan's Risks section)

Closes #550


lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comment(s) and 0 PR-level note(s), deduped 0.

bin/swe-workbench-pr-review-submit
```diff
@@ -0,0 +1,626 @@
...
+    owner, repo = args.repo.split("/", 1)
+    inline_findings, pr_level_findings = partition_findings(findings)
+    script_dir = _script_dir()
+
+    try:
+        _check_siblings(script_dir)
+        threads = fetch_threads(owner, repo, args.pr)
+        survivors, deduped = dedup_and_react(threads, inline_findings, args.current_user)
+
+        pr_diff = _gh(["pr", "diff", args.pr]).stdout
```
lugassawan: `pr_diff = _gh(["pr", "diff", args.pr]).stdout` discards `returncode` — a failed/timed-out `gh pr diff` call silently yields an empty string that flows straight into `validate_lines`. Every survivor then fails `_line_is_in_diff` (verified: `swe-workbench-diff-line-lookup` exits 1 on an empty `--stdin` diff), so **all** inline findings get silently demoted into one generic pr-level comment — with no error, no warning on stderr, and `SUBMITTED=true` still reported. This directly undercuts the module's own stated design goal (per-line inline comments over "a wall of text"), and it's exactly the kind of transient network failure the module docstring says the sibling gh-timeout wrapping exists to guard against — yet this one call site isn't checked. No test exercises a failing `gh pr diff` response.

Suggested fix: check `proc.returncode` after the `gh pr diff` call; on failure, either raise `PostingError` (fail loud, matching the rest of the module's posture) or explicitly log a `[warn]` and proceed with all-inline-demoted as a documented degraded mode rather than a silent one.

bin/swe-workbench-pr-review-submit
```diff
@@ -0,0 +1,626 @@
...
+    if n > 0:
+        # Atomic path: one POST carrying every candidate comment plus the review event/body.
+        proc = _post_review(owner, repo, pr, head_sha, event, summary_fn(n, posted_pr_level), comments)
+        if proc.returncode == 0:
+            posted_inline, submitted, review_url = n, True, _extract_review_url(proc.stdout)
+        elif "422" in (proc.stdout + proc.stderr):
+            # Confirmed 422 (stale commit_id, or a line pre-validate missed): re-fetch HEAD,
+            # genuinely re-validate every comment against it, demote anything now out-of-diff
+            # into the pr-level batch, then retry the atomic POST exactly once.
+            new_head_data, _p = _gh_json(["pr", "view", str(pr), "--json", "headRefOid"])
```
lugassawan, line 424: On a confirmed 422, the retry re-fetches `head_sha` via `gh pr view --json headRefOid` but re-validates every comment's line against the same stale `pr_diff` string captured before the failure — it is never re-fetched. `_line_is_in_diff` pulls file content from the new `head_sha` (via `git show`) but pattern-matches it against the old diff text. For the exact scenario this retry exists to handle (the diff moved between pre-validate and submit), the re-validation checks new content against a diff that may no longer reflect reality, producing unreliable in-diff/demote decisions.

Suggested fix: re-fetch `pr_diff` alongside `head_sha` in the 422 branch before re-validating, or explicitly document why reusing the stale diff is acceptable if intentional.

bin/swe-workbench-pr-review-submit
```diff
@@ -0,0 +1,626 @@
...
+def validate_payload(args) -> tuple[list[dict] | None, tuple[str, str] | None]:
+    """Returns (findings, error) — error is a (field, problem) tuple, or None on success."""
+    if not re.match(r"^[^/\s]+/[^/\s]+$", args.repo or ""):
+        return None, ("--repo", "must be in the form owner/repo")
+    if not re.match(r"^[1-9][0-9]*$", args.pr or ""):
+        return None, ("--pr", "must match [1-9][0-9]*")
+    if not re.match(r"^[0-9a-fA-F]{40}$", args.head_sha or ""):
+        return None, ("--head-sha", "must be a 40-character git SHA")
+    if not args.base:
+        return None, ("--base", "must be non-empty")
```
lugassawan, line 533: `--base` is a required CLI argument, validated as non-empty, but never read anywhere else in the script. Minor dead-parameter surface, carried forward from the pre-existing bash contract, not a regression introduced by this PR.

Suggested fix: either drop `--base` from the contract (updating SKILL.md's input table and all callers) or note in the module docstring why it's required-but-currently-unused.

bin/swe-workbench-pr-review-submit
lugassawan: Addressed in ee187e5: gh pr diff's exit status is now checked (_fetch_pr_diff raises PostingError on failure instead of silently flowing an empty diff into validate_lines).

bin/swe-workbench-pr-review-submit
lugassawan: Addressed in ee187e5: the 422 retry now re-fetches the PR diff alongside HEAD before re-validating (best-effort — falls back to the stale diff on a refetch failure rather than aborting an in-flight submission).

bin/swe-workbench-pr-review-submit
lugassawan: Leaving this as-is: --base is carried forward from the pre-existing bash contract (the old SKILL.md prose also declared BASE non-empty without using it, per Step 1's original preflight). It's reserved input, not a regression introduced by this PR — happy to revisit dropping it from the contract in a follow-up if it's truly dead weight.

bin/swe-workbench-pr-review-submit
```diff
@@ -0,0 +1,652 @@
...
+    """Loud, actionable failure — an empty diff from a failed/timed-out call would otherwise
+    flow silently into validate_lines and demote every inline finding to one pr-level comment,
+    with no error and no stderr warning."""
+    proc = _gh(["pr", "diff", str(pr)])
+    if proc.returncode != 0:
+        raise PostingError(f"failed to fetch PR diff: {proc.stderr.strip() or proc.stdout.strip()}")
+    return proc.stdout
+
+
+def _refetch_pr_diff(pr: str, stale_pr_diff: str) -> str:
```
lugassawan, line 318: `_refetch_pr_diff`'s failure/fallback branch (refetch fails during a 422 retry, warns, and falls back to the stale diff) has zero test coverage — the two updated 422-retry tests only exercise the "refetch succeeds" path. This is a new error-handling branch introduced by this exact fix; a regression here (e.g., returning the wrong variable, or the print raising) would go undetected.

Suggested fix: add a case where the `gh pr diff` re-fetch call in the stub sequence returns a non-zero exit during the 422-retry path, and assert the retry still proceeds using the stale diff (and that the `[warn]` message appears on stderr).

bin/swe-workbench-pr-review-submit
```diff
@@ -0,0 +1,652 @@
...
+    """Best-effort refetch for the 422-retry path: re-validating against a stale diff after a
+    stale-commit_id 422 is the exact bug this exists to fix, but a refetch failure here is a
+    degraded retry, not a foundational one — fall back to the stale diff (no worse than before
+    this fix) rather than aborting an in-flight submission."""
+    proc = _gh(["pr", "diff", str(pr)])
+    if proc.returncode != 0:
+        print(f"[warn] PR diff re-fetch failed during 422 retry — reusing the pre-retry diff: {proc.stderr.strip()}", file=sys.stderr)
```
lugassawan: `_refetch_pr_diff`'s warn message uses only `proc.stderr.strip()`, while the sibling `_fetch_pr_diff` falls back to `proc.stdout.strip()` when stderr is empty. Minor inconsistency between two near-identical helpers added in the same commit; if the failure text ever lands on stdout, the warning would read with no detail.

Suggested fix: `proc.stderr.strip() or proc.stdout.strip()` to match `_fetch_pr_diff`'s pattern.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 2 inline comment(s) and 0 PR-level note(s), deduped 0.

bin/swe-workbench-pr-review-submit
lugassawan: Addressed in 06766b5: added test_422_retry_falls_back_to_stale_diff_when_refetch_fails, driving a failed pr-diff refetch during the 422 retry and asserting the fallback + [warn] behavior.

bin/swe-workbench-pr-review-submit
lugassawan: Addressed in 06766b5: _refetch_pr_diff now uses proc.stderr.strip() or proc.stdout.strip(), matching _fetch_pr_diff's pattern.

PR #581, opened 2026-08-02
[feat] Warn consumer repos that don't gitignore .claude/cache/

lugassawan:
## Summary

- Add a read-only "Repo hygiene" section to `bin/swe-workbench-doctor` that warns when `.claude/cache/` (the workflow-state + skill-usage sidecar directory) is not gitignored in the consumer repo, or when files under it are already tracked — using `git check-ignore --no-index` and `git ls-files`, never `.gitignore` text-parsing, never a write.
- Fix a stale "five runtime dependencies" claim in `commands/doctor.md` to "six" (the script has always probed `python3` too) and extend its regression test.
- Document the consumer-side gitignore requirement in `README.md` and `docs/workflow-state.md`, pointing at `/swe-workbench:doctor` as the backstop.

## Test Plan

- [x] `shellcheck --severity=warning bin/swe-workbench-doctor` — clean
- [x] `pytest tests/` — 2282 passed, 1 skipped (pre-existing unrelated skip)
- [x] Manual smoke test: ran the doctor script against this repo (clean, `.claude/cache/` already ignored) and against a fresh synthetic repo with no `.gitignore` (warns, exit 0 in both cases)

- [x] New behavior (the Repo hygiene section) is covered by 8 new tests in `tests/test_doctor_script.py`, including a regression guard for the tracked-and-ignored `--no-index` trap and an asymmetric-subdirectory-coverage case

Both required Phase 4 reviewers (`superpowers:requesting-code-review` and `swe-workbench:reviewer`) ran in parallel and returned no Critical/Important findings; two independently-converged Minor/Low findings (a misleading code comment, a missing regression test for asymmetric coverage) were fixed inline before this PR.

Closes #577


tests/test_doctor_script.py
```diff
@@ -151,3 +173,122 @@ def test_script_writes_no_files(tmp_path):
...
+    cache_dir = repo / ".claude" / "cache"
+    cache_dir.mkdir(parents=True)
+    (cache_dir / "tracked.json").write_text("{}\n")
+    subprocess.run(["git", "add", "-f", ".claude/cache/tracked.json"], cwd=repo, env=env, check=True)
+    subprocess.run(["git", "commit", "-q", "-m", "init"], cwd=repo, env=env, check=True)
+    (repo / ".gitignore").write_text(".claude/cache/\n")
+    result = _run_doctor(env, cwd=repo)
+    assert result.returncode == 0
+    assert "already tracked" in result.stdout.lower() or "tracked" in result.stdout.lower()
+    assert "1" in result.stdout
```
lugassawan: `assert "1" in result.stdout` doesn't validate what it claims to, in `test_hygiene_warns_when_already_tracked`. The substring "1" is already present elsewhere in this test's stdout regardless of the hygiene logic (e.g. `jq-1.7.1`, `Claude Code 1.2.3`, `Python 3.11.0` from the mocked dependency rows), so this assertion would still pass even if the script printed a wrong tracked-file count (e.g. "2 files already tracked"). Suggested fix: pin to the actual rendered line, e.g. `assert "1 file already tracked under .claude/cache/." in result.stdout`, or scope the check to `hygiene_section` the way the sibling tests already do.

bin/swe-workbench-doctor
```diff
@@ -83,4 +83,59 @@ else
...
 fi
 
+# ── Repo hygiene: warn if .claude/cache/ (workflow-state + skill-usage sidecars)
+#    is committable in this repo. Outside a git repo, print nothing and skip.
+root=$(git rev-parse --show-toplevel 2>/dev/null) || true
+
+if [[ -n "${root}" ]]; then
+  WARN=0
+
+  # Check A — is the pattern present? One check-ignore call per representative
```
lugassawan: The 'Check A' inline comment block runs 4 lines, exceeding the soft 2-line cap for `#` comments. The content is genuine non-obvious WHY (why 3 separate probes instead of one top-level check), so this is a minor style note rather than a real problem. Suggested fix (optional): trim to ~2 lines, e.g. 'Check A — pattern present? One `check-ignore` call per path (no batching: `-q` takes one pathname). Nested probes catch asymmetric per-subdirectory coverage.' — or leave as-is if the fuller rationale is judged worth the extra lines.

lugassawan (commented):
_Reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 2 inline comment(s) and 1 PR-level note(s), deduped 0.

tests/test_doctor_script.py
lugassawan: Addressed in 26726d37ee2ef40b2b7e09469061d7da2f26e426: pinned the assertion to the exact rendered line ("1 file already tracked under .claude/cache/.") instead of a bare "1" substring check.

bin/swe-workbench-doctor
lugassawan: Addressed in 26726d37ee2ef40b2b7e09469061d7da2f26e426: trimmed the Check A comment to the file's 2-line convention.

lugassawan (commented):
_Re-reviewed by `reviewer`_ ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comment(s) and 0 PR-level note(s), deduped 0.

PR #583, opened 2026-08-02
[feat] LSP code-intelligence literacy for orchestrator skills and depth-first agents

lugassawan:
## Summary
- Adds LSP code-intelligence literacy to four depth-first agents (`reviewer`, `auditor`, `debugger`, `refactorer`) via a new shared doc `agents/shared/lsp.md` teaching the "Grep/Glob locates the anchor, `LSP` expands from it" two-step handoff (LSP cannot search by name — every op needs `filePath`/`line`/`character`), plus a mandatory one-attempt-then-fallback gate.
- Adds `check_lsp_tool_gate` to `scripts/validate.py`: any agent granting `LSP` in frontmatter `tools:` must reference `@./shared/lsp.md`, keyed strictly on the frontmatter token — never a body-text regex, since `agents/shared/principles.md` and `skills/principle-solid/SKILL.md` legitimately use "LSP" to mean Liskov Substitution Principle.
- Adds the same symbol-navigation guidance (inline, since skills have no `@./shared/` include mechanism) to the three orchestrator skills: `workflow-development` (net +1 line, staying at its 300-line `orchestrator: true` cap), `workflow-pr-review`, `workflow-codebase-audit`.
- Documents the optional, graceful-fallback nature of LSP servers in `docs/dependencies.md` (new section, contrasted explicitly with the existing hard-gated browser-automation servers).
- New `tests/test_lsp_literacy.py` pins all of the above against the real repo tree.

## Resolved open questions (for reviewers)
- **`background:` key on the four agents:** resolved as **none of the above** — no `background` key is added anywhere. A spike confirmed an agent with no `background` key keeps the `LSP` tool usable, so `debugger` stays parallel-eligible in `workflow-codebase-audit --depth=deep`'s fan-out with no special-casing needed.
- **Manual transcript-grep acceptance item:** superseded by real CI — `check_lsp_tool_gate` (structural, frontmatter-keyed) plus `tests/test_lsp_literacy.py` (real-tree pinning) together discharge this; no manual canary is needed.
- `code-impl` deliberately stays out of scope (background-eligible, no LSP) — `workflow-delegated-implementation`'s parallel worktree-isolated dispatch is worth more than LSP access for it.

## Test Plan
- [x] Changed skills/commands/agents load without errors — `python scripts/validate.py` passes clean, including the new `check_lsp_tool_gate`.
- [x] Examples in the diff were exercised manually — ran the plan's targeted negative check confirming zero false positives against `agents/shared/principles.md` / `skills/principle-solid/SKILL.md`'s pre-existing "LSP" (Liskov) usage.
- [x] `wc -l skills/workflow-development/SKILL.md` reads exactly 300 (at its `orchestrator: true` cap).

- [x] New behaviour verified: each of the four agents' `tools:` frontmatter now grants `LSP` and references `@./shared/lsp.md`; `agents/shared/lsp.md` names all nine LSP operations and carries the fallback gate sentence verbatim (real em dash, U+2014); the three orchestrator skills and `docs/dependencies.md` carry the same sentence inline.
- [x] Full suite: `pytest tests/ -v` — 2369 passed, 1 skipped (pre-existing, unrelated).

Closes #559

### DIFF
--- a/agents/auditor.md
+++ b/agents/auditor.md
@@ -2,7 +2,7 @@
 name: auditor
 description: Cold-start codebase audit specialist — readonly multi-domain sweep across security, performance, reliability, tooling, and testing. Surfaces ranked findings with root-cause reasoning chains and counter-evidence calibration. Invoke when you want a time-boxed audit of an unfamiliar codebase, not a single-domain depth-first pass.
 model: sonnet
-tools: Read, Grep, Glob, Bash, Skill
+tools: Read, Grep, Glob, Bash, Skill, LSP
 skills:
   - swe-workbench:principle-code-review
   - swe-workbench:principle-security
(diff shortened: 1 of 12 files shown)


lugassawan (commented):

PR #584, opened 2026-08-02
[chore] Route skill-local scripts through a bin/ dispatcher

lugassawan:
## Summary

- Add `bin/swe-workbench-skill-script <skill-name> <script-name> [args...]` — a dispatcher that self-locates the plugin root, rejects `/` and `..` in either argument, validates the target `skills/<skill>/scripts/<script>` exists, and `exec bash`s it through with stdout/exit-code passthrough.
- Rewrite the 10 call sites across `skills/workflow-cleanup-merged/SKILL.md` (6) and `skills/workflow-branch-sync/SKILL.md` (4) that previously derived a skill-local root via `_RT="$(cd "$(dirname "$(command -v swe-workbench-doctor)")/.." && pwd)"` — they now invoke the dispatcher directly. Re-point both skills' preflight guards to `command -v swe-workbench-skill-script` (the command actually invoked downstream).
- Retarget `tests/test_runtime_root_binding.py`: delete the two tests that pinned the old `_RT=` derivation in place (they would have passed vacuously forever once the pattern was gone — the exact failure mode a prior fix in this repo called out), add `test_no_path_resolution_in_skill_prose` and `test_skill_local_scripts_invoked_via_dispatcher`, both asserting a non-empty fixture list before looping.
- Add 6 runtime/e2e tests for the dispatcher itself in `tests/test_bin_scripts.py` (traversal rejection, exit-code passthrough, stdout passthrough, args-with-spaces, missing target, missing args) — the existing static checks (shebang/syntax/executability) didn't exercise any of this.
- Fix a prose-pinning test in `tests/test_branch_sync_skill.py` that asserted the literal old `$_SCRIPTS/detect-conflicts.sh` string.
- Docs: `bin/README.md` gets the new dispatcher row + an extended Reference pattern note; `docs/plugin-platform-decisions.md` gets a new ruling recording why the doctor-anchor derivation was replaced and why the `command -v` PATH guard itself can't be moved into a PATH-resolved helper.

## Test Plan

- [x] `bash -n bin/swe-workbench-skill-script` — clean
- [x] `shellcheck --severity=warning bin/swe-workbench-skill-script` — 0 issues
- [x] `pytest tests/ -q` — 2420 passed, 1 skipped (pre-existing, unrelated)
- [x] Manually invoked the dispatcher against real skill scripts (`resolve-rimba.sh`, `preflight-guard.sh`, `probe-worktree.sh`) and against traversal/missing-target/missing-args inputs
- [x] Post-change `grep -rn '_RT=' skills/` and `grep -rn 'dirname "$(command -v' skills/` both return empty

- [x] `pytest tests/ -q` passes in full (not just changed files) — a prior migration in this repo found prose-pinning tests outside the obviously-changed file, which is exactly what happened here (`test_branch_sync_skill.py`)

Closes #569

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -28,6 +28,7 @@ full implementation, invocable directly by its bare `swe-workbench-<name>` comma
 | `swe-workbench-pr-review-submit` | Posting mechanism for workflow-pr-review-post's Steps 1-4: fetch review threads (paginated), Jaccard dedup + 👍 reactions, diff-line pre-validate, pr-level batching, self-review/diff-scoping decision flip, atomic Reviews-API submit with a bounded 422 retry and a per-comment fallback. `--findings-json <path\|->` in; `printf %q`-quoted `KEY=VALUE` lines out |
 | `swe-workbench-reap-run-dir` | Safe `rm -rf` for a single run-scoped scratch dir allocated by `swe-workbench-new-run-dir` (depth-exactly-one, name-shape, ownership, and `.git`-absence checks) |
 | `swe-workbench-reply-and-resolve` | Post a PR review thread reply (REST) and optionally resolve it (GraphQL) |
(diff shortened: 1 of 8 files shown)


lugassawan (commented):

PR #587, opened 2026-08-02
[chore] Collapse PR-review skill pair, buy line-cap headroom

lugassawan:
## Summary

- **#565** — Fold `workflow-pr-review-followup` into `workflow-pr-review` using a shell-variable mode table (`MODE_TAG`/`STATE_SUFFIX`/`CALLER_TAG`/`BYLINE`) instead of maintaining a second ~175-line near-duplicate skill. Deletes `workflow-pr-review-followup/` entirely, updates every downstream reference/test/prose mention, reduces (not drops) the BM25 sibling-set exemption from 3 names to 2, and removes a dead trigger surface that advertised a read-only/digest mode that never existed.
- **#567** — Buy real headroom on the 3 `workflow-*` skills actually sitting near the `orchestrator: true` 300-line cap (`workflow-address-feedback`, `workflow-development`, `workflow-cleanup-merged`) by extracting conditional/mechanics content into companion `reference/*.md` files, per the house convention. Adds a non-fatal `scripts/validate.py` warning at 90% of a skill's line cap so future near-cap drift surfaces early without failing the build.
- Fixes stale "Step 2"/"Step 4" pointers left over from PR #580's collapse of `workflow-pr-review-post` into a single `## Post` section (found while investigating the fold).
- Corrects two stale line-count claims in `docs/extending.md` (`workflow-cleanup-merged` reported as 298 lines; actually 203 after extraction) and `docs/gh-api-field-flags.md` (referenced "Step 2"/"Step 4" of `workflow-pr-review-post`, which no longer exist).

**Standing warnings after this PR:** 4 base-cap skills remain within 10% of their 150-line cap — `principle-product-design` (144), `workflow-codebase-knowledge` (137), `language-bash` (137), `language-python` (136). These are pre-existing and out of scope for this PR; the new headroom check will keep surfacing them as a non-fatal reminder.

**Migration note:** any consumer allow-rule naming the deleted `workflow-pr-review-followup` skill goes inert (a permission prompt, not an error) — no deprecation shim was added, since a shim large enough to stay in the skill-trigger corpus would forfeit the entire routing fix this PR delivers.

## Test Plan

- [x] Examples in the diff were exercised manually (BM25 self-ranking re-measurement on the committed tree; direct 3-case exit-code reproduction of the fixed OPEN-state gate)
- [x] `bash scripts/validate.sh` — all checks pass, exit 0; exactly the 4 standing base-cap warnings above, no others
- [x] `pytest tests/ -v` — 2461 passed, 1 skipped (pre-existing, root-gated)
- [x] `wc -l` on the 3 extracted skills — all ≤ 270 (269 / 243 / 203)
- [x] `git grep -n "workflow-pr-review-followup"` — zero hits in the tracked tree

- [x] `scripts/validate.py`'s `check_skill_skill_refs` used as the completeness oracle for the deleted-skill reference sweep — passes clean
- [x] Full branch diff went through a final whole-branch review (in addition to per-task reviews) with one fix wave for 3 Important + 2 Minor findings, re-reviewed clean

Closes #565
Closes #567


lugassawan (commented):

PR #588, opened 2026-08-02
[chore] Namespace machine-actionable skill/agent refs (#586)

lugassawan:
## Summary

- Issue #586's premise didn't hold — bare refs in agent preload declarations are already a hard validator failure, and the true count (529 bare refs / 81 files) was mostly legitimate prose/catalog/README content that stays bare by design. The real, narrower defect: some `commands/*.md` and `skills/*/SKILL.md` files write machine-actionable dispatch instructions in bare form (e.g. "Delegate to the `senior-engineer` subagent") while the identical construct elsewhere uses the namespaced `swe-workbench:<id>` form, and no validator caught the drift.
- Adds `check_bare_actionable_refs()` to `scripts/validate.py` as a regression ratchet: Rule 1 requires every bare skill/agent id in `commands/*.md` (outside fenced code) to be namespaced; Rule 2 flags bare ids only on skill dispatch-cued lines in `skills/*/SKILL.md`, reusing the existing `_ACTION_RE`/`_POINTER_RE`/`_SLASH_CMD_RE` classifier `check_no_cycles` already uses. A `<!-- validate: prose-ref -->` marker exempts genuine non-dispatch mentions.
- Namespaces ~76 bare dispatch refs across 17 `commands/*.md` files and 11 `skills/*/SKILL.md` files (the code-review pass caught 4 more sites the mechanical sweep's classifier missed, since they used verbs like "Hand"/"Pass" outside the reused action-cue word list).
- Retargets 4 test files whose assertions pinned the bare form on now-rewritten lines, and adds a `TestCheckBareActionableRefs` suite (10 cases + a live-tree assertion) plus a CONTRIBUTING.md bullet documenting the new check.
- Out of scope by design: `agents/*.md` prose, `docs/catalog.md`, `README.md`, and command ids (`report-issue`, `capture`, `sync`) — a follow-up issue covers the optional prose-only pass.

## Test Plan

- [x] `python3 scripts/validate.py` — 0 issues; deliberately regressed one normalized line and stripped one exemption marker to confirm the new check fires and names the escape hatch, then restored both
- [x] `pytest tests/ -v` — 2472 passed, 1 skipped
- [x] Dependency-flow cycle simulation (`_build_dep_graph`) re-run before/after: 63 nodes/80 edges → 68 nodes/109 edges, 0 cycles both times
- [x] Both required reviewers (swe-workbench:reviewer subagent + plan-alignment reviewer) ran in parallel; all Important findings fixed, Minor findings addressed

- [x] Repo builds/validates cleanly after the change (`scripts/validate.py`, full `pytest`)

Closes #586

### DIFF
--- a/scripts/validate.py
+++ b/scripts/validate.py
@@ -565,6 +565,107 @@ def check_skill_skill_refs(cache=None):
                 )
 
 
+_BARE_ID_RE = re.compile(r'`([\w-]+)`')
+_PROSE_REF_EXEMPTION_MARKER = '<!-- validate: prose-ref -->'
+
+
+def _bare_actionable_id_set():
+    """Return the set of every resolvable skill and agent id, mirroring the
+    resolution index _build_dep_graph builds for the cycle checker."""
+    ids = set()
+    for p in (ROOT / "skills").glob("*/SKILL.md"):
+        ids.add(p.parent.name)
+    for p in (ROOT / "agents").glob("*.md"):
+        if p.parent.name == "shared":
+            continue
+        ids.add(p.stem)
+    return ids
+
+
+def check_bare_actionable_refs(cache=None):
+    """Machine-actionable dispatch references must use the namespaced
+    `swe-workbench:<id>` form (#586).
+
+    check_command_skill_refs / check_skill_skill_refs only resolution-check
+    the already-namespaced `` `swe-workbench:<id>` `` pattern, so a bare
+    dispatch id (e.g. "Delegate to the `senior-engineer` subagent") is never
+    validated and can silently drift from the namespaced form used for the
+    identical construct elsewhere. Two rules, both scoped to
+    machine-actionable dispatch — prose, catalog tables, and README
+    enumerations are untouched:
+
+    Rule 1 (commands/*.md): every bare skill/agent id outside a fenced code
+    block must be namespaced — command bodies are dispatch scripts, so no
+    action-cue heuristic is applied. A line may opt out with the
+    '<!-- validate: prose-ref -->' marker (a genuinely non-dispatch mention,
+    e.g. an example list of literal identifier strings).
+
+    Rule 2 (skills/*/SKILL.md): only lines the action-cued activation
+    classifier (_ACTION_RE / _POINTER_RE / _SLASH_CMD_RE, shared with
+    check_no_cycles's _build_dep_graph) would treat as a dispatch edge. A
+    skill naming its own id is exempt (self-reference), and the same
+    exemption marker opts out a line the classifier over-flags as a
+    dispatch cue when it is really prose.
+    """
+    if cache is None:
+        cache = _build_cache()
+    skills_cache = cache[1]
+
+    ids = _bare_actionable_id_set()
+    commands_dir = ROOT / "commands"
+    skills_dir = ROOT / "skills"
+
+    # Rule 1 — commands/*.md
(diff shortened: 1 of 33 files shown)


scripts/validate.py
```diff
@@ -565,6 +565,107 @@ def check_skill_skill_refs(cache=None):
                 )
 
 
+_BARE_ID_RE = re.compile(r'`([\w-]+)`')
```
lugassawan, line 568: `_BARE_ID_RE` only matches backtick-wrapped identifiers, so a future dispatch instruction written without backticks (e.g. "Delegate to the senior-engineer subagent") would silently bypass both Rule 1 and Rule 2. The check's guarantee is "no bare backtick-wrapped id" rather than "no bare id" — a stylistic drift from the codebase's current (consistent) backtick convention wouldn't be caught. Optional: note this scope limitation in the docstring/CONTRIBUTING.md entry; not worth widening the regex given the current 100% backtick convention.

scripts/validate.py
```diff
@@ -565,6 +565,107 @@ def check_skill_skill_refs(cache=None):
...
+    enumerations are untouched:
+
+    Rule 1 (commands/*.md): every bare skill/agent id outside a fenced code
+    block must be namespaced — command bodies are dispatch scripts, so no
+    action-cue heuristic is applied. A line may opt out with the
+    '<!-- validate: prose-ref -->' marker (a genuinely non-dispatch mention,
+    e.g. an example list of literal identifier strings).
+
+    Rule 2 (skills/*/SKILL.md): only lines the action-cued activation
+    classifier (_ACTION_RE / _POINTER_RE / _SLASH_CMD_RE, shared with
```
lugassawan: Docstring says Rule 2 shares its classifier "with check_no_cycles's `_build_dep_graph`," but Rule 2 additionally runs `_strip_fenced_code_blocks` on the text before applying `_ACTION_RE`/`_POINTER_RE`, while `_build_dep_graph`'s `_scan` does not strip fenced blocks first. A dispatch-cued line inside a fenced example block would produce a cycle-graph edge in `check_no_cycles` but would never be scanned by Rule 2's bare-ref check, despite the docstring's "shared" framing. Optional: either strip fenced blocks in `_build_dep_graph` too for full parity, or soften the docstring wording to "same classifier regexes" rather than implying identical line-selection behavior.

lugassawan (commented):

scripts/validate.py
lugassawan: This is an accepted scope limitation, not a gap to close: the check's guarantee tracks the codebase's existing convention that machine-actionable dispatch references are always backtick-wrapped — every sibling ref check in this file (check_command_skill_refs, check_skill_skill_refs, check_no_cycles's _CYCLE_MENTION_RE) already only matches the backtick-wrapped `swe-workbench:<id>` form for the same reason. Widening the regex to catch un-backticked mentions would risk false positives on ordinary prose that names an id without intending a dispatch. Leaving as-is; thread stays open for visibility.

scripts/validate.py
lugassawan: Addressed in 1432cc7b565e5abd7b71ff5c196275088c7b694a: softened the Rule 2 docstring to say the classifier is the same regexes (not identical line-selection), and called out explicitly that Rule 2 also strips fenced code blocks first while _build_dep_graph does not.

PR #593, opened 2026-08-06
[feat] Add /swe-workbench:converge command

lugassawan:
## Summary

- Adds `/swe-workbench:converge` — an orchestrator command that runs an unattended local review→fix→re-review convergence loop on the current branch (up to 4 review passes, `swe-workbench:reviewer` for review rounds and `swe-workbench:code-impl` for fix rounds), and pushes exactly once via `swe-workbench:workflow-commit-and-pr` only when the reviewer converges (no findings ≥Medium left).
- Findings never post to GitHub mid-loop — they live in a local scratch file under `$RUN_DIR`, reaped immediately after each round consumes it.
- Verifies findings before acting on them: mechanical anchor validation (file exists, in-diff, line resolves inside a hunk at the claimed line) runs before anything reaches the fixer, then the fixer itself must verify each claim against the actual code — `UNFOUNDED` (false premise) is tracked distinctly from `REJECTED` (true premise, declined to act), since a rising `UNFOUNDED` rate is the signal the reviewer itself is degrading.
- Ownership-gated: refuses to run (no scratch state allocated) unless the current `gh` user is the PR's author, since the loop commits and pushes autonomously.
- Hard stops that end the loop without pushing: a red test suite, a fix round that produces zero edits, oscillation (a previously-fixed finding reappears), more than 4 `UNVERIFIABLE` findings in one round, a footer/findings parse mismatch, or cap exhaustion (default 4 reviews, `--cap 2..6`). The loop invariant is enforced in the termination check itself: it always ends immediately after a review, never after an unreviewed fix.
- Reuses existing repo mechanisms rather than inventing new ones: the findings schema validated by `_finding_problem()` in `bin/swe-workbench-pr-review-submit`, `bin/swe-workbench-diff-line-lookup` for anchor validation, `bin/swe-workbench-new-run-dir`/`swe-workbench-reap-run-dir` for scratch-state lifecycle, and `workflow-pr-review`'s decision-footer parsing contract.
- Adds `tests/test_converge_command.py` (21 tests asserting specific bounds — cap=4, floor=Medium, Jaccard=0.4 — not just "a bound exists"), and registers the command in `docs/catalog.md`, `README.md`, and `tests/test_no_bare_command_mentions.py`.

## Test Plan

- [x] `bash scripts/validate.sh` — 0 issues (tmp-write gate, bare-refs ratchet, cycle check, echo/printf hazard scan all clean)

- [x] `python3 -m pytest tests/ -q` — full suite green (2494 passed, 1 skipped), including the 21 new `test_converge_command.py` assertions
- [x] Two independent parallel reviews (`superpowers:requesting-code-review` + `swe-workbench:reviewer`) run against the diff; both independently caught the same Critical bug (cap check firing after a fix instead of after a review) plus a missing out-of-diff/UNFOUNDED distinction and a `code-impl` output-contract mismatch — all three fixed and locked in with new regression tests
- [ ] End-to-end smoke run of `/swe-workbench:converge --cap 2` on a throwaway branch with deliberate defects (deferred — see follow-up note below)

Closes #592


commands/converge.md
```diff
@@ -0,0 +1,412 @@
...
+destructive-by-default hazard, not a style issue. This is a **stronger** gate than
+`swe-workbench:workflow-address-feedback`'s owner check, which only shapes prompts.
+
+```bash
+gh auth status >/dev/null 2>&1 || {
+  echo "converge: gh is not authenticated — run 'gh auth login' first." >&2
+  exit 1
+}
+CURRENT_USER=$(gh api /user -q .login)
+PR_AUTHOR=$(gh pr view --json author -q .author.login 2>/dev/null || true)
```
lugassawan: **High | Ownership-gate fail-open on transient `gh` errors**

`PR_AUTHOR=$(gh pr view --json author -q .author.login 2>/dev/null || true)` collapses "no PR exists for this branch" and "`gh` API/network failure" into the same empty-string outcome, and the doc explicitly treats empty `PR_AUTHOR` as "proceed, you own it by construction." A transient `gh` failure (rate limit, network blip, API hiccup) while checking a branch that *does* have an open PR authored by someone else silently bypasses the refuse gate — exactly the scenario this section exists to prevent. `bin/swe-workbench-preflight-pr` uses `set -euo pipefail` and explicit empty/null field checks rather than swallowing `gh` errors with `|| true`, so this departs from an established, safer pattern already in this repo.

**Why it matters:** a masked `gh` error lets the autonomous commit/fix/push loop run against a PR it doesn't own whenever the ownership check itself fails transiently — the one gate this command exists to enforce.

**Suggested fix:** distinguish "no PR for this branch" (a specific, expected non-zero exit / empty JSON) from any other `gh pr view` failure; on the latter, fail closed with a "could not verify PR ownership — try again" message rather than falling through to the empty-`PR_AUTHOR` branch.

commands/converge.md
```diff
@@ -0,0 +1,412 @@
...
+question in its returned line rather than asking. This command body — running in the orchestrator
+— owns the ask:
+
+- **One batched `AskUserQuestion` per round, never one per finding.** Up to 4 findings become the
+  4 questions of a single call. Each offers *Real — fix it* / *Not real — drop it* / *Real, don't
+  fix now*, with the finding body and the agent's investigation notes as context.
+- **More than 4 `UNVERIFIABLE` findings in one round → stop the loop** and jump to Phase 6
+  (too-many-unverifiable). At that density the review needs a human reading, not an
+  interrogation.
+- **Answers are cached in the ledger's `adjudicated[]`, keyed by fingerprint, and replayed into
```
lugassawan: **High | `fingerprint` used by `adjudicated[]` is never defined, and contradicts the document's own dedup rationale**

`adjudicated[]`, keyed by fingerprint (and the fixer's return line, `<fingerprint> | FIXED …`) relies on a term the document never defines — the `{severity, body, anchor, path, line}` findings schema (deliberately reused from `_finding_problem()`) has no `fingerprint` field, so it's unclear how the fixer derives one, or how it stays stable across rounds. The "Cross-round dedup" section argues *against* a line-sensitive identity for exactly this reason — "local fixes shift line numbers within a round, so a line-sensitive fingerprint would fail precisely when it matters most" — and adopts path+Jaccard instead for `retired[]`/`rejected[]`/`unfounded[]` matching. `adjudicated[]` reintroduces the undefined "fingerprint" concept for the one guarantee with its own dedicated regression test (never-ask-twice), using exactly the identity scheme the document itself says breaks under line shift.

**Why it matters:** if `adjudicated[]` lookups miss after a fix shifts surrounding line numbers, the same already-answered `UNVERIFIABLE` question gets re-asked via `AskUserQuestion` in a later round — turning the unattended loop into a repeated prompt loop, contradicting the explicit "never asked twice" guarantee this section states.

**Suggested fix:** define `fingerprint` explicitly (e.g. normalized-path + Jaccard-bucket, matching the scheme already used for `retired[]`/`rejected[]`/`unfounded[]`), or drop the separate "fingerprint" vocabulary and state that `adjudicated[]` is matched via the same path+Jaccard rule as the rest of the ledger.

commands/converge.md
```diff
@@ -0,0 +1,412 @@
...
+jq -c '.[] | select(.anchor == "inline")' "$RUN_DIR/round-${N}-findings.json" | while IFS= read -r row; do
+  ROW_PATH=$(printf '%s' "$row" | jq -r '.path')
+  ROW_LINE=$(printf '%s' "$row" | jq -r '.line')
+  [ -f "$ROW_PATH" ] || { echo "UNFOUNDED (no such file): $ROW_PATH"; continue; }
+  printf '%s\n' "$DIFF_FILES" | grep -qxF "$ROW_PATH" \
+    || { echo "OUT-OF-DIFF (excluded from predicate): $ROW_PATH"; continue; }
+  TOTAL_LINES=$(awk 'END{print NR}' "$ROW_PATH")
+  [ "$ROW_LINE" -le "$TOTAL_LINES" ] || { echo "UNFOUNDED (line out of range): $ROW_PATH:$ROW_LINE"; continue; }
+  CONTENT=$(sed -n "${ROW_LINE}p" "$ROW_PATH")
+  RESOLVED=$(swe-workbench-diff-line-lookup "$ROW_PATH" "$CONTENT" --range="origin/$BASE...HEAD" 2>/dev/null)
```
lugassawan, line 205: **Medium | Anchor-validation feeds the raw full line as the match pattern, which can false-positive on ambiguity and drop a valid finding**

Phase 1b feeds `swe-workbench-diff-line-lookup` the entire raw line content (`CONTENT=$(sed -n "${ROW_LINE}p" "$ROW_PATH")`) rather than a narrowed/distinctive snippet. The script's own contract does a substring match per `+` line and errors out with no stdout when the pattern matches more than one added line in the file's diff — common for short lines (`}`, blank lines, `pass`, `return`) that recur across a multi-hunk diff. When that happens, `$RESOLVED` is empty, the mismatch check fails, and a genuinely valid, in-diff finding is silently reclassified `UNFOUNDED (anchor not in diff)` and dropped before ever reaching the fixer.

**Why it matters:** a real, correctly-anchored finding whose line content isn't unique within that file's diff is silently discarded and reported as "the claim did not hold against the code" — factually false, and exactly the false-negative failure mode the mechanical check exists to avoid, just inverted.

**Suggested fix:** narrow the snippet before calling the tool (trim to a distinctive substring, or fall back to hunk-membership-by-line-number parsing when the full-line match is ambiguous) rather than relying on content uniqueness alone.

commands/converge.md
```diff
@@ -0,0 +1,412 @@
...
+- **One batched `AskUserQuestion` per round, never one per finding.** Up to 4 findings become the
+  4 questions of a single call. Each offers *Real — fix it* / *Not real — drop it* / *Real, don't
+  fix now*, with the finding body and the agent's investigation notes as context.
+- **More than 4 `UNVERIFIABLE` findings in one round → stop the loop** and jump to Phase 6
+  (too-many-unverifiable). At that density the review needs a human reading, not an
+  interrogation.
+- **Answers are cached in the ledger's `adjudicated[]`, keyed by fingerprint, and replayed into
+  later rounds' suppression block — the same question is never asked twice.** Without this, an
+  unattended 4-round loop would re-ask the same unanswerable thing every round.
+- Answers map to normal dispositions: *Real* → fixed this round; *Not real* → `unfounded[]` with
```
lugassawan: **Medium | "Real → fixed this round" doesn't specify whether a fix is actually dispatched**

Phase 3b's answer-mapping table states "*Real* → fixed this round" without specifying whether this triggers an actual re-dispatch of a fix (another `code-impl` call scoped to the confirmed finding) or is purely ledger bookkeeping. If it's the latter, a finding the user just confirmed is real gets recorded as `fixed`/suppressed with no code ever changed — which would then satisfy Phase 2's termination predicate or Phase 4's zero-edit-stop bookkeeping incorrectly, defeating the purpose of the escalation.

**Why it matters:** an ambiguous "fixed this round" risks the loop treating a human-confirmed real bug as resolved without ever emitting a code change, silently shipping the unfixed defect.

**Suggested fix:** state explicitly that a "Real" answer triggers a follow-up fix dispatch (scoped to just that finding) before the round's dispositions are finalized — not merely a ledger status flip.

lugassawan (commented):
_Reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 4 inline comment(s) and 0 PR-level note(s), deduped 0.

commands/converge.md
lugassawan: Addressed in 29f117fb7d4f3998f8e635b3afc3ca213c99160c: distinguish a genuine "no PR" from any other gh pr view failure and fail closed on the latter.

commands/converge.md
lugassawan: Addressed in 29f117fb7d4f3998f8e635b3afc3ca213c99160c: defined fingerprint as the same normalized-path + Jaccard identity used elsewhere in the ledger, not line-sensitive.

commands/converge.md
lugassawan: Addressed in 29f117fb7d4f3998f8e635b3afc3ca213c99160c: ambiguous diff-line-lookup matches now fall back to hunk-range membership instead of UNFOUNDED.

commands/converge.md
lugassawan: Addressed in 29f117fb7d4f3998f8e635b3afc3ca213c99160c: a Real adjudication now dispatches a scoped fix and only records FIXED once it lands.

commands/converge.md
```diff
@@ -0,0 +1,456 @@
...
+destructive-by-default hazard, not a style issue. This is a **stronger** gate than
+`swe-workbench:workflow-address-feedback`'s owner check, which only shapes prompts.
+
+```bash
+gh auth status >/dev/null 2>&1 || {
+  echo "converge: gh is not authenticated — run 'gh auth login' first." >&2
+  exit 1
+}
+CURRENT_USER=$(gh api /user -q .login)
+PR_VIEW_OUT=$(gh pr view --json author,number 2>&1)
```
lugassawan: **High | The ownership-gate fix reintroduces the exact bug it was meant to close, via a different path**

`gh pr view --json author,number 2>&1` now merges stderr into the same string fed to `jq -r '.author.login'` on the success path (`PR_VIEW_EXIT -eq 0`) — the prior code used `2>/dev/null` here. If `gh` writes anything to stderr on a successful call (an update-available notice, a deprecation warning), the merged text is no longer valid JSON, `jq` fails silently (no exit-code check), and `PR_AUTHOR`/`PR_NUM` both come back empty. The `[ -n "$PR_AUTHOR" ]` check then reads that as "no PR yet" and lets the loop **proceed** — the same "collapse ambiguous failure into proceed" bug this commit closed, reintroduced through parse corruption on the success branch instead of a non-zero exit code.

**Why it matters:** undermines the "refuse, do not warn" ownership gate — a PR authored by someone else could be silently treated as "no PR" whenever `gh` emits any stderr noise on a 0-exit call.

**Suggested fix:** capture stdout and stderr separately and check jq's own exit status before trusting `PR_AUTHOR`/`PR_NUM`; treat a jq parse failure on `PR_VIEW_EXIT -eq 0` as the same "could not verify" fail-closed case, not as empty/no-PR.

lugassawan (commented):
_Re-reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comment(s) and 0 PR-level note(s), deduped 0.

commands/converge.md
```diff
@@ -0,0 +1,456 @@
...
+   correctly — return `UNFOUNDED` with the actual code as evidence. Do not "fix" your way into
+   agreement with a wrong claim.* Every disposition must cite evidence: `FIXED` states what the
+   code did and now does; `UNFOUNDED` quotes what is actually at `path:line`.
+4. **Explicit prohibitions** — the most important paragraph in the brief: never delete, skip,
+   `xfail`, `@Ignore`, weaken an assertion, lower a coverage threshold, or broaden a lint-ignore
+   to make a finding go away. If a finding can only be satisfied that way, return `REJECTED` with
+   the reason. No new features, no drive-by refactors.
+5. **Return format:** one line per finding —
+   `<fingerprint> | FIXED <what changed> | UNFOUNDED <what is actually there> | REJECTED <reason> | DEFERRED <reason> | UNVERIFIABLE <what was investigated, and the specific question the codebase cannot answer>`.
+   **`fingerprint` is the same normalized-path + Jaccard-bucket identity "Cross-round dedup" below
```
lugassawan: **Medium | `fingerprint` is still not a computable value, just reworded**

The fix defines `fingerprint` as "the same normalized-path + Jaccard-bucket identity" the dedup section uses — but that section only defines a pairwise match predicate (same normalized path AND Jaccard ≥ 0.4 between a new finding and a ledger entry), not a computable per-finding identity. Jaccard similarity isn't transitive either, so there's no well-defined "bucket" a finding can be deterministically assigned to. The Phase 3 return line and `adjudicated[]`'s keying both need a single string an agent can emit and later look up — this still doesn't say how to derive one.

**Why it matters:** without a concrete algorithm, the fixer has no deterministic way to produce this value; a likely fallback (normalized path alone) would silently collapse multiple distinct findings in the same file into one identity, breaking suppression, oscillation detection, and the never-ask-twice guarantee whenever a file has more than one finding across rounds.

**Suggested fix:** spell out a concrete algorithm the agent can compute today (e.g. `normalized_path + ":" + sha256(first-N-tokens-of-body)[:8]`), or state explicitly that `fingerprint` is match-time-only — there's no static key, the orchestrator always re-runs the Jaccard comparison, and `<fingerprint>` in the return line is just `path`, a coarse pre-filter before the Jaccard check.

commands/converge.md
```diff
@@ -0,0 +1,456 @@
...
+tokens. For each row in `round-<N>-findings.json` with `anchor: inline`:
+
+| Check | Fails → |
+|---|---|
+| `path` exists in the working tree | `UNFOUNDED (no such file)` |
+| `path` ∈ `git diff origin/$BASE...HEAD --name-only` | out-of-diff → excluded from the convergence predicate (not unfounded) |
+| `line` ≤ the file's line count | `UNFOUNDED (line out of range)` |
+| the line at `path:line` resolves inside a diff hunk | `UNFOUNDED (anchor not in diff)` |
+
+The last check reuses `bin/swe-workbench-diff-line-lookup`, and — critically — confirms the
```
lugassawan, line 189: **Medium | The ambiguous-match fallback contradicts the stronger guarantee claimed just above it**

The prose above the anchor-validation table claims the check "confirms the resolved line number matches the finding's claimed line, not merely that some added line in the file contains matching text." The ambiguous-match fallback contradicts that for its own trigger case: on an ambiguous diff-line-lookup result, it falls back to checking only that the claimed line falls within some added hunk's numeric range — which includes unchanged context lines inside that hunk, not just `+` lines.

**Why it matters:** trades "false UNFOUNDED on a genuinely ambiguous-but-correct anchor" for "false anchor-valid on a line that was never actually added by this diff" — the same failure mode the surrounding prose says is avoided, just relocated.

**Suggested fix:** tighten the hunk-range fallback to check the line is specifically on a `+` line (track additions within the hunk the same way `diff-line-lookup` itself does), or soften the prose so it doesn't claim a stronger guarantee than the fallback path actually provides.

commands/converge.md
lugassawan: Addressed in d09a543: capture gh pr view's stdout and stderr separately so stderr noise on success can never corrupt the JSON parse, and check jq's own exit status before trusting the fields.

commands/converge.md
lugassawan: Addressed in d09a543: fingerprint is now defined concretely as the finding's path, with every lookup re-running the same path+Jaccard>=0.4 match predicate rather than a separate bucket algorithm.

commands/converge.md
lugassawan: Addressed in d09a543: the ambiguous-match fallback now confirms the claimed line is itself an added (+) line, tracking new-file line numbers like diff-line-lookup does, instead of just checking hunk-range membership.

PR #631, opened 2026-08-21
[feat] Pi: subagent dispatch (task tool) (#610)

### LINKED ISSUE #610: [feat] Pi: subagent dispatch + provider-namespaced model-tier mapping
## Goal

Dispatching `swe-workbench:reviewer` (or any of the 22 agents) from a Pi session runs a first-party
nested session with the agent's declared `tools`, `skills:` preload, and `@../shared/agents/*.md`
includes resolved — honoring the CC-only `model:` tier intent (17 `sonnet` / 5 `haiku`) only for users
who opt in, and falling back safely for everyone else.

## Motivation

Part of the Pi Coding Agent port (see umbrella issue). Pi exposes `createAgentSession` /
`SessionManager.inMemory()` for first-party nested sessions — this phase is the payoff of Phase 2's
golden-fixture contract: the runtime parser that derives an agent definition from `agents/*.md` is the
same parser Phase 2 snapshots, so a frontmatter drift that breaks Pi dispatch fails CI before merge.

`agents/*.md` `model:` values (`sonnet`/`haiku`) are Claude-Code-only aliases with no portable meaning
on Pi, where a session may run on any provider (Anthropic, OpenAI, Google, local llama.cpp, …). This
phase adds an **optional, user-configured, provider-namespaced tier mapping**:
`swe-workbench.modelTiers.<provider>.{sonnet, haiku}` in Pi settings (`~/.pi/agent/settings.json` or
project-local `.pi/settings.json`). Pi model IDs are already `provider/id`-shaped, so the provider
segment is a natural, non-arbitrary key — not a flat `{sonnet, haiku}` map, which would collide the
moment a user runs Pi against more than one provider. `subagent.ts` parses the **active session's
current provider** from its `provider/id` model string and resolves
`modelTiers[activeProvider]?.[tier] ?? parentSessionModel` — it never falls across providers, and an
unconfigured install runs every agent on the parent session's active model with no error.

`bin/swe-workbench-pi-setup` is the bootstrap: a bare-command, idempotent script (matching the
existing `bin/swe-workbench-*` convention) that detects the settings file and the currently active
provider, then merges — never overwrites — that provider's `modelTiers` sub-block with default
tier→model-id mappings, in a stable, deterministic key order so re-runs and hand-edits diff cleanly.
Other providers' previously-configured sub-blocks are left untouched. When a provider ships a newer
model version than the script's defaults, the user edits the mapped value directly — a plain
`tier: "model-id"` string, no schema migration required.

## Surface

- [ ] Skill — `principle-*`
- [ ] Skill — `language-*`
- [ ] Skill — integration (`swe-workbench:ticket-context`)
- [ ] Skill — `workflow-*`
- [x] Subagent
- [ ] Slash command
- [ ] Hook
- [x] Build / CI / release (`bin/swe-workbench-pi-setup`)

## Sketch (optional)

`pi/extensions/subagent.ts`: dispatch → parse `agents/<name>.md` (via Phase 2's parser) → resolve
`tools`/`skills:`/includes → resolve model via provider-namespaced `modelTiers` lookup or fallback →
`createAgentSession`. `bin/swe-workbench-pi-setup [--provider <name>]`: detect settings file + active
provider → merge default tier map into `modelTiers.<provider>` without clobbering existing keys.

## Acceptance (optional)

- [ ] Dispatching `swe-workbench:reviewer` runs with its declared skills preloaded and declared tool
      set
- [ ] Phase 2's golden fixture gates the parse (a frontmatter drift fails CI, not silently at dispatch
      time)
- [ ] An unconfigured Pi install runs every agent on the parent session's model with no error
- [ ] Running `swe-workbench-pi-setup` twice in a row produces no diff on the second run
- [ ] Hand-editing a `modelTiers.<provider>` value and re-running the script preserves the hand-edit
- [ ] Configuring both `anthropic` and `openai` sub-blocks, then switching `--provider`, resolves each
      agent to the correct provider's model with no cross-provider bleed
- [ ] All extension state (including any per-dispatch caching) is session-scoped — #401's
      module-scoped `hintedSkills` Set is the named anti-pattern to avoid repeating here

**Depends on:** Phase 0 (#603), Phase 2 (#605), Phase 5 (#608), Phase 6 (#609)

lugassawan:
## Summary

- Adds a first-party `task` tool (`pi/extensions/subagent.ts` + `pi/extensions/agent-spec.ts`) that dispatches any of this plugin's `agents/*.md` definitions (e.g. `swe-workbench:reviewer`) as a nested Pi child process, preserving that agent's declared `tools:` and preloaded `skills:` content
- The third-party `pi-subagents` package's `skills:` field only makes a skill *available* on demand — it never preloads skill body into context, which this repo's agent convention requires (`docs/skill-preload.md`); this first-party dispatcher exists specifically to close that gap
- Transport is `pi.exec()` (never raw `node:child_process`); recursion through `task`/`subagent` is blocked via `--exclude-tools task,subagent` on the child's argv (verified against the SDK's tool-registry construction, and via a live zero-model-call CLI probe); the composed system prompt is delivered via `--append-system-prompt <tempfile>` (never inline text) to avoid `MAX_ARG_STRLEN` and a silent-overflow failure mode in the SDK's own exec wrapper
- `pi/extensions/tool-vocab.ts` and `index.ts` updated so the anti-hallucination preamble prose correctly names the real `task` tool only when it is actually live in the session's tool registry (checked via `pi.getActiveTools()`, not just the kill-switch env var)
- **Model-tier mapping**: a dispatched agent's `model: haiku/sonnet/opus` frontmatter now resolves to a real per-provider model, via a hardcoded name table in `pi/extensions/model-tier.ts` (`resolveModelForTier`) — scoped to the parent session's active provider (`ctx.model.provider`) and only ever selecting from already-authenticated models (`ctx.scopedModels` when the session is scoped, else `ctx.modelRegistry.getAvailable()`), never a new provider/endpoint/credential; a stale or unresolvable table entry falls back to the parent's own model unchanged
- Preloaded skill sections now state their absolute on-disk directory, so a skill's own bare `examples/` pointer (e.g. `swe-workbench:principle-solid`) is resolvable by a dispatched child with `read`, instead of being a dead reference
- `docs/plugin-platform-decisions.md` §9 records the full design rationale, including the model-tier-mapping safety posture and an accepted, tracked recursion gap: an agent granted `Bash` can still shell out to `pi -p ...` directly for a further, unguarded level of recursion — tracked as a follow-up, not fixed here
- Addressed PR review feedback: a CI test drives the real hand-rolled parser against every actual `agents/*.md` file (not just synthetic fixtures), `capOutput`'s truncation no longer splits a UTF-16 surrogate pair at the cap boundary, and the recursion-gap deferral is now explicit sign-off rather than implicit

## Test Plan

- [x] `bash scripts/validate.sh && npm run typecheck --prefix pi` — 0 issues
- [x] `pytest tests/ -v` — 2768 passed, 2 skipped (pre-existing root-only skips, unrelated)
- [x] Manual end-to-end dispatch against the real `pi` CLI (model `openai-codex/gpt-5.4-mini`): the model called `task` for the `e2e-test-verifier` agent, the child ran with the translated tool set (`read,grep,find,bash,ask_user_question`) and `--exclude-tools task,subagent` visible in the captured argv, and the child's response echoed back the exact `preload-canary` token of its one preloaded skill — confirming real preload injection, not just a passing unit test
- [x] Live zero-model-call CLI probe (`tests/test_pi_contract.py::test_exclude_tools_structurally_prevents_task_tool_activation`) confirms `--exclude-tools` structurally removes the real `task` tool from `pi.getActiveTools()`
- [x] Manual end-to-end model-tier dispatch: parent session running on `openai-codex/gpt-5.4` dispatched a `model: haiku` agent (`e2e-test-verifier`) and the captured child argv showed `--model openai-codex/gpt-5.6-luna`, confirming tier resolution against real, live credentials and a real model registry

- [x] New behaviour is covered by tests, not just manually verified
- [x] No existing tool/command behaviour changed unexpectedly

Closes #610

### DIFF
--- a/pi/extensions/agent-spec.ts
+++ b/pi/extensions/agent-spec.ts
@@ -1,0 +1,168 @@
+/**
+ * Pure parsing/composition for this plugin's agents/*.md convention, plus thin I/O wrappers.
+ *
+ * Layer: domain, same posture as tool-vocab.ts. This file must import NOTHING from the Pi SDK,
+ * not even as a type — only node:fs/node:path, plus a relative import of RENAME_TABLE from
+ * ./tool-vocab.ts (also SDK-free). subagent.ts owns everything that touches Pi itself.
+ *
+ * agents/*.md frontmatter is a fixed, uniform 5-key shape across all 22 files today (name,
+ * description, model, tools, skills) — `tools:` is always one comma-separated inline string,
+ * `skills:` is always a YAML block sequence of `swe-workbench:<id>` entries. This is a small
+ * purpose-built extractor for exactly that shape, not a general YAML parser — it mirrors
+ * scripts/validate.py's parse_frontmatter() key/item regexes so the two hand-rolled parsers
+ * can't silently diverge on what a line means.
+ */
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { RENAME_TABLE } from "./tool-vocab.ts";
+
+export interface AgentSpec {
+  readonly name: string;
+  readonly description: string;
+  readonly tools: readonly string[];
+  readonly skillIds: readonly string[];
+  readonly body: string;
+}
+
+const SKILL_NAMESPACE_PREFIX = "swe-workbench:";
+
+/** Tool tokens with no Pi tool equivalent — dropped rather than renamed. Exported so
+ *  tests/test_pi_contract.py can assert exhaustiveness over TOOL_TOKENS. */
+export const DROP_TOKENS = new Set(["Skill", "WebFetch"]);
+
+const FM_KEY_RE = /^([\w-]+):\s*(.*)$/;
+const FM_ITEM_RE = /^-\s+(.*\S)\s*$/;
+
+/** Mirrors the installed Pi SDK's dist/utils/frontmatter.js extractFrontmatter (kept as prose
+ *  here, deliberately not a code dependency — see this file's own header): normalize newlines,
+ *  find the closing '---' via indexOf from index 3, slice from 4 chars past that match, then
+ *  strip() the body. Throws (rather than degrading) on a missing/malformed block — every file
+ *  this is called on is one of this plugin's own bundled resources, not untrusted input, so a
+ *  malformed one is a bug to surface, not a case to silently tolerate. */
+function splitFrontmatter(rawText: string): { frontmatter: string; body: string } {
+  const text = rawText.replace(/\r\n/g, "\n").replace(/\r/g, "\n");
+  if (!text.startsWith("---")) {
+    throw new Error("expected a '---' frontmatter block at the start of the file");
+  }
+  const end = text.indexOf("\n---", 3);
+  if (end === -1) {
+    throw new Error("frontmatter block has no closing '---'");
+  }
+  return { frontmatter: text.slice(3, end), body: text.slice(end + 4).trim() };
+}
+
+/** Parses exactly this plugin's agents/*.md frontmatter shape: `tools:` a comma-separated
+ *  inline string, `skills:` a block sequence, everything else a plain scalar. Pure. */
+export function parseAgentSpec(text: string): AgentSpec {
+  const { frontmatter, body } = splitFrontmatter(text);
+
+  const scalars: Record<string, string> = {};
+  const lists: Record<string, string[]> = {};
+  let pending: string | null = null;
+
+  for (const rawLine of frontmatter.split("\n")) {
+    const stripped = rawLine.trim();
+    if (stripped === "") continue;
+
+    if (pending !== null) {
+      const item = FM_ITEM_RE.exec(stripped);
+      if (item) {
+        (lists[pending] ??= []).push(item[1].trim());
+        continue;
+      }
+    }
+
+    const key = FM_KEY_RE.exec(stripped);
+    if (key) {
+      const name = key[1].toLowerCase();
+      const value = key[2].trim();
+      scalars[name] = value;
+      pending = value === "" ? name : null;
+    } else {
+      pending = null;
+    }
+  }
+
+  const name = scalars.name;
+  const description = scalars.description;
+  if (!name) throw new Error("agent frontmatter missing required 'name' key");
+  if (!description) throw new Error("agent frontmatter missing required 'description' key");
+
+  const tools = scalars.tools
+    ? scalars.tools
+        .split(",")
+        .map((t) => t.trim())
+        .filter((t) => t.length > 0)
+    : [];
+  const skillIds = lists.skills ?? [];
+
+  return { name, description, tools, skillIds, body };
+}
+
+/** Translates Claude Code tool tokens (as found in an agent's `tools:` frontmatter) into Pi
+ *  tool names, via RENAME_TABLE for tokens with a real Pi equivalent and DROP_TOKENS for tokens
+ *  with none. Throws on any token that is neither — exhaustive over this repo's known token
+ *  vocabulary, no silent gaps — and throws if the translated result is empty (an empty --tools
+ *  allowlist would silently mute every tool in the dispatched child session). Pure; order- and
+ *  duplicate-preserving beyond de-duplication. */
+export function translateToolTokens(tokens: readonly string[]): string[] {
+  const seen = new Set<string>();
+  const translated: string[] = [];
+  for (const token of tokens) {
+    const renamed = RENAME_TABLE.find(([cc]) => cc === token)?.[1];
+    if (renamed !== undefined) {
+      if (!seen.has(renamed)) {
+        seen.add(renamed);
+        translated.push(renamed);
+      }
+      continue;
+    }
+    if (DROP_TOKENS.has(token)) continue;
+    throw new Error(`translateToolTokens: no Pi mapping (rename or drop) for tool token "${token}"`);
+  }
+  if (translated.length === 0) {
+    throw new Error(
+      "translateToolTokens: translated result is empty — refusing to build an empty --tools allowlist",
+    );
+  }
+  return translated;
+}
+
+/** Composes an agent's system prompt: its own body, followed by each preloaded skill's
+ *  (frontmatter-stripped) body in `skills:` order. Pure string composition. */
+export function composeSystemPrompt(
+  spec: Pick<AgentSpec, "body">,
+  skills: ReadonlyArray<{ readonly id: string; readonly body: string }>,
+): string {
+  const sections = [spec.body.trim()];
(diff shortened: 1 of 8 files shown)


pi/extensions/subagent.ts
```diff
@@ -0,0 +1,188 @@
...
+        writeFileSync(promptFile, systemPrompt, { mode: 0o600 });
+
+        const args = [
+          "-p",
+          prompt,
+          "--append-system-prompt",
+          promptFile,
+          "--tools",
+          toolNames.join(","),
+          "--exclude-tools",
```
lugassawan: The recursion guard (`--exclude-tools task,subagent`) only blocks recursion through the `task`/`subagent` tools; it does nothing to stop a dispatched agent that also carries `bash` (the translated form of `Bash`) from shelling out to `pi -p ...` directly, which spawns a fresh child session with `task` re-registered and no `--exclude-tools` at all. Roughly 18 of the 22 `agents/*.md` definitions (`reviewer`, `debugger`, `refactorer`, `auditor`, `migrator`, `code-impl`, etc.) carry `Bash`, so this covers the large majority of real dispatches, not an edge case.

Why it matters: issue #610's acceptance criteria state "recursion must be blocked (agents can't dispatch further task/subagent calls)" — for Bash-granted agents that criterion isn't actually met: a misbehaving or prompt-injected dispatched agent can recurse unboundedly, each level spawning its own model-tier-resolved child process, which is a real cost/resource-exhaustion vector.

Note: this gap is explicitly named and accepted in `docs/plugin-platform-decisions.md:224` ("Accepted gap, not closed here") with a concrete follow-up plan (close via `hooks/bash_guard.sh` pattern-matching a `pi ... -p`/`--print` invocation). Flagging for explicit sign-off that the acceptance criterion is being closed across two PRs rather than one, not because the gap is undocumented.

Suggested fix: either land the `bash_guard.sh` follow-up in the same PR, or explicitly note in the issue/PR that this acceptance criterion is deferred to a tracked follow-up before merging.

pi/extensions/agent-spec.ts
```diff
@@ -0,0 +1,189 @@
...
+      `## Preloaded skill: ${skill.id}\n` +
+        `(relative paths this skill's body mentions, e.g. \`examples/...\`, resolve against: ${skill.dir})\n\n` +
+        skill.body.trim(),
+    );
+  }
+  return sections.join("\n\n---\n\n");
+}
+
+/** Reads and parses agents/<name>.md from this plugin's own bundled resource tree. */
+export function readAgentSpec(root: string, name: string): AgentSpec {
```
lugassawan: No test drives the real hand-rolled TS parser (`parseAgentSpec`/`readAgentSpec`/`translateToolTokens`) against the actual bundled `agents/*.md` files — `tests/test_pi_extension.py`'s parser tests use only synthetic fixtures, and `tests/test_pi_contract.py`'s exhaustiveness ratchets cross-check against `validate.parse_frontmatter()` (the separate Python parser), never the TS one, on real files.

Why it matters: a future agent frontmatter edit that's valid under the Python parser's rules but breaks an assumption the hand-rolled TS regex parser makes (e.g. a block-style `tools:` sequence instead of the assumed inline comma string) would stay CI-green and only surface as a runtime throw the first time that agent is actually dispatched — precisely the "frontmatter drift should fail CI, not silently break at dispatch time" failure mode issue #610 calls out. (Confirmed empirically: all 22 real files parse cleanly today, so this is a coverage gap rather than a live bug.)

Suggested fix: add a CI test to `tests/test_pi_contract.py` that runs the real `readAgentSpec` + `translateToolTokens` (Node driver, same pattern as `test_task_tool_translation_table_is_exhaustive_over_tool_tokens`) against every file under `agents/*.md` and asserts none throw.

pi/extensions/subagent.ts
```diff
@@ -0,0 +1,188 @@
...
+const PI_SUBAGENTS_TOOL_NAME = "subagent";
+
+const TASK_TIMEOUT_MS = 15 * 60 * 1000;
+const OUTPUT_CAP_CHARS = 50_000;
+
+/** Applied to both the success path's stdout and the failure path's stderr — a dispatched
+ *  child's output (or its error output on a bad exit) becomes part of the PARENT model's
+ *  context, so an uncapped dump either way is the same context-bloat risk. */
+function capOutput(text: string): string {
+  return text.length > OUTPUT_CAP_CHARS
```
lugassawan: `capOutput` truncates with `text.slice(0, OUTPUT_CAP_CHARS)`, a UTF-16 code-unit slice that can land exactly inside a surrogate pair (e.g. an emoji) near the 50,000-char boundary, leaving a lone surrogate in the returned/logged text.

Why it matters: cosmetic — a malformed character could appear in the parent's context or trip up downstream JSON serialization of the tool result, but only in the rare case a multi-byte character sits right at the cap.

Suggested fix: back off to the last valid code-point boundary before truncating, e.g. `text.slice(0, OUTPUT_CAP_CHARS).replace(/[\uD800-\uDBFF]$/, "")`.

lugassawan (commented):
_Reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comment(s) and 0 PR-level note(s), deduped 0.

pi/extensions/subagent.ts
lugassawan: Addressed in b67bdb4: referenced the tracked follow-up issue (#632) for the bash-escape-hatch recursion gap and corrected the accepted-gap paragraph in docs/plugin-platform-decisions.md §9 to the real count (20 of 22 agents grant Bash, not just the 4 previously named). Not closing the gap itself within this PR, per your call on scope.

pi/extensions/agent-spec.ts
lugassawan: Addressed in b67bdb4: added test_real_agents_parse_and_translate_without_throwing to tests/test_pi_contract.py — drives the real readAgentSpec/translateToolTokens against every actual agents/*.md file on disk (not synthetic fixtures, not the separate Python parser), asserting none throw.

pi/extensions/subagent.ts
lugassawan: Addressed in b67bdb4: capOutput now strips a dangling lone high surrogate before appending the truncation marker, so a surrogate pair straddling the cap boundary never produces ill-formed UTF-16. Added test_cap_output_does_not_split_a_surrogate_pair_at_the_boundary — a direct unit test on the exported capOutput (avoids the stdout/JSON/UTF-8 round trip, which doesn't preserve a lone surrogate reliably); confirmed it fails without the fix.

PR #718, opened 2026-09-04
[feat] Repo-scope ephemeral state and cleanup

### LINKED ISSUE #713: [feat] repo-aware ephemeral state: scope /tmp/swe-workbench-* artifacts and cleanup by repository
## Problem

All plugin ephemeral artifacts are keyed only by PR number and flow tags — never by repository:

- Preflight state files: `/tmp/swe-workbench-pr-review/<N>.json`, `<N>-followup.json`, `<N>-review-<mode>.json`
- Run dirs: `/tmp/swe-workbench-run/<tag>-<N>-<rand>/` (e.g. `pr-followup-<N>-<rand>`, `address-feedback-<N>-<rand>`)

Consequences observed in a live review session:

1. **Cross-repo ambiguity** — two repos can legitimately both have PR `<N>` (a small plugin repo vs a large internal monorepo). `swe-workbench-sweep-residuals <N>` and other cleanup paths can only guess whether an artifact belongs to "this" repo, so the agent has to doubt and manually research ownership before removing anything.
2. **Lingering foreign artifacts** — during session cleanup, a run dir belonging to a *different* repo's flow (`address-feedback-<N>-…`) could not be safely attributed or reaped, because nothing on disk records which repo created it.
3. **Collision risk** — same-numbered PRs reviewed concurrently in different repos can race on identical PR-keyed filenames.

## Value

Anyone running review/feedback flows across multiple repositories gets deterministic, repo-scoped cleanup — no guessing by PR number, no cross-repo deletion risk, and no orphaned foreign artifacts accumulating in `/tmp`.

## Acceptance criteria

- Every new `/tmp/swe-workbench-*` state filename and run-dir name embeds a normalized owner-repo slug (e.g. `/tmp/swe-workbench-run/<tag>-<owner-repo>-<N>-<rand>/`, `/tmp/swe-workbench-pr-review/<owner-repo>-<N>.json`), keeping concurrent same-`<N>` reviews in different repos collision-free.
- Cleanup tools (`swe-workbench-sweep-residuals`, `swe-workbench-reap-run-dir`, `swe-workbench-clean-state-files`, and their skill call-sites) accept or derive the repo scope and never remove artifacts belonging to a different repository.
- Legacy un-scoped artifacts are handled backward-compatibly (migrated or swept under an explicit legacy rule) without breaking existing flows.
- The naming/migration convention is documented in `shared/docs/` so downstream skills don't drift.

## Impact / Effort

- **Impact: M** — prevents wrong-repo deletion (safety for in-flight review runs) and removes recurring cleanup friction; each individual incident is recoverable, so not L.
- **Effort: M** — touches `swe-workbench-new-run-dir`, preflight scripts, `sweep-residuals`, `reap-run-dir`, filename interpolation across several SKILL.md contracts, plus a backward-compat migration path.

## Additional context

- Observed during a live PR-review session: cleanup found a foreign `address-feedback-<N>-<rand>` run dir from another repo's flow and had to leave it untouched because ownership was not recorded on disk.
- `swe-workbench-sweep-residuals` is PR-number-scoped by contract ("safe to force-remove anything keyed to that specific N"), which silently assumes a single-repo world.
- Related surfaces: `/tmp/swe-workbench-run/` (Block C sweep), `/tmp/swe-workbench-pr-review/` state files, and `RUN_DIR` allocations in `workflow-pr-review` (both modes), the specialist post sub-flow, and `workflow-address-feedback`.

---
_Reported via `/swe-workbench:report-issue` — plugin v0.1.37, Pi 0.84.4._

lugassawan:
## Ticket context: #713

**Title:** [feat] repo-aware ephemeral state: scope /tmp/swe-workbench-* artifacts and cleanup by repository
**Type / Status:** enhancement / OPEN
**Summary:** All plugin ephemeral artifacts (preflight state files, run dirs, git-fallback worktree dirs) were keyed only by PR number and flow tag — never by repository — so cleanup could not attribute ownership across repos, foreign artifacts lingered, and concurrent same-numbered PRs could collide.
**Source:** https://github.com/lugassawan/swe-workbench/issues/713

## Summary
- Every `/tmp/swe-workbench-*` ephemeral artifact now embeds a normalized owner-repo slug via a new Tier-S helper `swe-workbench-repo-scope` (ladder: explicit `--repo` → PR-JSON `url` → `git remote get-url origin` → empty ⇒ legacy un-scoped naming; owner is length-prefixed before the join so the mapping stays genuinely injective for any owner/repo content — plain `tr '/' '-'` would let e.g. `acme-app/foo` and `acme/app-foo` collide, and a hyphen-doubling escape tried first is *also* not injective, e.g. `owner='a-'/repo='b'` and `owner='a'/repo='-b'` both produce `a---b`): run dirs `<tag>-<slug>-<N>-<rand>`, state files `<slug>-<N>*.json`, git-fallback worktree dirs `<slug>-<N>` (rimba labels unchanged — per-repo git objects already). All three `resolve_*` paths (`--repo`, `--pr-json`, cwd origin) now share one shape validator so a `.github`-style repo name resolves consistently everywhere.
- Cleanup is repo-scoped and dual-read: `swe-workbench-sweep-residuals` accepts `--repo`/`--head-sha`, sweeps slugged names by exact-tag-anchored globs (ours by construction, computed via `swe-workbench-repo-scope` itself rather than a second independent encoding), and sweeps legacy un-scoped artifacts only when their content attributes to the sweeping repo (PR-JSON `url`, `headRefOid` fingerprint, `repository_url`, or the artifact's own git remote). Unattributable legacy artifacts — including triage resume points — are retained with a reason in the additive `data.retained_state_files` envelope field; an explicitly invalid `--repo` fails closed to retain-only.
- Mid-upgrade compat: preflight state files now carry `url` (self-attribution), the address-feedback fetch envelope exposes `resume_triage_path` (legacy resume points stay readable only when a same-numbered legacy preflight JSON's own `url` also attributes them to this repo — the triage file itself has no owner/repo field to check directly; saves migrate to the scoped spelling), and both `pr-review-worktree` and `address-feedback-worktree` release/acquire dual-read legacy fallback dirs/receipts behind the same origin-attribution gate.
- Convention documented in `shared/docs/repo-scoped-ephemeral-state.md`; all six call-sites updated (`workflow-pr-review`, `workflow-address-feedback`, `workflow-cleanup-merged`, `commands/review.md`, `commands/converge.md`, plus envelope-driven paths in the fetch).

## Test Plan
- [x] Full suite green: `python3 -m pytest tests/ -q` → 3851 passed, 5 skipped
- [x] `bash scripts/validate.sh` → all checks passed
- [x] Comment scan re-run → 3 KEEP'd section dividers only

- [x] New behavior pinned: slug ladder (`test_repo_scope.py`), slugged allocation + legacy fallback outside git (`test_new_run_dir.py`), dual-shape reap allowlist (`test_reap_run_dir.py`)
- [x] Cross-repo safety pinned: foreign slugged/legacy artifacts retained across state files, run dirs, and fallback worktrees; slug-suffix collision (`x-acme-app` vs `acme-app`) not swept; invalid `--repo` retain-only (`test_sweep_residuals.py`)
- [x] Migration pinned: preflight JSON `url` self-attribution, `headRefOid` fingerprint vs `--head-sha`, legacy triage dual-read via `resume_triage_path`, legacy receipt/fallback dual-read behind attribution gates
- [x] Contracts intact: `names`↔sweep golden ratchet extended to both spellings; envelope schema additive; bin-scripts ledger and Pi preamble ratchet updated with recorded reason

Closes #713


bin/swe-workbench-repo-scope
```diff
@@ -0,0 +1,94 @@
...
+fi
+
+MODE="${1:-}"
+ARG="${2:-}"
+
+resolve_explicit() {
+  # GitHub owner/repo names start with an alphanumeric (never '.', '-', '_'),
+  # which also rejects traversal-shaped inputs like '../evil'.
+  printf '%s' "$1" | grep -qE '^[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9][A-Za-z0-9._-]*$' || return 1
+  printf '%s\n' "$1" | tr '/' '-'
```
lugassawan: The owner-repo slug (`tr '/' '-'`) is not injective — two distinct owner/repo pairs that split the hyphen boundary differently collide onto the identical slug. Empirically verified: `--repo acme-app/foo` and `--repo acme/app-foo` both print `acme-app-foo`, and this propagates unchanged through `resolve_pr_json` (line ~50) and `resolve_origin` (line ~65).

**Why it matters:** this directly breaks this PR's own acceptance criterion ("keeping concurrent same-N reviews in different repos collision-free") and the "never remove artifacts belonging to a different repository" safety guarantee. `swe-workbench-sweep-residuals` treats any filename/run-dir matching the caller's own `${SCOPE_SLUG}-${N}...` shape as "ours by construction" and sweeps it unconditionally, with zero attribution check (that check only runs for the bare-legacy-number spellings). A live state file for repo `acme/app-foo`'s PR #N and a live state file for repo `acme-app/foo`'s PR #N land on the byte-identical filename `acme-app-foo-N.json`. A routine merged-PR cleanup (`workflow-cleanup-merged` → `swe-workbench-sweep-residuals --repo acme-app/foo`) would force-remove that file, and by the same mechanism force-remove/branch-delete a same-numbered worktree, believing it belongs to the sweeping repo. GitHub org/user names and repo names both commonly contain hyphens, so this is a realistic collision. The new `tests/test_sweep_residuals.py::TestRunDirExactTagMatching::test_slug_suffix_collision_not_swept` covers the *other* collision class (an unanchored glob absorbing a longer foreign slug) but has no test for this identical-slug case — confirming this was not a deliberately-accepted risk, just an untested gap.

**Suggested fix:** escape literal hyphens within each half before joining so the join becomes injective, e.g. `owner.replace('-','--') + '-' + repo.replace('-','--')` (doubling internal hyphens and reserving the single hyphen as the sole separator is a standard, reversible escaping trick — `acme-app/foo` → `acme--app-foo`, `acme/app-foo` → `acme-app--foo`, now distinguishable). Apply the same transform in all three `resolve_*` functions and in `shared/docs/repo-scoped-ephemeral-state.md`'s "The slug" section. Add a regression test asserting the two colliding pairs above resolve to different slugs, and ideally a sweep-residuals test proving repo A's cleanup never removes repo B's artifact when the two would collide under the current `tr`-only scheme.

bin/swe-workbench-address-feedback-fetch
```diff
@@ -307,9 +333,21 @@ def main(argv: list[str] | None = None) -> int:
...
+        # Dual-read: a pre-upgrade session's resume point lives at
+        # the legacy un-scoped triage path — still visible to a scoped run.
+        # resume_triage_path locates where the data actually lives (slugged if
+        # present, else legacy, else empty); triage_path keeps naming where new
+        # saves go, so the resume offer can point at a readable file while every
+        # save migrates to the scoped spelling.
+        resume_triage_path = ""
+        if triage_path.exists():
+            resume_triage_path = str(triage_path)
+        elif slug and legacy_triage_path.exists():
```
lugassawan: `resume_triage_path` / `resume_available` dual-reads the legacy `<PR>-triage.json` with no attribution check, unlike the analogous legacy-artifact handling elsewhere in this same PR (`swe-workbench-sweep-residuals`'s `_attribute_legacy_state "$N-triage.json"` case, and `swe-workbench-pr-review-worktree`'s `_attributes_here`).

**Why it matters:** a legacy triage file left by an unrelated repository's earlier (or pre-upgrade) address-feedback run for the same PR number is silently offered as a resume point, contradicting this skill's own documented claim ("same-numbered PRs in different repositories never collide" — `skills/workflow-address-feedback/SKILL.md:56`). The practical blast radius is softened because triage keys are globally-unique GitHub review-thread node IDs / comment IDs (`skills/workflow-address-feedback/SKILL.md:113,115`), so a foreign repo's decisions can't silently misapply to the current PR's actual threads — but the user still gets a confusing "resume previous session?" prompt sourced from an unrelated repo, and the safety property rests entirely on an implementation detail (ID global-uniqueness) rather than an explicit check. `tests/test_address_feedback_fetch_script.py::test_resume_dual_reads_legacy_triage` only exercises the same-repo case; there's no negative test for a foreign-repo legacy file.

**Suggested fix:** gate the dual-read the same way the sibling scripts gate their legacy artifacts — e.g. skip offering resume from the legacy path unless a same-numbered legacy preflight/worktree-receipt file also exists and attributes to the current scope, or simply mark the resume offer as unavailable once a scope has resolved and no scoped triage file exists yet (forcing a fresh session rather than trusting an unscoped leftover). Add a test that plants a legacy triage file next to a foreign-repo preflight JSON and asserts resume is not silently offered.

lugassawan (commented):
_Reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 2 inline comment(s) and 0 PR-level note(s), deduped 0.

bin/swe-workbench-repo-scope
lugassawan: Addressed in 807572a29f26fb71e3238f7f6d5fd5e3a42b66bb: doubled literal hyphens within owner/repo before joining so the slug stays injective.

bin/swe-workbench-address-feedback-fetch
lugassawan: Addressed in 807572a29f26fb71e3238f7f6d5fd5e3a42b66bb: gated the legacy triage dual-read behind an attribution check against a same-numbered legacy preflight JSON's .url.

lugassawan (commented):
_Re-reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comment(s) and 0 PR-level note(s), deduped 0.

bin/swe-workbench-sweep-residuals
```diff
@@ -117,50 +161,205 @@ SPECIALIST_MODES=(security accessibility dependency performance tests ux)
...
+  SCOPE_SLUG=$("$SCRIPT_DIR/swe-workbench-repo-scope" 2>/dev/null) || SCOPE_SLUG=""
+fi
+PR_REVIEW_DIR="/tmp/swe-workbench-pr-review"
+ADDR_FEEDBACK_DIR="/tmp/swe-workbench-address-feedback"
+
+# GitHub URL forms seen in the state files -> owner-repo slug (same tr-based
+# normalization swe-workbench-repo-scope applies, so comparison stays in slug space).
+_url_scope_slug() {
+  printf '%s' "$1" | sed -nE \
+    -e 's#^https://github\.com/([A-Za-z0-9._-]+)/([A-Za-z0-9._-]+)/pull/[0-9]+/?$#\1-\2#p' \
```
lugassawan: `_url_scope_slug()` reimplements the exact pre-fix non-injective join (`\1-\2` via `sed`, equivalent to plain `tr '/' '-'`) that `escape_owner_repo()` in `bin/swe-workbench-repo-scope` was added to eliminate in this same PR. This reintroduces the collision class the prior High finding fixed, in the very tool that makes actual delete-vs-retain decisions on other repos' artifacts.

**Empirically verified:**
```
$ bash bin/swe-workbench-repo-scope --repo "lugassawan/swe-workbench"
lugassawan-swe--workbench
$ printf '%s' "https://github.com/lugassawan/swe-workbench/pull/718" | sed -nE 's#^https://github\.com/([A-Za-z0-9._-]+)/([A-Za-z0-9._-]+)/pull/[0-9]+/?$#\1-\2#p'
lugassawan-swe-workbench
```
`SCOPE_SLUG` (via the *fixed* `escape_owner_repo`) and `_url_scope_slug`'s output (still undoubled) diverge for this very repo — any repo whose owner or repo name contains a hyphen. Two concrete consequences against `_attribute_legacy_state`'s comparisons (lines ~205, ~227, ~249):
1. **False negative (guaranteed):** for any hyphenated repo, sweep-residuals will never recognize its own legacy `<N>.json`/`<N>-pr-comments.json`/`<N>-threads.json` as belonging to itself — permanently retained instead of swept.
2. **False positive (cross-repo collision, demonstrated):** `SCOPE_SLUG` for `acme/app-foo` (`acme-app--foo`) collides with `_url_scope_slug` of a *different* repo `acme/app--foo` (also `acme-app--foo`):
```
$ bash bin/swe-workbench-repo-scope --repo "acme/app-foo"
acme-app--foo
$ printf '%s' "https://github.com/acme/app--foo/pull/9" | sed -nE 's#^https://github\.com/([A-Za-z0-9._-]+)/([A-Za-z0-9._-]+)/pull/[0-9]+/?$#\1-\2#p'
acme-app--foo
```
This lets sweep-residuals mistake a foreign repo's legacy artifact as its own, violating "cleanup tools never remove artifacts belonging to a different repository." No test in `tests/test_sweep_residuals.py` exercises a hyphenated owner/repo through this `.url`-based attribution path.

**Suggested fix:** route `_url_scope_slug`'s output through the same hyphen-doubling `escape_owner_repo` scheme `bin/swe-workbench-repo-scope` now applies (extract and share the escaping logic, e.g. via a small sourced helper or by shelling out to `swe-workbench-repo-scope --repo "$owner/$repo"`), so both sides of every comparison use the same encoding. Add a regression test with a hyphenated owner/repo through the `.url` attribution path.

bin/swe-workbench-address-feedback-worktree
```diff
@@ -275,7 +296,12 @@ do_release() {
     return
   fi
 
-  local RECEIPT="${STATE_DIR}/${PR}-worktree.json"
+  local RECEIPT
+  RECEIPT="$(receipt_path)"
+  # Dual-read: a pre-upgrade acquire wrote the legacy receipt.
+  if [ ! -f "$RECEIPT" ] && [ "${SCOPE_SLUG:+x}" = "x" ] && [ -f "${STATE_DIR}/${PR}-worktree.json" ]; then
```
lugassawan: The new legacy-receipt dual-read in `do_release` has no attribution check at all, unlike the sibling `_attributes_here` gate correctly applied to `bin/swe-workbench-pr-review-worktree` in this same PR. A foreign repo's stale `<PR>-worktree.json` receipt (any prior run, any repo, same PR number) can silently satisfy the "matching ownership receipt" gate that exists specifically to refuse acting on unverified worktree state.

```bash
local RECEIPT
RECEIPT="$(receipt_path)"
# Dual-read: a pre-upgrade acquire wrote the legacy receipt.
if [ ! -f "$RECEIPT" ] && [ "${SCOPE_SLUG:+x}" = "x" ] && [ -f "${STATE_DIR}/${PR}-worktree.json" ]; then
  RECEIPT="${STATE_DIR}/${PR}-worktree.json"
fi
```

Compare to `bin/swe-workbench-pr-review-worktree`'s `_attributes_here()` (added in the same diff), which gates its equivalent legacy-fallback adoption on the target directory's own git origin matching `$SCOPE_SLUG`, on both `acquire`'s stale-reap and `release`'s path resolution — and has dedicated negative tests (`test_acquire_leaves_foreign_legacy_fallback_alone`, `test_release_foreign_legacy_fallback_not_found`). `shared/docs/repo-scoped-ephemeral-state.md` explicitly claims "every legacy-fallback adoption is attribution-gated — the worktree receipt by its own recorded path's git origin" — this claim is false for `address-feedback-worktree`; only `pr-review-worktree` implements it. `tests/test_address_feedback_worktree_script.py` has a happy-path dual-read test (`test_release_dual_reads_legacy_receipt`) but no foreign-repo negative case, unlike its sibling's coverage.

**Suggested fix:** mirror `pr-review-worktree`'s `_attributes_here` pattern — before adopting the legacy receipt, `cd` into its recorded `.path` and confirm `swe-workbench-repo-scope`'s output there equals `$SCOPE_SLUG`. Add a foreign-repo negative test matching the sibling script's coverage.

bin/swe-workbench-repo-scope
```diff
@@ -0,0 +1,108 @@
...
+
+if [ $# -eq 1 ] && { [ "$1" = "--help" ] || [ "$1" = "-h" ]; }; then
+  awk 'NR>1 && /^#/ { sub(/^# ?/, ""); print; next } NR>1 { exit }' "$0"
+  exit 0
+fi
+
+MODE="${1:-}"
+ARG="${2:-}"
+
+resolve_explicit() {
```
lugassawan: The three `resolve_*` functions validate owner/repo shape inconsistently. `resolve_explicit`'s regex (`^[A-Za-z0-9][A-Za-z0-9._-]*/[A-Za-z0-9][A-Za-z0-9._-]*$`) requires both owner *and* repo to start with an alphanumeric character, while `resolve_pr_json` and `resolve_origin`'s capture groups (`[A-Za-z0-9._-]+`) impose no such constraint.

**Two consequences, verified empirically:**
1. `resolve_explicit` spuriously rejects the real GitHub `.github` special-repo convention, which the other two paths accept for the identical name — silently degrading `--repo`-scoped callers to legacy/unscoped naming for such repos:
```
$ bash bin/swe-workbench-repo-scope --repo "octocat/.github"; echo "exit=$?"
exit=1
$ (cd <repo-with-origin-octocat/.github> && bash bin/swe-workbench-repo-scope)
octocat-.github
```
2. The looser regex in `resolve_pr_json`/`resolve_origin` permits a leading hyphen in owner/repo, which breaks the `escape_owner_repo` injectivity proof this PR exists to establish:
```
owner='a-', repo='b'   -> escape_owner_repo -> 'a---b'
owner='a',  repo='-b'  -> escape_owner_repo -> 'a---b'   # collision
```
`resolve_explicit`'s own regex would reject `'a/-b'` (leading-alnum on both halves), but `resolve_pr_json`/`resolve_origin`'s regex would not catch the equivalent case through a PR-JSON `.url` or a git-remote origin. `tests/test_repo_scope.py` never exercises a leading-hyphen/leading-dot component through any of the three modes.

**Suggested fix:** use one shared, correct character-class regex across all three functions — anchor the leading-alnum requirement on `owner` only (matches real GitHub org/user rules, which forbid a leading hyphen) and drop it for `repo` (GitHub repo names may start with `.`, e.g. `.github`). Add tests exercising a leading-dot repo name and a leading-hyphen component through all three resolution modes.

bin/swe-workbench-sweep-residuals
lugassawan: Addressed in 6fec75132b3998099e0f18648cb39deb461284cb: routed _url_scope_slug through swe-workbench-repo-scope --repo directly instead of reimplementing the join, so both sides of every comparison share the same (now injective) encoding.

bin/swe-workbench-address-feedback-worktree
lugassawan: Addressed in 6fec75132b3998099e0f18648cb39deb461284cb: do_release's legacy-receipt dual-read now verifies the receipt's own recorded path attributes to this repo (mirrors pr-review-worktree's _attributes_here) before adopting it.

bin/swe-workbench-repo-scope
lugassawan: Addressed in 6fec75132b3998099e0f18648cb39deb461284cb: all three resolve_* functions now share one _owner_repo_ok validator (owner requires a leading alphanumeric; repo doesn't, but excludes the traversal-shaped "."/".."). Also replaced the underlying hyphen-doubling scheme itself with a length-prefix join, since doubling turned out to be non-injective too (owner='a-'/repo='b' and owner='a'/repo='-b' both produced 'a---b') — see the updated shared/docs/repo-scoped-ephemeral-state.md.

PR #719, opened 2026-09-04
[fix] Sync peerDependencies floor after pi 0.84.3->0.84.4 bump

lugassawan:
## Summary

- Dependabot PRs #715/#716 bumped `@earendil-works/pi-coding-agent`/`pi-tui` from 0.84.3 to 0.84.4 in `devDependencies`, but left `peerDependencies`' floor at `>=0.84.3` — `test_package_json_values` requires the floor stay in exact lockstep with the pin (this drifted once before, see `0e34767`), so `main`'s pytest job is currently red.
- Also fixes a related `package-lock.json` drift: the nested `pi-coding-agent -> pi-tui` dependency range had collapsed from `^0.84.4` to an exact `0.84.4`, diverging from the version actually published upstream.
- Root cause for why this recurs: Dependabot only ever bumps the exact `devDependencies` pin — it never touches `peerDependencies`, because `>=0.84.3` already satisfies `0.84.4` (no semver violation for its updater to fix). Added `scripts/sync-peer-deps.sh` (`--check` / apply modes, covered by `tests/test_sync_peer_deps_sh.py`) plus `.github/workflows/dependabot-peer-sync.yml`, a `pull_request_target` job gated to `actor == dependabot[bot]` and `paths: [package.json]` that overwrites its own checked-out copy of the script with the trusted `origin/main` version before running it in place (never executing anything else from the PR tree — no `npm install`, so a compromised package's install script can't run under this job's write token), and runs it against the PR's own manifests, pushing a fixup commit when drift is found — so future bumps of these two packages self-heal instead of needing a manual fix every time.
- Per [GitHub's docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow), a `GITHUB_TOKEN`-authored push to a PR *does* retrigger the `pull_request` run, but lands it in an **approval-required** state — so after this fixup commit lands, a maintainer still needs one click of "Approve workflows to run" on the resulting `pr.yml` run. Getting a fully automatic re-run would need a PAT or GitHub App token instead of `GITHUB_TOKEN`, which this PR deliberately avoids to keep the workflow secret-free.
- The issue-reference opt-out below is additionally gated on `package.json` actually being in the PR's diff (checked via the GitHub API), not just on the body text containing the bump-style line — otherwise any PR could copy-paste that one line to dodge the issue-reference requirement.
- Small related fix: `pr.yml`'s issue-reference check now also accepts a manually-authored PR body carrying Dependabot's own `Bumps X from Y to Z.` line as an opt-out, same as the existing `dependabot[bot]`-actor exemption (mirrored in `tests/test_pr_validation.py`).
- `scripts/sync-peer-deps.sh` also asserts pi-coding-agent's and pi-tui's `devDependencies` pins stay in lockstep before deriving the expected floor from either one, so a partial bump of just one of the two packages fails loudly instead of silently syncing both floors to a pin only one package actually moved to.

- **Second review pass caught a High-severity bug, reproducible from this repo's own history**: pi-coding-agent and pi-tui bump as two *separate* dependabot PRs (`dependabot.yml` has no npm `groups:`), so the first PR of every such pair (already happened twice: #715/#716, and earlier #685/#686) leaves the two packages' `devDependencies` pins out of lockstep. `sync-peer-deps.sh` correctly refused to guess a floor in that state, but the workflow treated *any* nonzero `--check` exit as actionable drift and ran the apply step anyway — which hit the identical guard and failed the whole job, turning a clean "nothing to do yet" into a confusing extra red check. Fixed by giving the script a distinct exit code (`2` = hard error, cannot proceed vs. `1` = actionable drift) and having the workflow only apply on `1`, quietly no-op with a warning annotation otherwise (`tests/test_dependabot_peer_sync_workflow.py` locks this in against the actual embedded YAML bash).
- Also fixed from the same pass: the script overwriting itself via `git show origin/main:... > scripts/sync-peer-deps.sh` and then invoking it *in place* (not copied to `/tmp`, which broke its `BASH_SOURCE`-relative root detection — a Critical bug that would have made the whole automation fail on its first real run); the issue-reference opt-out is now gated on `package.json` actually being in the PR's changed-files list (via a live `gh api` call), not just body text, so it can't be copy-pasted into an unrelated PR; the script also asserts pi-coding-agent's and pi-tui's `devDependencies` pins stay in lockstep before deriving either floor; and added a static test (`test_pi_extension.py::test_package_lock_peer_dependencies_match_package_json`) asserting the checked-in `package-lock.json` peerDependencies actually matches `package.json`'s.

- **Third (independent, from-scratch) review pass approved** with only two non-blocking nits: the live `gh api` changed-files gate on the "Bumps X from Y to Z." opt-out had no behavioral test coverage, and that gate hard-failed the whole step on a transient API error instead of falling through to the closing-keyword check. Both fixed: added `tests/test_pr_validation.py::TestIssueReferenceBumpOptoutGating` (drives the actual bash with a stubbed `gh`), and the `gh api` call now falls through to `CHANGED_FILES=""` on error rather than aborting the step. Also hardened `test_dependabot_peer_sync_workflow.py`'s bash-extraction regex, which previously could silently truncate its capture on an unrelated future edit instead of failing loudly.

## Test Plan

- [x] `bash scripts/validate.sh` passes
- [x] Full `pytest tests/` (3816 tests) passes locally, including the new `test_pi_extension.py::test_package_json_values` and `test_sync_peer_deps_sh.py`/`test_pr_validation.py` coverage
- [x] `npm ci` installs cleanly from the corrected `package-lock.json`
- [x] Manually verified `scripts/sync-peer-deps.sh` end-to-end in an isolated scratch copy: simulated drift → `--check` fails → apply fixes both files → `--check` passes → output byte-identical to the real repo's already-fixed files

- [x] Reproduced the original failure (`test_package_json_values` red against the un-synced floor) before fixing, and confirmed it's green after
- [x] Verified the fix addresses the root cause (the floor/pin coupling), not just this one instance — added automation so it holds on the next dependabot bump too

Issue: N/A — dependency-bump follow-up fix and CI hardening, not tied to a filed issue

### DIFF
--- a/.github/workflows/dependabot-peer-sync.yml
+++ b/.github/workflows/dependabot-peer-sync.yml
@@ -11,13 +11,25 @@ name: Dependabot peerDependencies Sync
 #
 # pull_request_target is required for write access to push back to the PR
 # branch. To stay safe against a compromised PR, this workflow:
-#   - never executes anything checked out from the PR itself — it fetches the
-#     sync script from the base branch (origin/main) and runs only that copy
+#   - never executes anything checked out from the PR itself — it overwrites
+#     scripts/sync-peer-deps.sh with the copy from the base branch
+#     (origin/main) before running it, so only that trusted content ever runs,
+#     invoked in place so its own BASH_SOURCE-relative root detection still
+#     resolves to the checked-out repo (not /tmp)
 #   - only ever reads/writes package.json and package-lock.json — no `npm
 #     install`, so a malicious package's install/postinstall script can never
 #     run with this job's write token
 #   - is gated on the PR author being dependabot[bot] and the diff touching
 #     package.json
+#
+# The resulting fixup-commit push authenticates with the default
+# GITHUB_TOKEN, so per GitHub's own docs the pull_request run it triggers on
+# pr.yml lands in an approval-required state rather than running immediately
+# — https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow.
+# A maintainer still needs one click of "Approve workflows to run" on that
+# run; getting a fully automatic re-run would require a PAT or GitHub App
+# token in place of GITHUB_TOKEN, which we deliberately avoid here to keep
+# this workflow secret-free.
 on:
   pull_request_target:
     types: [opened, synchronize]
@@ -38,25 +50,27 @@ jobs:
     timeout-minutes: 5
     permissions:
       contents: write
-      pull-requests: write
     steps:
       # Checked out for git history/parentage only (so a fixup commit lands
-      # on top of the PR's real commits) — nothing from this tree is executed.
+      # on top of the PR's real commits) — every file in this tree is
+      # untrusted until the next step overwrites the one script we execute.
       - name: Checkout PR head
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
 
-      - name: Fetch trusted sync script from main
+      # Overwrite in place (not copied to /tmp) so the script's own
+      # BASH_SOURCE-relative root detection still resolves correctly.
+      - name: Overwrite sync script with the trusted copy from main
         run: |
           git fetch --depth 1 origin main
-          git show origin/main:scripts/sync-peer-deps.sh > /tmp/sync-peer-deps.sh
+          git show origin/main:scripts/sync-peer-deps.sh > scripts/sync-peer-deps.sh
 
       - name: Check for peerDependencies drift
         id: check
         run: |
-          if bash /tmp/sync-peer-deps.sh --check; then
+          if bash scripts/sync-peer-deps.sh --check; then
             echo "drift=false" >> "$GITHUB_OUTPUT"
           else
             echo "drift=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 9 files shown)


lugassawan (commented):
_Reviewed by `reviewer`_([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comment(s) and 0 PR-level note(s), deduped 0.

lugassawan:
## Review summary

Three review passes ran against this PR (`swe-workbench:reviewer`, dispatched independently each time against the current HEAD):

**Pass 1** (against `c193778`) found a **Critical** bug: `dependabot-peer-sync.yml` copied `scripts/sync-peer-deps.sh` into `/tmp` before running it, but the script derives its own root via `BASH_SOURCE`, so the relocation made every path resolve against `/` instead of the checked-out repo — the whole automation would have failed on its first real dependabot bump. Also found: a loose issue-reference opt-out any PR could copy-paste to dodge the check, a missing pi-tui/pi-coding-agent `devDependencies` lockstep check, a missing error guard, and a close/reopen race. All fixed in `4f74531`. Separately (via GitHub's own docs, not the review), confirmed the close/reopen step never actually achieved automatic re-triggering — a `GITHUB_TOKEN` push already retriggers `pull_request` runs, just gated behind "Approve workflows to run" — so that mechanism was removed entirely rather than fixed.

**Pass 2** (against `4f74531`) found a **High**-severity bug reproducible from this repo's own history: `pi-coding-agent`/`pi-tui` bump as two *separate* dependabot PRs (no npm `groups:` in `dependabot.yml`), so the first PR of every such pair — already happened twice, #715/#716 and earlier #685/#686 — left the two packages' `devDependencies` pins out of lockstep. The new script correctly refused to guess a floor in that state, but the workflow treated any nonzero `--check` exit as actionable drift and ran the apply step anyway, hitting the identical guard and failing the whole job. Fixed in `770a9a0` by giving the script distinct exit codes (`1` = actionable drift, `2` = hard error) and adding `tests/test_dependabot_peer_sync_workflow.py`, which drives the actual embedded workflow bash against stubbed exit codes.

**Pass 3** (against `770a9a0`) was a from-scratch, skeptical re-review that independently re-derived the lockstep scenario from real git history and empirically verified every prior fix rather than trusting the PR's own narrative. **Approved**, with two non-blocking nits: the live `gh api` changed-files gate on the opt-out had no behavioral test coverage, and it hard-failed the step on a transient API error instead of falling through to the closing-keyword check. Both fixed in `acb6e0f`.

Full `pytest tests/` (3816 tests) passes at the final commit; `scripts/validate.sh` and `npm ci` both clean.

